PR #71, opened 2020-11-06
Getting Started tutorial suggestions

FreekDB:
- Used (for example) androidx.compose.material.Text instead of androidx.compose.foundation.Text, because the latter function is deprecated.
- Used val for count instead of var.
- The Gradle task can be found in the "compose desktop" category, so used "sample/Tasks/compose desktop/run" instead of "sample/Tasks/applications/run".
- Minor textual changes.

### DIFF
--- a/tutorials/Getting_Started/README.md
+++ b/tutorials/Getting_Started/README.md
@@ -22,8 +22,8 @@ The recommended way of building Compose for Desktop projects is by using Gradle.
 JetBrains provides a simple way of building Compose for Desktop projects
 using a special Gradle plugin.
 
-One could clone an existing template [desktop](https://github.com/JetBrains/compose-jb/tree/master/templates/desktop-template) or
-[multiplatform](https://github.com/JetBrains/compose-jb/tree/master/templates/multiplatform-template) application templates, or create it from scratch.
+One could clone an existing template for a [desktop](https://github.com/JetBrains/compose-jb/tree/master/templates/desktop-template) or
+[multiplatform](https://github.com/JetBrains/compose-jb/tree/master/templates/multiplatform-template) application, or create it from scratch.
 
 First create a new directory, named `sample`.
 ```shell script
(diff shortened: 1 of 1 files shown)


tutorials/Getting_Started/README.md
```diff
@@ -81,7 +81,7 @@ import androidx.compose.ui.unit.IntSize
 import androidx.compose.ui.unit.dp
 
 fun main() = Window(title = "Compose for Desktop", size = IntSize(300, 300)) {
-    var count = remember { mutableStateOf(0) }
+    val count = remember { mutableStateOf(0) }
```
NahroTo, line 84: I don't think `val` applies here, since `count` is a mutable state, hence `count = 2` is possible. Not sure why the readme uses the `count.value` notation though.

FreekDB: Hi Orhan,

Thanks for your message! The count variable is a bit mysterious. It is created during composition by the remember function (see link 1). It calls the mutableStateOf function (see link 2), which creates a MutableState object that contains a mutable value (which is why the count.value notation has to be used). The count variable itself can be immutable, since it is assigned only once; IntelliJ IDEA also gives a warning about this. This approach is somewhat comparable to creating a mutable list and assigning it to an immutable val variable.

Cheers, Freek


Link 1: https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#remember

Link 2: https://developer.android.com/reference/kotlin/androidx/compose/runtime/package-summary#mutablestateof

NahroTo: Hi Freek,

Your approach indeed applies to the current way the count state is setup.
Right now it reads as
`val count: MutableState<Int> = remember { mutableStateOf(0) }`
which I suggest we change it to the more idiomatic approach using property delegation:
`var count: Int by remember { mutableStateOf(0) }`.
You can read more about this in the Compose guide under the linked section (second blue text block): https://developer.android.com/jetpack/compose/state#use-rememeber-to-create-internal-state-in-composable

olonho: Yeah, delegation is also possible, but requires additional imports, so you need to add those as well to ensure sample compiles.

PR #98, opened 2020-11-13
Add Gradle plugin integration tests

AlexeyTsvetkov:


gradle-plugins/compose/build.gradle.kts
```diff
@@ -54,3 +56,25 @@ val jar = tasks.named<Jar>("jar") {
     from(zipTree(shadow.get().archiveFile))
     this.duplicatesStrategy = DuplicatesStrategy.INCLUDE
 }
+
+val gradleVersionForTests = "6.7"
+val java14Home: String? = when (JavaVersion.current()) {
+    JavaVersion.VERSION_14 -> System.getProperty("java.home")
```
olonho, line 62: As discussed, let's make it >=.

gradle-plugins/compose/src/test/test-projects/application/jvm/settings.gradle
```diff
@@ -0,0 +1,11 @@
+pluginManagement {
+    plugins {
+        id 'org.jetbrains.kotlin.jvm' version '1.4.10'
```
olonho, line 3: Interesting, why 1.4.10? Maybe 1.4.20 (latest) or 1.4.0 (used in Compose)?

olonho (commented):
And as discussed, let's specify smallest supported Gradle version, and in test in more versions.

gradle-plugins/compose/src/test/test-projects/application/jvm/settings.gradle
AlexeyTsvetkov: Fixed

olonho (approved):

PR #123, opened 2020-11-19
Added tutorial on Swing integration.

Rsedaikin:


tutorials/Swing_Integration/README.md
```diff
@@ -0,0 +1,152 @@
+# Integration Compose for Desktop into Swing-based aplication
+
+## What is covered
+
+In this tutorial, we will show you how to use ComposePanel in your Swing application.
+
+## Using ComposePanel
+
+ComposePanel lets you create a UI using Compose for Desktop in a Swing-based UI. To achieve that you need to create instance of ComposePanel, add it to your Swing layout and describe composition inside `setContent`. Also you may need to crear the CFD application events via `AppManager.setEvents`.
```
olonho, line 9: create?

Rsedaikin: may be "build" is better?

tutorials/Swing_Integration/README.md
```diff
@@ -0,0 +1,152 @@
...
+
+```kotlin
```
igordmn, line 11: Maybe add information about disposing ComposePanel?

If we add ComposePanel into another AWT component it will be disposed automatically.

If we don't add then we should call "panel.dispose()" manually.

Rsedaikin: Where do you think is the best place to add this? I could add this inside this place:
```
window.contentPane.add(
    actionButton(
        text = "SOUTH/REMOVE COMPOSE",
        action = {
            // if you directly remove ComposePanel from Swing container it is disposed automatically,
            // but in some cases you may want to dispose ComposePanel manually via composePanel.dispose()
            window.contentPane.remove(composePanel)
        }
    ),
    BorderLayout.SOUTH
)
```

igordmn: Looks fine 👍

Rsedaikin: No, the panel.dispose () method does not exist. The ComposePanel works like any other Swing component and is only removed if you remove it from the container.

tutorials/Swing_Integration/README.md
```diff
@@ -0,0 +1,152 @@
+# Integration Compose for Desktop into Swing-based aplication
+
+## What is covered
+
+In this tutorial, we will show you how to use ComposePanel in your Swing application.
+
+## Using ComposePanel
+
+ComposePanel lets you create a UI using Compose for Desktop in a Swing-based UI. To achieve that you need to create instance of ComposePanel, add it to your Swing layout and describe composition inside `setContent`. Also you may need to clear the CFD application events via `AppManager.setEvents`.
```
B1ggDave, line 9: ComposePanel lets you create a UI using Compose for Desktop in a Swing-based UI. To achieve this you need to create an instance of ComposePanel, add it to your Swing layout, and describe the composition inside `setContent`. You may also need to clear the CFD application events via `AppManager.setEvents`.

tutorials/Swing_Integration/README.md
igordmn: ComposePanel will be disposed automatically now:
https://android-review.googlesource.com/c/platform/frameworks/support/+/1507296

PR #166, opened 2020-12-03
Add iOS target and app for todoapp example

arkivanov:
First of all I would like to mention, please feel free to reject the PR if you think it is not in line with the goals of this project.

This PR adds iOS target and sample app for the todoapp example project. Because HMPP does not support shared JVM source sets, I had to extract all Compose UI to the separate module: `compose-ui`. Also updated Jetpack and JetBrains Compose to the latest versions. And migrated to `kotlin-parcelize` plugin.

Pros:
- It demonstrates how we can add platform specific UI for targets where JetBrains Compose can not be used
- The navigation is still shared across all targets

Cons:
- Updating the todoapp example project to a new Kotlin version may be delayed if ABI is not compatible


olonho (commented):
Looks really cool, could you please rebase on latest master?

arkivanov:
> Looks really cool, could you please rebase on latest master?

Rebased

olonho:
Checked PR, and got this error in IDE
image
However desktop, iOS and Android version seems to work, so let's merge it. @arkivanov if you got a chance to fix IDE error - would be cool.

arkivanov:
@olonho I will try to fix but I might need you help. Seems like an IDE error in the desktop module.

PR #206, opened 2020-12-21
Custom components template.

olonho:


components/settings.gradle.kts
```diff
@@ -0,0 +1,3 @@
+include(":VideoPlayer")
```
igordmn, line 1: Maybe we keep only two modules? The purpose of ":VideoPlayer" is unclear.

":VideoPlayer:common" will be for the main library which we will publish in the maven.
And ":VideoPlayer:desktop" is for examples?

olonho: Yep, indeed, fixed.

igordmn: We can also remove file `components/VideoPlayer/build.gradle.kts`

components/VideoPlayer/common/src/commonMain/kotlin/org/jetbrains/compose/videoplayer/VideoPlayer.kt
```diff
@@ -0,0 +1,10 @@
+package org.jetbrains.compose.videoplayer
+
+import androidx.compose.runtime.Composable
+
+@Composable
+fun VideoPlayer(url: String, width: Int, height: Int) {
+    VideoPlayerImpl(url, width, height)
+}
+
+expect fun VideoPlayerImpl(url: String, width: Int, height: Int)
```
igordmn, line 10: internal?

olonho: Yep, think alike, already fixed that in update :).

PR #228, opened 2020-12-28
Update README.md (Add trailing commas)

YektaDev:
Add trailing commas


olonho (changes requested):
Not sure if we want to enforce trailing commas.

YektaDev:
I was wondering to know why not?

olonho:
It's not adding value.

PR #242, opened 2020-12-30
Add widget gallery example

AlexeyTsvetkov:


examples/widgetsgallery/common/src/commonMain/kotlin/org/jetbrains/compose/widgets/theme/Theme.kt
```diff
@@ -0,0 +1,136 @@
+package org.jetbrains.compose.widgets.theme
```
olonho, line 1: Better use another package i.e. `org.jetbrains.compose.demo.widgets`.

olonho (approved):

PR #253, opened 2021-01-07
Add custom Spoiler component

xetra11:


components/Spoiler/common/src/commonMain/kotlin/org/jetbrains/compose/custom/Spoiler.kt
```diff
@@ -0,0 +1,39 @@
+import androidx.compose.foundation.clickable
```
olonho, line 1: All custom components must be in their own package.
This particular component looks rather trivial at the moment, do you see how it could be evolved?

xetra11: Like adding an icon property to is (Bitmap) that will be shown on the left hand side of the dropdown/button/label maybe?

xetra11: Also I was unsure what the requirements for components for this were. There I just tried adding something via this PR

olonho: Generally important requirement is that the component is non-trivial, and provides value. I.e. table, videoplayer, etc.

xetra11: Ah ok non trivial it is - then forget about my thing here

PR #298, opened 2021-01-21
Avoid adding incompatible arguments when creating application image

### LINKED ISSUE #296: Creating application image fails with certain DSL options
As of `0.3.0-build140` running `createDistributable` or `runDistributable` fails on Windows by default and on Linux with certain options enabled (e.g. `linux { shortcut = true }`)

AlexeyTsvetkov:
Fixes #296


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
```diff
@@ -187,13 +189,15 @@ abstract class AbstractJPackageTask @Inject constructor(
 
         when (currentOS) {
```
olonho, line 190: Shouldn't `if (targetFormat != TargetFormat.AppImage) {` just go before when?

PR #342, opened 2021-02-02
Improve `buildSrc` setup

JavierSegoviaCordoba:


examples/todoapp/buildSrc/buildSrc/build.gradle.kts
```diff
@@ -2,6 +2,11 @@ plugins {
     `kotlin-dsl`
 }
 
-repositories {
-    jcenter()
+repositories.jcenter()
+
+sourceSets.main {
+    java {
```
olonho, line 8: Why `java`?

JavierSegoviaCordoba: I can't get this behavior using the `kotlin` block and searching on Google I only see implementations using the `java` block.

examples/todoapp/buildSrc/buildSrc/README.md
```diff
@@ -0,0 +1,4 @@
+# Why this subfolder/submodule
+
+This folder, `buildSrc/buildSrc`, force compiling `Deps.kt` file, so it can be used
```
arkivanov, line 3: So the Deps.kt file is moved from `/buildSrc/buildSrc` to `/buildSrc` folder, but we still need this empty folder?

JavierSegoviaCordoba: It is not empty, it has two files, `README.md` and `build.gradle.kts` indicating where `Deps.kt` file is.

arkivanov: So if we still need the nested `buildSrc` folder, I would keep Deps.kt in there, because there is less configuration code and it is also less confusing. Other changes LGTM.

arkivanov (approved):

examples/todoapp/buildSrc/buildSrc/README.md
JavierSegoviaCordoba: Changed!

PR #351, opened 2021-02-05
Updates the intelliJPlugin example with currently IDEA Theme colors

DevSrSouza:
This pull request change the example of the IntelliJ Plugin to use the currently `background` and `foreground` colors from IDEA Theme.

It also listens to `com.intellij.ide.ui.LafManagerListener` in order to update the state from `background` and `onBackground`.

image

Example using theme **Dark Purple**

### DIFF
--- a/examples/intelliJPlugin/src/main/kotlin/com/jetbrains/compose/theme/intellij/SwingColor.kt
+++ b/examples/intelliJPlugin/src/main/kotlin/com/jetbrains/compose/theme/intellij/SwingColor.kt
@@ -1,0 +1,35 @@
+package com.jetbrains.compose.theme.intellij
+
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.ui.graphics.Color
+import javax.swing.UIManager
+import java.awt.Color as AWTColor
+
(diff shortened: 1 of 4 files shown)


examples/intelliJPlugin/src/main/kotlin/com/jetbrains/compose/theme/intellij/SwingColor.kt
```diff
@@ -0,0 +1,35 @@
...
+import java.awt.Color as AWTColor
+
+interface SwingColor {
+    companion object : SwingColor by SwingColorImpl
+
+    val background: Color
+    val onBackground: Color
+}
+
+object SwingColorImpl : SwingColor {
```
olonho, line 16: This singleton object doesn't look optimal (as almost any mutable singleton), could we instead pass color change notifications to plugin instance.

DevSrSouza: Yeh, I agree with that, but the idea here was to do something copy-paste.
I will update it to register the listener programmatically and use the lifecycle of the Composable to unregister the listener.

olonho (approved):

PR #383, opened 2021-02-15
Fix dependencies clash when packaging

### LINKED ISSUE #319: Multi module app packaging issue when having sub modules with the same name
When we have a multi module app with sub modules using same name one of the sub modules are not included in package.

How to reproduce: Create an app with sub modules like this
root
->android
->desktop
->common
--->compose-ui
------>main
------>feature1
------>feature2
--->logic
------>main
------>feature1
------>feature2

AlexeyTsvetkov:
Fixes #319

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/files/fileUtils.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/files/fileUtils.kt
@@ -1,0 +1,20 @@
+package org.jetbrains.compose.desktop.application.internal.files
+
+import java.io.File
+import java.security.DigestInputStream
+import java.security.MessageDigest
+
+internal fun fileHash(file: File): String {
+    val md5 = MessageDigest.getInstance("MD5")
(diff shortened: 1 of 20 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/files/fileUtils.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import java.io.File
+import java.io.FileInputStream
+import java.security.MessageDigest
+import java.util.*
+import java.util.zip.ZipInputStream
+
+internal fun jarHash(file: File): String {
+    val md5 = MessageDigest.getInstance("MD5")
+    val entryMap = TreeMap<String, ByteArray>()
```
olonho, line 11: Why do you need ordered map here?

olonho: Also why not just hash the whole file as binary? Why unpack it?

Issue #272, opened 2021-01-14
Using AndroidX Compose specific dependencies for Android target


audkar:
#### Current problem
Can't use Android target specific Compose libraries (like [Navigation](https://developer.android.com/jetpack/androidx/releases/navigation) or [Paging](https://developer.android.com/jetpack/androidx/releases/paging)) when using compose-jb dependencies in same project (MPP with desktop and android targets).

e.g. adding android target specific dependency `implementation("androidx.navigation:navigation-compose:1.0.0-alpha05")`
fails with:
```
Execution failed for task ':android-app:checkDebugDuplicateClasses'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckDuplicatesRunnable
   > Duplicate class androidx.compose.animation.core.AndroidAnimationClockKt found in modules animation-core-1.0.0-alpha10-runtime (androidx.compose.animation:animation-core:1.0.0-alpha10) and animation-core-debug-runtime (org.jetbrains.compose.animation:animation-core-android-debug:0.3.0-build138)
<...>
```

#### Reasoning
Google and community will create more AndroidX Compose specific libraries. Having possibility to use them in android target is mandatory in long term.

#### Short term
It's becoming quite challenging to have real-word MPP project with with both targets. We are forced to choose between having shared code or using existing libraries. Maybe it would be possible to clone at-least official AndroidX compose libraries (navigation/paging) into compose-jb project?

olonho:
This is indeed a legitimate issue, which could be worked around by using `exclude(group = "androidx.compose")` inside problematic dependencies, but generally we're aiming at having consolidated publications.

igordmn:
For now, we can exclude all androidx.compose dependencies from the project:

```
configurations {
    "implementation" {
        exclude(group = "androidx.compose.animation")
        exclude(group = "androidx.compose.foundation")
        exclude(group = "androidx.compose.material")
        exclude(group = "androidx.compose.runtime")
        exclude(group = "androidx.compose.ui")
    }
}
```
It is not ideal, but should work. We'll figure out something better in the future.

audkar:
My concern is that implementations can missmatch and cause runtime problems. Knowing how fast code is changing atm I see high possibility of such case happening.

igordmn:
>My concern is that implementations can missmatch and cause runtime problems. 
Yes, there is a problem. In many cases it can be resolved just using the latest version of `org.jetbrains.compose`.

[...]

PR #399, opened 2021-02-18
Gradle plugin. Replace 'androidx.compose' artifacts by 'org.jetbrains.compose' artifacts.

igordmn:
Fixes https://github.com/JetBrains/compose-jb/issues/272

### DIFF
--- a/gradle-plugins/buildSrc/build.gradle.kts
+++ b/gradle-plugins/buildSrc/build.gradle.kts
@@ -9,4 +9,6 @@ repositories {
 dependencies {
     compileOnly(gradleApi())
     implementation(kotlin("stdlib"))
+    implementation("io.ktor:ktor-client-core:1.5.1")
+    implementation("io.ktor:ktor-client-cio:1.5.1")
 }
--- a/gradle-plugins/buildSrc/src/main/kotlin/printAllAndroidxReplacements.kt
+++ b/gradle-plugins/buildSrc/src/main/kotlin/printAllAndroidxReplacements.kt
@@ -6,16 +6,17 @@ import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.runBlocking
+import org.gradle.api.Project
 
 private const val libsRepo = "https://maven.pkg.jetbrains.space/public/p/compose/dev/org/jetbrains/compose/"
-private const val version = "0.3.0-build154"
 private val exceptions = listOf(
     "desktop:desktop",
     "compose-full",
     "compose-gradle-plugin"
 )
 
-fun printAllAndroidxReplacements() = runBlocking {
+fun Project.printAllAndroidxReplacements() = runBlocking {
+    val version = BuildProperties.composeVersion(project)
     HttpClient().use { client ->
         client
             .allRecursiveFolders(libsRepo)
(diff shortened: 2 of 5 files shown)


gradle-plugins/buildSrc/build.gradle.kts
```diff
@@ -9,4 +9,6 @@ repositories {
 dependencies {
     compileOnly(gradleApi())
     implementation(kotlin("stdlib"))
+    implementation("io.ktor:ktor-client-core:1.5.1")
```
olonho, line 12: Why it is needed?

gradle-plugins/buildSrc/src/main/kotlin/printAllAndroidxReplacements.kt
```diff
@@ -0,0 +1,66 @@
...
+import io.ktor.client.request.*
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.runBlocking
+
+private const val libsRepo = "https://maven.pkg.jetbrains.space/public/p/compose/dev/org/jetbrains/compose/"
+private const val version = "0.3.0-build154"
```
olonho, line 11: Please mark it with __LATEST_COMPOSE_RELEASE_VERSION__

gradle-plugins/buildSrc/build.gradle.kts
igordmn: It is for printAllAndroidxReplacements script (fetch all needed dependencies from maven repo).

It is only available for plugin build, and won't be in the plugin itself.

gradle-plugins/buildSrc/src/main/kotlin/printAllAndroidxReplacements.kt
igordmn: I think it is better to read this version from gradle.properties

olonho (approved):

jimgoog:
👍  I really look forward to the day when we are unified.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -37,11 +38,59 @@ class ComposePlugin : Plugin<Project> {
...
 
             project.dependencies.add(
-                    "kotlinCompilerPluginClasspath",
-                    "org.jetbrains.compose.compiler:compiler:$composeVersion"
+                "kotlinCompilerPluginClasspath",
+                "org.jetbrains.compose.compiler:compiler:$composeVersion"
             )
         }
 
+        fun ComponentModuleMetadataHandler.replaceAndroidx(original: String, replacement: String) {
```
AlexeyTsvetkov, line 46: Should the replacements be configurable? (i.e. should there be a flag to disable replacements?)

AlexeyTsvetkov (commented):
I think re

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
igordmn: I thought about it. But haven't figure out where it might be useful.

In all cases when we have two libraries in classpath ("androidx.compose" and "org.jetbrains.compose") we always will have an error.

If we have only one library, then everything will work as before.

PR #406, opened 2021-02-20
Print notarization status

AlexeyTsvetkov:


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/MultiOutputStream.kt
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    override fun write(b: Int) {
+        super.write(b)
+        secondaryStream.write(b)
+    }
+
+    override fun flush() {
+        super.flush()
+        secondaryStream.close()
```
olonho, line 27: flush as well, maybe?

igordmn, line 27: secondaryStream.flush() ?

AlexeyTsvetkov: Yes! Fixed

PR #417, opened 2021-02-23
Update the screenshot from IntelliJ Plugin example

DevSrSouza:


olonho:
Not sure what's the point of having an empty file as screenshot.

jimgoog:
@olonho afaik, the file is not empty.  It looks like the biggest change is that it is slightly higher resolution, and has the various controls in slightly different positions.

That said, I'm not sure I understand the motivation for this change.  @DevSrSouza Was there something wrong with the previous screenshot or why is the update necessary?

DevSrSouza:
@olonho This screenshot is a crop taken from the [Milestone 3 blog post](https://blog.jetbrains.com/kotlin/2021/02/jetpack-compose-for-desktop-milestone-3-released/).

The current screenshot is outdated because of the PR #351

jimgoog:
The colors look the same to me.  The colors in PR https://github.com/JetBrains/compose-jb/pull/351 look different (specifically, the "ok" button is purple in PR #351 but it remains blue in this PR's screenshot.

DevSrSouza:
Yes, in the PR looks more different because I use a different IntelliJ theme. Comparing the two screenshots "side by side" I think it will make easier to understand the changes. The background/onBackground on Compose UI is different because it was not using the IntelliJ Theme background/foreground color.

Currently screenshot:


New screenshot taken from the Blog post:

olonho:
Thanks for clarification.

PR #427, opened 2021-02-25
Specify minimum of JDK 15 for jpackage

jimgoog:


tutorials/Native_distributions_and_local_execution/README.md
```diff
@@ -152,14 +152,14 @@ Versions must follow the rules:
 
 ## Customizing JDK version
 
-The plugin uses `jpackage`, which is available since [JDK 14](https://openjdk.java.net/projects/jdk/14/).
+The plugin uses `jpackage`, for which you should be using at least [JDK 15](https://openjdk.java.net/projects/jdk/15/).
```
olonho, line 155: Actually it's mandatory only in case of Linux.

olonho: Seems it's required for Windows as well, so let's commit.

PR #433, opened 2021-02-26
Set dock and menu names on macOS

### LINKED ISSUE #246: [macOS] compose.desktop.application.nativeDistributions.packageName only used for menu title, not menu items
When configuring

```kotlin
compose.desktop {
    application {
        mainClass = "com.thomaskuenneth.tkdupefinder.MainKt"
        nativeDistributions {
            targetFormats(TargetFormat.Dmg,
                    TargetFormat.Msi)
            packageName = "TKDupeFinder"
            version = "0.1-SNAPSHOT"
```

the package name is picked up nicely for the menu title, but not for **About** and **Quit** menu items, as seen in the following screenshot:

image

The usual behavior on macOS that the application name (package name) is also used for this two menu items.

AlexeyTsvetkov:
Resolves #246

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
@@ -120,6 +120,10 @@ abstract class AbstractJPackageTask @Inject constructor(
 
     @get:Input
     @get:Optional
+    val macDockName: Property<String?> = objects.nullableProperty()
+
+    @get:Input
+    @get:Optional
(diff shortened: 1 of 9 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
```diff
@@ -118,6 +118,10 @@ abstract class AbstractJPackageTask @Inject constructor(
     @get:Optional
     val macPackageName: Property<String?> = objects.nullableProperty()
 
+    @get:Input
+    @get:Optional
+    val macDockName: Property<String?> = objects.nullableProperty()
```
olonho, line 123: Shalln't we document it in tutorial?

PR #501, opened 2021-03-16
Add suggestRuntimeModules task

### LINKED ISSUE #463: Verify module dependencies when building native images
For context, see: https://kotlinlang.slack.com/archives/C01D6HTPATV/p1614771641211500

When running native distributables, users are getting `ClassNotFoundException` for various packages.  The problem is that they failed to add `module(X)` inside `nativeDistributions` block, where `X` = the missing module.  This feels like the type of thing we should have caught at compile time, and should have given a message that was more actionable than `ClassNotFoundException`.

AlexeyTsvetkov:
Resolves #463

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/osUtils.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/osUtils.kt
@@ -1,7 +1,9 @@
 package org.jetbrains.compose.desktop.application.internal
 
+import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.Internal
 import org.jetbrains.compose.desktop.application.dsl.TargetFormat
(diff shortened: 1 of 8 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/osUtils.kt
```diff
@@ -63,6 +64,15 @@ internal object MacUtils {
     }
 }
 
+internal fun jvmToolFile(toolName: String, javaHome: Provider<String>): File {
+    val jtool = File(javaHome.get()).resolve("bin/${executableName(toolName)}")
+    check(jtool.isFile) {
+        "Invalid JDK: $jtool is not a file! \n" +
+                "Ensure JAVA_HOME or buildSettings.javaHome is set to JDK 14 or newer"
```
olonho, line 71: JDK 15?

olonho:
Please add documentation entry for this task.

PR #509, opened 2021-03-18
Don't apply Compose compiler plugin to JS and Native compilations

### LINKED ISSUE #352: Cannot compile multiplatform project with both Jetpack Compose and JS with IR compiler
I am trying to create a multiplatform library with both Compose and JS targets. With the JS target in LEGACY mode, everything compiles just fine. However, if I switch to the IR compiler, then the compilation fail with the following error:

```
e: androidx.compose.compiler.plugins.kotlin.IncompatibleComposeRuntimeVersionException: You are using an outdated version of Compose Runtime that is not compatible with the version of the Compose Compiler plugin you have installed. The compose compiler plugin you are using (version 1.0.0-alpha13) expects a minimum runtime version of 1.0.0-alpha13.
        at androidx.compose.compiler.plugins.kotlin.VersionChecker.outdatedRuntimeWithUnknownVersionNumber(VersionChecker.kt:102)
```
While the project is configured to use the latest version of Gradle, Kotlin, and Jetpack Compose, it also fails with earlier versions. I have created a sample project to reproduce the problem:

https://github.com/lucgirardin/mpp-compose-js

Some time ago, I created a thread about this issue in Slack but it did not yield any solution:

https://kotlinlang.slack.com/archives/C01D6HTPATV/p1611065954091900

It would be so cool to be able to mix Compose and JS with expect/actual construct ;-)

Thanks for listening!

Luc

AlexeyTsvetkov:
Resolves #352

The fix resolves the issue only for Kotlin compiler builds the following commit:
https://github.com/JetBrains/kotlin/commit/d023f09bd45e6b21fa341f8481521c9a21c8b924

See also https://youtrack.jetbrains.com/issue/KT-45020

### DIFF
--- a/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/TestKotlinVersion.kt
+++ b/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/TestKotlinVersion.kt
@@ -1,0 +1,8 @@
+package org.jetbrains.compose.test
+
+@Suppress("EnumEntryName")
+enum class TestKotlinVersion(val versionString: String) {
+    // __KOTLIN_COMPOSE_VERSION__
+    Default("1.4.31"),
+    V1_5_20_dev_3226("1.5.20-dev-3226")
+}
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
@@ -1,0 +1,23 @@
+package org.jetbrains.compose
+
+import org.gradle.api.provider.Provider
+import org.jetbrains.kotlin.gradle.plugin.*
+
+class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
+    override fun getCompilerPluginId(): String =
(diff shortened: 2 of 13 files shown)


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/TestKotlinVersion.kt
```diff
@@ -0,0 +1,8 @@
+package org.jetbrains.compose.test
+
+@Suppress("EnumEntryName")
+enum class TestKotlinVersion(val versionString: String) {
+    // __KOTLIN_COMPOSE_VERSION__
+    Default("1.4.31"),
+    V1_5_20_dev_3226("1.5.20-dev-3226")
```
olonho, line 7: Shalln’t some recent 1.5.0 work here as well?

olonho (approved):

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/TestKotlinVersion.kt
AlexeyTsvetkov: I was not able to find any dev builds with the relevant change https://github.com/JetBrains/kotlin/commit/8124a9546729b05dee7b8ea489fdd5b29e60ef00

I don't think it worth spending a lot of time figuring out the relevant build

PR #562, opened 2021-04-05
error message for 32 bit arch

### LINKED ISSUE #560: "Unknown OS arch: x86" when running from command line
build.gradle.kts:
```kt
println("arch is ${System.getProperty("os.arch")}")
dependencies {
    implementation(compose.desktop.currentOs)
}
```

when running gradle from the command line, all commands fail on `implementation(compose.desktop.currentOs)` due to the following error:

```
Unknown OS arch: x86
```
image

works when running from intelliJ, as `System.getProperty("os.arch")` returns `amd64` instead:

image

DetachHead:
closes #560

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/osUtils.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/osUtils.kt
@@ -36,6 +36,7 @@ internal val currentArch by lazy {
     when (osArch) {
         "x86_64", "amd64" -> Arch.X64
         "aarch64" -> Arch.Arm64
+        "x86" -> error("32-bit arch not supported: $osArch")
         else -> error("Unknown OS arch: $osArch")
     }
 }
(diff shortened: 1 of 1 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/osUtils.kt
```diff
@@ -36,6 +36,7 @@ internal val currentArch by lazy {
     when (osArch) {
         "x86_64", "amd64" -> Arch.X64
         "aarch64" -> Arch.Arm64
+        "x86" -> error("32-bit arch not supported: $osArch")
```
olonho, line 39: Not sure if x86 any different than anything else here, maybe just change "Unknown" -> "Unsupported" in branch below.

DetachHead: done

olonho (approved):

PR #566, opened 2021-04-06
Idea plugin entry point

AlexeyTsvetkov:


idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/idea/preview/PreviewEntryPoint.kt
```diff
@@ -0,0 +1,54 @@
+/*
```
olonho, line 1: Guess here Android copyright is not applicable.

idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/idea/preview/PreviewEntryPoint.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.intellij.psi.PsiElement
+import com.intellij.psi.PsiMethod
+import org.jdom.Element
+import org.jetbrains.annotations.Nls
+
+/**
+ * [EntryPoint] implementation to mark `@Preview` functions as entry points and avoid them being flagged as unused.
+ */
+class PreviewEntryPoint : EntryPoint() {
+    private var ADD_PREVIEW_TO_ENTRIES: Boolean = true
```
olonho, line 34: Why UPPER_CASE?

idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/idea/preview/PreviewEntryPoint.kt
AlexeyTsvetkov: I'm not sure.
Here's the quote from Apache 2.0 license (part 4):
```
...
b. You must cause any modified files to carry prominent notices stating that You changed the files; and
c. You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those notices that do not pertain to any part of the Derivative Works; and
...
```
So it seems, that we must both explicitly mention the files are modified and keep the original copyright notices for the derivative parts (but we can state copyright on the modifications).

olonho (approved):

PR #620, opened 2021-04-28
Support Kotlin JS targets in Compose Gradle plugin

AlexeyTsvetkov:


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -6,23 +6,48 @@
...
 
 import org.gradle.api.provider.Provider
+import org.jetbrains.compose.internal.webExt
 import org.jetbrains.kotlin.gradle.plugin.*
+import org.jetbrains.kotlin.gradle.targets.js.ir.KotlinJsIrTarget
 
 class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
     override fun getCompilerPluginId(): String =
-        "org.jetbrains.compose"
+        "androidx.compose.compiler.plugins.kotlin"
```
olonho, line 15: Why this change?

gradle-plugins/compose/build.gradle.kts
```diff
@@ -25,6 +25,7 @@ buildConfig {
     packageName = "org.jetbrains.compose"
     clsName = "ComposeBuildConfig"
     buildConfigField("String", "composeVersion", BuildProperties.composeVersion(project))
+    buildConfigField("Boolean", "isComposeWithWeb", BuildProperties.isComposeWithWeb(project).toString())
```
olonho, line 28: Why this property make sense? Shalln't we make it marked for removal.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/WebExtension.kt
```diff
@@ -0,0 +1,60 @@
...
+    @Suppress("unused")
+    fun targets(vararg targets: KotlinTarget) {
+        check(requestedTargets == null) {
+            "compose.web.targets() was already set!"
+        }
+
+        val jsIrTargets = linkedSetOf<KotlinJsIrTarget>()
+        for (target in targets) {
+            check(target is KotlinJsIrTarget) {
+                """|'${target.name}' is not a JS(IR) target:
```
olonho, line 28: So we really cannot make configuration more automated?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/WebExtension.kt
```diff
@@ -0,0 +1,60 @@
...
+            targetsToConfigure
+                ?: requestedTargets
+                ?: defaultJsTargetsToConfigure(project)
+
+        return targetsToConfigure!!
+    }
+
+    private fun defaultJsTargetsToConfigure(project: Project): Set<KotlinJsIrTarget> {
+        val mppTargets = project.mppExt?.targets?.asMap?.values ?: emptySet()
+        val jsIRTargets: Set<KotlinJsIrTarget> = mppTargets.filterIsInstanceTo(LinkedHashSet())
```
olonho, line 49: Maybe pass type argument to the right?
`val jsIRTargets = mppTargets.filterIsInstanceTo(mutableSetOf<KotlinJsIrTarget>())`

gradle-plugins/gradle.properties
```diff
@@ -7,6 +7,7 @@ kotlin.code.style=official
 #
 # __LATEST_COMPOSE_RELEASE_VERSION__
 compose.version=0.4.0-build182
+compose.with.web=false
```
olonho, line 10: Why this change is required?

eymar:
there is a need to set a certain compiler plugin flag fo js compilation.
We do something like this now:

```
tasks.named("compileKotlinJs") { it ->
    it.kotlinOptions.freeCompilerArgs += [
            "-P", "plugin:androidx.compose.compiler.plugins.kotlin:generateDecoys=true"
    ]
}
```

would it be possible to add to the plugin as well?

gradle-plugins/compose/build.gradle.kts
AlexeyTsvetkov: This is simply a build time constant explicitly telling to build `with-web`. We can check the version of course, but I would prefer not to do that, because it could lead to unexpected errors.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/WebExtension.kt
AlexeyTsvetkov: Note that it's not required, when there is one JS IR target.

As for our options:
* We cannot change a type of target via DSL, after a target is defined.  
* We can probably force IR compiler for all JS targets in a subproject by setting `kotlin.js.compiler=ir` for that particular subproject. A downside is that users will need another flag to override that behaviour.
* We can perform configuration for any defined JS IR target, even if there is more than one.
* We can also add an extension function, which would perform the necessary configuration, e.g.
```
kotlin {
     composeWeb() // or jsCompose or jsComposeWeb
     composeWeb("customTargetName")
}
```

AlexeyTsvetkov:
@eymar it's already added, just not directly, but using Support plugin API (see ComposeCompilerKotlinSupportPlugin.kt)

gradle-plugins/gradle.properties
AlexeyTsvetkov: It's not required per se. 
Our "with-web" plugin is expected to have additional properties & extensions, as we discussed previously (e.g. we discussed, that it should not be possible to add `compose.web.web` dependency with non-web plugin version).
We can implement the restrictions by literally checking a version number, but it seems more error prone and less intentional to me.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/WebExtension.kt
AlexeyTsvetkov: OK

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
AlexeyTsvetkov: `getCompilerPluginId` is used by Kotlin Gradle plugin, when building CLI args for Kotlin compiler.
Basically it transforms options, returned by `applyToCompilation`, to `plugin:$compilerId:${option.key}=${option.value}`
The compiler plugin id is defined in https://github.com/androidx/androidx/blob/androidx-main/compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/ComposePlugin.kt#L52

Our build of the Compose compiler plugin does not override the plugin id.
Previous value was incorrect.

gradle-plugins/compose/build.gradle.kts
olonho: Ack.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/WebExtension.kt
olonho: OK, let's keep it this way.

gradle-plugins/gradle.properties
olonho: Ack.

olonho (approved):

PR #638, opened 2021-05-05
Debug writer

Rsedaikin:
DebugWriterTool - simple app for getting basic info about PC/Mac when using CFD.


debug-writer/README.md
```diff
@@ -0,0 +1,4 @@
+Compose Desktop Application
```
igordmn, line 1: Don't forget to change README :)

debug-writer/README.md
```diff
@@ -0,0 +1,4 @@
...
+
+- `./gradlew run` - run application
```
igordmn, line 3: When I run on Windows:
image

Looking at the code it seems this project is just a template and we will add functionality in the future?

debug-writer/src/main/kotlin/debugwriter/Main.kt
```diff
@@ -0,0 +1,94 @@
...
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.IntSize
+import debugwriter.decoration.AppTheme
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.async
+
+private val fileName = "${System.getProperty("user.home")}/.debug-writer/debug-info.txt"
+private var output by mutableStateOf("")
+private var isReady by mutableStateOf(true)
```
igordmn, line 31: If we set this to 'false' by default then we will show Loading immediately on the first frame

debug-writer/src/main/kotlin/debugwriter/Main.kt
```diff
@@ -0,0 +1,94 @@
...
+                            )
+                            Button("Close", Modifier.weight(1f), { window.close() })
+                        }
+                    }
+                }
+            }
+        }
+
+        if (result) {
+            writeDebugInfo()
```
igordmn, line 79: To avoid launching a coroutine on every recomposition, better to write something like this:
```
...
     Button("Refresh", Modifier.weight(1f), {
         isReady = false
         GlobalScope.launch { writeDebugInfo() }
     })
...

        if (result) {
            LaunchedEffect(Unit) {
                writeDebugInfo()
            }
        }
...
private suspend fun writeDebugInfo() {
    delay(2000L)
    val result = readDebugOutput(fileName)
    output = if (result.isEmpty()) "Something went wrong and $fileName is empty." else result
    isReady = true
}
```

debug-writer/src/main/kotlin/debugwriter/Main.kt
```diff
@@ -0,0 +1,94 @@
...
+                ) {
+                    Column {
+                        Header("Click [Refresh] to refresh info or [Open file] to see the output file.")
+                        if (isReady) {
+                            TextBox(output, Modifier.weight(1f).padding(start = 30.dp, end = 30.dp))
+                        } else {
+                            Loader(Modifier.weight(1f).fillMaxWidth())
+                        }
+                        Row(modifier = Modifier.fillMaxWidth()) {
+                            Button("Refresh", Modifier.weight(1f), { writeDebugInfo() })
```
igordmn, line 61: To avoid multiple parallel writings when user clicks multiple times better to hide/disable button when isReady == true

debug-writer/src/main/kotlin/debugwriter/Main.kt
```diff
@@ -0,0 +1,94 @@
...
+import kotlinx.coroutines.async
+
+private val fileName = "${System.getProperty("user.home")}/.debug-writer/debug-info.txt"
+private var output by mutableStateOf("")
+private var isReady by mutableStateOf(true)
+
+fun main() {
+    val result = enableDebugWritingTo(fileName)
+    if (!result) {
+        output = "Failed to cteate file: $fileName"
```
olonho, line 36: "create".
Overall placement of this code here looks somewhat strange, can we have it somewhere below and have an option to write to file in the way user wants it or copy-paste to clipboard.

esafak:
Is this still being worked on?

PR #685, opened 2021-05-21
Added a `web-getting-deployed` example

chris-hatton:
To complement the `web-getting-started` example, here is a further `web-getting-deployed` example.
It shows the original _getting started_ code being minimally packaged with, and served by, a Ktor Server inside a WAR file.

While this approach applies to Kotlin/JS Apps generally and is not special to those using **Compose for Web**, it nonetheless demonstrates **Compose for Web** code being readily placed in a useful context; as part of a Client/Server Web-Application.

I'm offering this PR on the basis that it might help promote adoption; where developer-users can easily see and play with **Compose for Web** in a deployable form.


mpetuska:
Would probably make more sense to have a standalone static webapp deployment guide first (since compose is client side rendered). But most of the stuff it's the same, regardless what serves the static content.

examples/web-getting-deployed/src/jvmMain/kotlin/Server.kt
```diff
@@ -0,0 +1,44 @@
...
+                body {
+                    div { id = "root" } // Required by Compose Web
+                    script(src = "/static/$jsFileName") {}
+                }
+            }
+        }
+
+        static("/static") {
+            resource(jsFileName)
+            resource(jsFileMapName)
```
eymar, line 41: Is there a reason to add jsFileMap here? Is it needed for something to make the app work?

examples/web-getting-deployed/webapp/WEB-INF/web.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="ISO-8859-1" ?>
```
eymar, line 1: a general question about this file. I'm not familiar with it.

Is it generated by some tool/task?

examples/web-getting-deployed/build.gradle.kts
```diff
@@ -0,0 +1,76 @@
...
+    sourceSets {
+        val jsMain by getting {
+            dependencies {
+                implementation(compose.web.web)
+                implementation(compose.runtime)
+            }
+        }
+        val jvmMain by getting {
+            dependencies {
+                implementation(compose.runtime) // Can this be decoupled?
```
eymar, line 35: Is compose.runtime needed in jvmMain for this example? Doesn't it work without it?

examples/web-getting-deployed/webapp/WEB-INF/web.xml
apatrida: This is for deploying in a WAR file, an alternative way to run instead of running from `main()`

apatrida:
I looked at this PR and I think it isn't the best direction to send people.

Really this PR does:

* packages a Compose app sample into a WAR file for deployment on any app server that supports WAR files
* does so by including a KTOR application that is executed by a KTOR servlet adapter
* does the Gradle work to package the WAR file, and provides the WAR artifacts such as web.xml
* the KTOR application basically just serves static content

And this could instead just:

* make a KTOR application with a `main()` that serves the static content

For WAR deployment, you don't even need to write a servlet like this sample does, you can use the default servlet to host static content if you want a WAR file.   

I think this sample would make more sense and be simplified to just be a KTOR server (run as `main()`) that serves the static content, and maybe have it actually call an API on the KTOR server to show it interacting with the client.  Otherwise 50% of this PR is about WAR packaging that I'm not sure is something to show here.

If you are going to replace `./gradlew jsBrowserRun` with a new sample, make it do something in a way that would be typical, and maybe have it do frontend/backend interacting by having one API endpoint that fetches simple hard coded data and shows it in the Compose UI.

Otherwise, this is just really a WAR packaging of a static website sample (and not the shortest way to show that).

apatrida:
And advantage of doing a sample for client calls to a KTOR backend would show how you would handle data loading in the client which is not really documented here specifically under the web docs (you have to go figure it out from other big samples, or from another Compose target)

chris-hatton:
@apatrida Thanks for your time and considerations. On reflecting, I share your view; any example that goes to the expense of including a complete Client/Server setup like this ought to at least demonstrate bidirectional data flow through a Compose Interface; and play on any Compose-specific aspects.  Second to this; including both Client and Server as 'targets' of the same project is just a little too misleading, even for the sake of brevity: they're conceptually not the same component and shouldn't be modeled as much.  Voluntarily declining this PR - I hope to come back with something that serves the purpose better and to a higher standard.  Thanks again for everyone's feedback.

examples/web-getting-deployed/build.gradle.kts
chris-hatton: Thanks for the question: This was indeed necessary to compile, but a side-effect of having clumsily modeled the Client and Server as 'targets' of the same project - where the Compose Compiler apparently expects to find runtime support in all targets of any project it is applied to.  As noted below; this modeling is not clean and should not be held as an example.  Part of my reason for voluntarily declining the PR.

examples/web-getting-deployed/webapp/WEB-INF/web.xml
chris-hatton: This file, `web.xml`,  is part of the [WAR standard](https://en.wikipedia.org/wiki/WAR_(file_format)) (mentioned under _Content and Structure_).  In this case the content was adapted from the official [Ktor WAR packaging example](https://ktor.io/docs/war.html).

examples/web-getting-deployed/src/jvmMain/kotlin/Server.kt
chris-hatton: It's not strictly required to make the App work; but highly recommended to include in a development project.  The browser will automatically read this accompanying, same-named `.map` file alongside the JS, enabling it to show reference to the original Kotlin source where exceptions occur.

PR #704, opened 2021-05-27
Dom applier with element

Schahen:
The idea behind this commit is to prepare our code base for working with Elements rather than HTMLElements whenever it's needed.


web/core/src/jsMain/kotlin/androidx/compose/web/DomApplier.kt
```diff
@@ -119,23 +87,36 @@ class DomNodeWrapper(val node: Node) {
             node.insertBefore(child, node.childNodes[toIndex]!!)
         }
     }
+}
+
 
-    companion object {
+class DomElementWrapper(override val node: HTMLElement): DomNodeWrapper(node) {
```
eymar, line 93: Didn't you want to use Element instead of HTMLElement?

eymar (approved):

Schahen:
This one is merged

PR #706, opened 2021-05-27
Simplify TodoApp example iOS framework export

arkivanov:
It seems that we can now use "common" artifacts for iOS framework export. Don't remember what was the original issue. Checked the iOS app, it compiles and runs fine.


arkivanov:
Not sure why does it work, looks like against the docs? https://kotlinlang.org/docs/mpp-build-native-binaries.html#export-dependencies-to-binaries

olonho (approved):

PR #708, opened 2021-05-28
Make content nullable so that we won't pass even empty lambda

Schahen:
Also - no inline for HTML API


eymar (approved):

Schahen:
This merged as f10eb96

PR #710, opened 2021-05-28
[compose-web-example] Fix package issues and code cleanup

theapache64:
image

Currently, `compose-web`'s `falling_balls_with_web` example has misaligned package names which show as a warning in the IDE. 
This PR will fix those issues and also does a code cleanup.

### DIFF
--- a/examples/falling_balls_with_web/build.gradle.kts
+++ b/examples/falling_balls_with_web/build.gradle.kts
@@ -30,9 +30,9 @@ kotlin {
         }
 
         val jvmMain by getting {
-             dependencies {
+            dependencies {
                 implementation(compose.desktop.currentOs)
-             }
+            }
         }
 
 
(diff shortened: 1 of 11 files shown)


theapache64:
cc @Schahen @eymar

Schahen:
It's an old known issue on the edge of Kotlin Tooling and IDEA integration that when package name and the actual physical location (that is, the nested folders that has one-to-one correspondence with the package name, like in Java)  IDE can warn about it. 
The thing is that we still don't want neither to make example more complicated with introducing such deep nested structure nor renaming packages just for the sake of IDE being not too grumpy.  In fact. many Kotlin projects are sticking to exactly the same approach. 

Your PR helped to realize that there's still some thing that need to be tuned (like this with-web folder name, which is obsolete) so we'll improve  something. I have to close this PR but, again, thank you for your time! It was actually helpful.

theapache64:
Understood :bow:

PR #712, opened 2021-05-28
Notepad example

igordmn:


examples/notepad/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+
+dependencies {
+    implementation(compose.desktop.currentOs)
+    implementation(compose.materialIconsExtended)
+}
+
+compose.desktop {
+    application {
+        mainClass = "MainKt"
+        jvmArgs("-Dskiko.rendering.laf.global=true")
```
olonho, line 24: So we no longer set it by default?

examples/notepad/build.gradle.kts
```diff
@@ -0,0 +1,43 @@
...
+            packageVersion = "1.0.0"
+
+            windows {
+                menu = true
+                // see https://wixtoolset.org/documentation/manual/v3/howtos/general/generate_guids.html
+                upgradeUuid = "61DAB35E-17CB-43B0-81D5-B30E1C0830FA"
+            }
+
+            macOS {
+                // Use -Pcompose.desktop.mac.sign=true to sign and notarize.
```
olonho, line 38: Guess smth else shall be here.

examples/notepad/src/main/kotlin/util/AwtIcons.kt
```diff
@@ -0,0 +1,77 @@
...
+import java.awt.Point
+import java.awt.image.BufferedImage
+import java.awt.image.ColorModel
+import java.awt.image.DataBuffer
+import java.awt.image.DataBufferInt
+import java.awt.image.Raster
+import java.awt.image.SinglePixelPackedSampleModel
+import kotlin.math.roundToInt
+
+suspend fun ImageVector.toAwtImage(tintColor: Color): BufferedImage {
```
olonho, line 28: Maybe add those(non-suspend versions) to https://github.com/JetBrains/skiko/blob/master/skiko/src/jvmMain/kotlin/org/jetbrains/skiko/Convertors.kt

examples/notepad/src/main/kotlin/util/Compose.kt
```diff
@@ -0,0 +1,59 @@
...
+                ) {
+                    result = content()
+                }
+            }
+        }
+    }
+    return result as T
+}
+
+private object YieldFrameClock : MonotonicFrameClock {
```
olonho, line 40: Do we really need those?

examples/notepad/build.gradle.kts
igordmn: We don't set in by default in Compose, but can set it by default in our plugin (we never did it before).

@AlexeyTsvetkov, what you think?

examples/notepad/src/main/kotlin/util/AwtIcons.kt
igordmn: ImageVector is a Compose class, so we can add it to Compose in the future

examples/notepad/src/main/kotlin/util/Compose.kt
igordmn: Compose.kt is needed for AwtIcons.kt, that allow us to convert built-in material icons to BufferedImage

examples/notepad/build.gradle.kts
igordmn: Removed. we don't need it

examples/notepad/src/main/kotlin/util/AwtIcons.kt
igordmn: We can replace part of this code by this:
```
    CanvasDrawScope().draw(
        density, layoutDirection, canvas, intrinsicSize
    ) {
        draw(intrinsicSize, colorFilter = colorFilter)
    }

    return bitmap.asDesktopBitmap().toBufferedImage()
```

Icon in Tray is ok, but icon in titlebar is not ok. Something is wrong in `toBufferedImage` (will figure out this later).

olonho (approved):

PR #715, opened 2021-05-29
Fixed compilation of web getting-started tutorial

solidogen:
I followed the tutorial here:
https://github.com/JetBrains/compose-jb/blob/master/tutorials/Web/Getting_Started/README.md

Yet code inside doesn't compile. I checked actual code from an example and it differs:
https://github.com/JetBrains/compose-jb/blob/master/examples/web-getting-started/src/jsMain/kotlin/Main.kt

I updated tutorial with code from example so it compiles and runs correctly.

### DIFF
--- a/tutorials/Web/Getting_Started/README.md
+++ b/tutorials/Web/Getting_Started/README.md
@@ -91,23 +91,23 @@ kotlin {
 #### 7. Add the `Main.kt` file:
 ```kotlin
 fun main() {
-    var count: Int by mutableStateOf(0)
+    val count = mutableStateOf(0)
 
     renderComposable(rootElementId = "root") {
         Div(style = { padding(25.px) }) {
             Button(attrs = {
-                onClick { count -= 1 }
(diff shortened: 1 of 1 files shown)


eymar:
Hi @solidogen 
Could you please let us know what was the error when you tried to compile the code from the tutorial?

I will update the code in example to match the code in the tutorial. The code in the tutorial is more idiomatic, so the web-getting-started example should match it.

eymar:
I updated the code in the example and added a note in the tutorial:
https://github.com/JetBrains/compose-jb/tree/master/tutorials/Web/Getting_Started#7-add-the-mainkt-file

One might need to add imports:
```
import androidx.compose.runtime.getValue
import androidx.compose.runtime.setValue
```

But let us know if you had another compilation error.

olonho (commented):
Yes, this looks like missing imports.

solidogen:
Yeah, I added getValue import because I remember there were problems, but I totally forgot setValue as well. Closing, thanks

PR #725, opened 2021-06-01
Commonize CSS interfaces and remove polyfill dependency

Skolotsky:
CSS interfaces now have implementation in K/JVM (with TODO() fallback for now), in K/JS and in native Browser structures.
`useNativeCSSTypedOM` is used for switching API which used for style applying.
Removed polyfill dependency


web/core/src/commonMain/kotlin/androidx/compose/web/css/CSSTypedOM.kt
```diff
@@ -0,0 +1,162 @@
...
+@file:Suppress("UNUSED")
+package androidx.compose.web.css
+
+import kotlin.js.JsName
+
+expect interface CSSStyleValue
+
+open class CSSStyleValueKT : CSSStyleValue {
+    companion object {
+        fun parse(property: String, cssText: String): CSSStyleValue = TODO()
```
eymar, line 15: is TODO() left intentionally?

Skolotsky: yes, it isn't used yet, and i think that maybe it should be removed from interfaces now

web/core/build.gradle.kts
```diff
@@ -30,7 +30,6 @@ kotlin {
         val jsMain by getting {
             dependencies {
                 implementation(kotlin("stdlib-js"))
-                implementation(npm("css-typed-om", "0.4.0"))
```
Schahen, line 33: Can you please remove css-typed-om license from licenses as well?

web/core/src/commonMain/kotlin/androidx/compose/web/css/CSSKT.kt
```diff
@@ -0,0 +1,107 @@
+/*
```
Schahen, line 1: Not a big deal but it seems to be that this file can have better (= more intuituve) name )

web/core/src/commonMain/kotlin/androidx/compose/web/css/CSSTypedOM.kt
```diff
@@ -0,0 +1,162 @@
...
+    companion object {
+        operator fun invoke(value: kotlin.String): String
+
+        operator fun invoke(value: kotlin.Number): Number
+
+        operator fun invoke(value: CSSStyleValue): StyleValue
+    }
+}
+
+// type CSSNumberish = number | CSSNumericValue
```
Schahen, line 44: I suggest to name it `CSSNumeric`  - it's honestly equally understandable and a bit more simple English.

Schahen (commented):
Apart from three comments I left the only thing I would like you to change is package names - 
let's keep all behind `org.compos.jetbarains.web` prefix

web/core/src/commonMain/kotlin/androidx/compose/web/css/CSSTypedOM.kt
Skolotsky: I just reused name from TypeScript d.ts typing for it. But CSSNumeric is better, agree)

web/core/src/commonMain/kotlin/androidx/compose/web/css/CSSKT.kt
Skolotsky: That object in  Browser API named just `CSS`, and it is kotlin implementation of that object... So i have another name CSSKotlinImpl.kt

web/core/build.gradle.kts
Skolotsky: Sure!

web/core/src/commonMain/kotlin/androidx/compose/web/css/CSSKT.kt
Skolotsky: Do you have better naming for that?

Schahen: I usually take the "biggest" entity in the file and name the file accordingly but, again, it's no a big deal )

PR #741, opened 2021-06-03
Web: Add Hr Element

hfhbd:


web/core/src/jsMain/kotlin/androidx/compose/web/elements/Elements.kt
```diff
@@ -358,6 +359,18 @@ fun Footer(
     )
 }
 
+@Composable
+fun Hr(
+    attrs: AttrBuilderContext<Tag.Hr> = {},
+    content: ContentBuilder<HTMLHRElement>? = null
```
eymar, line 365: Is there a need for `content` for <hr> tag?

hfhbd: You are right. It is even forbidden.

hfhbd: BTW Br is also not allowed to contain a content https://developer.mozilla.org/en-US/docs/Glossary/Empty_element

eymar: Good catch! Thanks. 
Would you like to remove content from Br element in this PR as well?  That would be helpful :)

eymar: oh I see you added an issue for EmptyElements. Thanks! Then I accept this PR.

PR #742, opened 2021-06-03
Web: Fix Option usage in Select Element

hfhbd:
Currently not possible: 
```kotlin
Select {
  Option(value = "foo") {
    Text("Foo")
  }
}
```

### DIFF
--- a/web/core/src/jsMain/kotlin/androidx/compose/web/elements/Elements.kt
+++ b/web/core/src/jsMain/kotlin/androidx/compose/web/elements/Elements.kt
@@ -249,7 +249,7 @@ fun Select(
 )
 
 @Composable
-fun DOMScope<HTMLUListElement>.Option(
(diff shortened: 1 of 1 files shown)


web/core/src/jsMain/kotlin/androidx/compose/web/elements/Elements.kt
```diff
@@ -249,7 +249,7 @@ fun Select(
 )
 
 @Composable
-fun DOMScope<HTMLUListElement>.Option(
+fun DOMScope<HTMLSelectElement>.Option(
```
eymar, line 252: I think `DomScope<...>` can be omitted here at all, since `<option>` can be use not only in `<select>`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option

let's keep it as `fun Option(...)` without any scope.

PR #743, opened 2021-06-03
Web: Select: Add predefined value attribute builder

hfhbd:


web/core/src/jsMain/kotlin/androidx/compose/web/attributes/Attrs.kt
```diff
@@ -260,6 +260,9 @@ fun AttrsBuilder<Tag.Select>.required(value: Boolean = true) =
 
 fun AttrsBuilder<Tag.Select>.size(numberOfRows: Int) =
     attr("size", numberOfRows.toString())
+    
+fun AttrsBuilder<Tag.Select>.value(value: String) =
+    attr("value", value)
```
eymar, line 265: Did you want to add this attribute for `<option>`?

I'm looking at docs - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#attributes
And I don't see `value` attribute for `<select>`. Am I missing something?

hfhbd: Sorry, looks like I was confused with the very helpful [React implementation](https://reactjs.org/docs/forms.html#the-select-tag). Didn't know, it is not in the standard...

eymar: I agree, it's convenient. 
At the same time we decided to keep our composable tags as close as possible to usual html definitions.

hfhbd: Yeah, in the core module, I would expect a strict html conformance too, especially for 3 party libraries. But it would be nice to have a on top implementation in the widget module, right?

eymar: there is no "select" component in the widgets at the moment. And I'm not sure it's going to be there. The purpose of widgets module is to provide the dom based implementation of the widgets from compose desktop/android (as close as possible). 


When it comes to such a case as you mentioned for `<select value="">`, it's probably can be a part of some standalone module which doesn't exist yet. that module would include other "shortcuts" and convenient things on top of dom composables. But this is just a thought, not an actual plan.

hfhbd:
Closing, as it is not part of the HTML standard.

PR #745, opened 2021-06-03
Web: Remove content builder for empty elements

### LINKED ISSUE #744: Web: No content builder on EmptyElements
Some HTML elements are not allowed to have a body.

https://developer.mozilla.org/en-US/docs/Glossary/Empty_element

Currently, at least `Br { Text("foo") }` is possible, but should not.

_Originally posted by @hfhbd in https://github.com/JetBrains/compose-jb/pull/741#discussion_r644674888_

hfhbd:
Fix #744

Reference: https://developer.mozilla.org/en-US/docs/Glossary/Empty_element


eymar:
Thanks @hfhbd! Merging it now

PR #761, opened 2021-06-07
CSS Unit Refactoring

Schahen:
This is part one of what's needed to move forward with two (related but different) tasks:

1) get rid of CSSOM polyfill - in particular no more polyfills for any CSSUnitValue-like types
2) introduce proper airthmetic operations on css units.


examples/web_landing/src/jsMain/kotlin/com/sample/style/Stylesheet.kt
```diff
@@ -7,18 +7,18 @@ object AppCSSVariables : CSSVariables {
     val wtColorGreyLight by variable<Color>()
     val wtColorGreyDark by variable<Color>()
 
-    val wtOffsetTopUnit by variable<CSSSizeValue>()
```
eymar, line 10: This change leads to breaking the web_landing example if it's built using published artifact, doesn't it?
I don't think it's critical. Anyway it will need to get updated very soon after release.

Schahen: I think it's safe enough to swap names for CSSUnit and CSSSize -> in that case anything won't change in that aspect.
Will try after all comments on this PR will be collected )

web/core/src/jsMain/kotlin/androidx/compose/web/css/CSS.kt
```diff
@@ -87,7 +87,7 @@ interface CSSNumberish {
 }
 
 fun CSSNumberish.asNumber() = this.asDynamic() as? Number
-fun CSSNumberish.asCSSNumericValue(): CSSNumericValue? = this.asDynamic() as? CSSNumericValueJS
+fun CSSNumberish.asCSSNumericValue(): CSSNumericValue? = this.asDynamic() as? CSSNumericValue
```
eymar, line 90: I'm definitely asking this without having a broader context in mind, but I'm just curious if it can be not nullable?

Schahen: I really see no reason why it can not so let me check!

web/core/src/jsMain/kotlin/androidx/compose/web/css/CSSUnits.kt
```diff
@@ -1,114 +1,270 @@
...
+interface CSSUnitResolution: CSSUnit
+interface CSSUnitFlex: CSSUnit
+
+typealias CSSAngleValue = CSSUnitValue<CSSUnitAngle>
+
+sealed interface CSSUnit {
+    val value: String
+
+    object percent: CSSUnitRel {
+        override val value: String
```
eymar, line 31: is there a reason to declare a getter?

can't it be just like below snippet?

```
override val value: String = "%
```

Schahen: no actual reason! ) it's remnant of previous approach, so I've pushed a simplification.

eymar (approved):

Schahen:
I promise, I promise, I will use the green button!

PR #764, opened 2021-06-08
Simplify assignment signatures in CSS properties

Schahen:


Schahen:
Merge a a separate commit but I promise I will use the Big Green Button from now on )))

PR #772, opened 2021-06-11
Clone node experiment

Schahen:
The goal of this exercise initially was to substitute element creation with clone node which is allegedly faster. It's still an open question whether cloneNode is faster but it's not slower ) However, if we'll want to switch back to "regular" elementCreation it would be very cheap. This, however, helped to catch a couple of typings bugs and therefore, is already in my opinion beneficial.

### DIFF
--- a/web/core/src/jsMain/kotlin/androidx/compose/web/elements/Base.kt
+++ b/web/core/src/jsMain/kotlin/androidx/compose/web/elements/Base.kt
@@ -20,7 +20,10 @@ import org.w3c.dom.HTMLAnchorElement
 import org.w3c.dom.HTMLButtonElement
 import org.w3c.dom.HTMLDivElement
 import org.w3c.dom.HTMLElement
+import org.w3c.dom.HTMLHeadElement
+import org.w3c.dom.HTMLHeadingElement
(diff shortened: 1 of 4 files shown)


web/core/src/jsMain/kotlin/androidx/compose/web/elements/Base.kt
```diff
@@ -85,3 +185,14 @@ fun <TTag : Tag, TElement : Element> TagElement(
         refEffect.effect?.invoke(this, scope.element) ?: onDispose {}
     }
 }
+
+@Composable
+fun <TTag : Tag, TElement : Element> TagElement(
```
eymar, line 190: If there is a variant of this function with `elementBuilder: ElementBuilder<TElement>`, then what's the need for this function with `tagName: String`? 

It's not used anywhere so far, right?

eymar (approved):

web/core/src/jsMain/kotlin/androidx/compose/web/elements/Base.kt
Schahen: I can imagine situation where people are already using this for creating custom tags.

PR #778, opened 2021-06-13
Add web app for the TodoApp example

arkivanov:
This PR adds a web browser app for the TodoApp example.

There were a lot of problems with the database. SQLDelight supports JavaScript, but the [driver instantiation is asynchronous](https://cashapp.github.io/sqldelight/js_sqlite/). Also I could not get integration tests working with the real database for JavaScript. I had to refactor some database related code, so it could handle the asynchronous driver and to make it testable.

There are two commits:
1. Added `js` target and refactored the database code
2. Added the web app

For the web app I used [Materialize](https://materializecss.com/) styles and [Material Icons](https://google.github.io/material-design-icons). Also managed to add the crossfade animation for navigation. Tested on Chrome (Ubuntu) and Safari (macOS).

Disclaimer: I have little experience in web dev, so I will appreciate any hints or suggestions here.

Demo video:

https://user-images.githubusercontent.com/26204457/121821960-46cd0700-cc94-11eb-9169-7e4599d96937.mp4


olonho (approved):

eymar:
It seems the data persistence doesn't work on web.
When I added a couple of items and refreshed the page, there were no items.
Same scenario works on desktop (added 2 items and restarted the app).

Is it supposed to restore the added items on refresh/reload?

eymar:
Potentially, the crossfade animation could be implemented using [CSS animations](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Animations/Using_CSS_animations) 

As for now, we haven't provided any API to facilitate this in compose-web, therefore it doesn't make a lot of sense to invent something within this example for CSS animations. 
So this comment is rather FYI, than an action point :)

eymar:
Would it be okay for update this PR to use the yesterday's build? :) 

Unfortunately, it will require to update some places as there were some breaking changes made: https://github.com/JetBrains/compose-jb/releases/tag/0.5.0-build225-web

arkivanov:
> Would it be okay for you to update this PR to use the yesterday's build? :)
> 
> Unfortunately, it will require to update some places as there were some breaking changes made: https://github.com/JetBrains/compose-jb/releases/tag/0.5.0-build225-web

Sure, will update tonight

arkivanov:
@eymar Updated Compose to `0.5.0-build225`

eymar (approved):
Thanks!

And here is a casual question if you don't mind:
What was the most annoying thing for you when you were porting the TodoApp for compose-web?

arkivanov:
@eymar 

> What was the most annoying thing for you when you were porting the TodoApp for compose-web?

I noticed two major points:
1. I could not share the existing Compose UI, which is used in Desktop and Android apps. This is because there are very few common Compose components available as of now.
2. Writing HTML is very verbose, and requires many `attrs` to be configured.

Given the above, personally I would like to have the `foundation` and `material` components to be shared with JS. :-)

PR #779, opened 2021-06-14
Simplify elements internal api by removing Tag interface completely and rely on existing HTMLElement hierarchy

Schahen:


eymar (approved):
No questions. I see the tests passed, so approving this PR.

PR #783, opened 2021-06-16
Remove the boolean parameter from hidden() to align with #780

hfhbd:
In #780, all attributes builder functions with a boolean parameter were removed, like `disabled(true)` => `disabled()`. 
This one was left. 

@Schahen


eymar:
that's correct, here is the reference - https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden

PR #786, opened 2021-06-16
Remove CSS Typed OM API

Skolotsky:
Removed CSS Typed OM API completely, improved CSS declarations to generate less JS code
Bundle size changed 657080 -> 637200


web/benchmark-core/src/jsMain/kotlin/com/sample/style/Stylesheet.kt
```diff
@@ -20,7 +20,7 @@ object AppCSSVariables : CSSVariables {
     val wtH3FontSize by variable<CSSUnitValue>()
     val wtH3LineHeight by variable<CSSUnitValue>()
 
-    val wtColCount by variable<Int>()
+    val wtColCount by variable<StylePropertyNumber>()
```
eymar, line 23: Was it a mistake to use Int type there?

Skolotsky: Variable can have value which can have style property, now style property can be CSSStyleValue, StylePropertyString  (real String) and StylePropertyNumber  (real Number). Because we can't set it as Any, but we haven't CSSStyleValue|String|Number type in Kotlin. In real world there are not `Int` css value, there are units, even `number` unit fo simple numbers as `opacity`

Skolotsky: So it's a way to add constraints to CSS variable types, they can be only values which are allowed as style property value

Skolotsky: I will think about removing even StylePropertyNumber, to simplify type model

web/core/src/jsMain/kotlin/androidx/compose/web/css/CSSUnits.kt
```diff
@@ -1,38 +1,30 @@
...
-interface CSSUnitRel : CSSUnitLength
-interface CSSUnitAbs: CSSUnitLength
-interface CSSUnitAngle: CSSUnit
-interface CSSUnitTime: CSSUnit
-interface CSSUnitFrequency: CSSUnit
-interface CSSUnitResolution: CSSUnit
-interface CSSUnitFlex: CSSUnit
+external interface CSSUnitLengthOrPercentage: CSSUnit
+external interface CSSUnitPercentage: CSSUnitLengthOrPercentage
+external interface CSSUnitLength: CSSUnitLengthOrPercentage
```
eymar, line 20: Would "CssUnitSize" be more appropriate name?

Skolotsky: There already is such name. I didn't change naming here, so let's discuss naming further? I think we can improve it but it isn't about this PR.

eymar (approved):

PR #790, opened 2021-06-17
Feature/calc

Skolotsky:
Added API for calc
Made variable value API more applicable: `CSSVariableValueAs<T>` gives ability provide fallback type for variable value, as CSSNumericValue for CSSSizeValue to distinguish abstract value from real class value.
Now it's simpler to write API for CSS properties using CSSNumericValue / CSSColorValue instead CSSSizeValue / Color

### DIFF
--- a/web/core/src/jsTest/kotlin/CSSUnitApiTests.kt
+++ b/web/core/src/jsTest/kotlin/CSSUnitApiTests.kt
@@ -5,40 +5,7 @@
 
 package org.jetbrains.compose.web.core.tests
 
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
-import org.jetbrains.compose.web.css.ch
-import org.jetbrains.compose.web.css.cm
-import org.jetbrains.compose.web.css.cssRem
-import org.jetbrains.compose.web.css.deg
-import org.jetbrains.compose.web.css.div
-import org.jetbrains.compose.web.css.dpcm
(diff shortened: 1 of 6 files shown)


web/core/src/jsTest/kotlin/CSSUnitApiTests.kt
```diff
@@ -306,6 +273,162 @@ class CSSUnitApiTests {
...
+        val typedResultLengthRight: CSSNumericValue<CSSUnitLength> = 4.pt - (4.px + 4.em)
+        assertEquals("calc(4pt - (4px + 4em))", typedResultLengthRight.toString())
+    }
+
+    @Test
+    fun calcVaraiables() {
+        val variables = object : CSSVariables {
+            val pxVar by variable<CSSSizeValue<CSSUnit.px>>()
+        }
+        val typedResultLength: CSSNumericValue<CSSUnitLength> = 4.pt + variables.pxVar.value()
```
eymar, line 425: What do you think about adding more operators overloads to make it possible to write like this:

```
4.pt + variables.pxVar
```

In this case it needs an overload:
```
operator fun <T: CSSUnit> CSSNumericValue<out T>.plus(b: CSSStyleVariable<CSSNumericValue<out T>>): CSSCalcValue<T> = CSSCalcValue(CSSPlus(this, b.value()))
```

I would prefer to not write ".value()" if I don't need a fallback.

Skolotsky: we can do that, but it will add some ambiguity - two ways to do same things. I'll think about different cases.

eymar (approved):

Issue #635, opened 2021-05-04
Web: DateInput onInput: Event is not InputEvent


hfhbd:
If you have:
```kotlin
Input(type = InputType.DateTimeLocal, attrs = {
    onInput {
        val event = it.nativeEvent
    }
})
```
then `onInput` will always fail with a `ClassCastException`, because https://github.com/JetBrains/androidx/blob/107c4ac00bb4c99f29b203d71738ae7cd00c6c75/compose/web/src/jsMain/kotlin/androidx/compose/web/attributes/WrappedEventListener.kt#L150
will always cast it to `InputEvent`.

In Chrome and Safari, the `event` is type of `Event`, not `InputEvent`.
<img width="1920" alt="Bildschirmfoto 2021-05-04 um 19 24 00" src="https://user-images.githubusercontent.com/22521688/117044249-86f0af80-ad0e-11eb-8dd2-9f5df8897228.png">
<img width="1920" alt="Bildschirmfoto 2021-05-04 um 19 23 47" src="https://user-images.githubusercontent.com/22521688/117044263-89eba000-ad0e-11eb-880e-827fdcfd21cf.png">

Safari: Version 14.1 (16611.1.21.161.6)
Chrome: Version 90.0.4430.93 (Offizieller Build) (x86_64)
macOS: 11.3.1 (20E241)

Sample Project: https://github.com/hfhbd/ComposeTodo

eymar:
Hi @hfhbd, thanks for reporting this issue.

it's possible to workaround this problem:

```koltin
Input(
        type = InputType.DateTimeLocal,
        attrs = {
            addEventListener("input") { 
                val event = it.nativeEvent // nativeEvent is of type `Event `
            }
        }
    )
```
Let me know if this helps.

In general, we want to have accurately typed events. So we'll keep on improving it.

hfhbd:
@eymar Thanks, it helps, was to easy to find :)

eymar:
fixed in 0.5.0-build228

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-6403">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #799, opened 2021-06-18
Web Input element and onInput event refactoring

eymar:
```kotlin
Input( 
            type = InputType.File,
            attrs = {
                onInput { it -> // it's a SyntheticInputEvent. Its `value` type depends on the `InputType`
                    filePath = it.value // it's a String in this case
                }
            }
        )
```

Since the new `onInput {}` callback returns the event object with a properly typed value, there is no need to use old methods:
`onTextInput, onCheckboxInput, onRadioInput, onRangeInput`. They are deprecated now and we'll probably remove them soon.

This behaviour was achieved by creating new `InputAttrsBuilder` and `TextAreaAttrsBuilder` (they both inherit from existing AttrsBuilder).
Therefore `onInput {}` can be called now only within Input or TextArea elements.

If one needs to add "input" eventListener from elsewhere, then the only option is to add it manually:

```
Div(attrs = {
    addEventListener("input") {}
})
```

This PR also adds another breaking change:
The old `Input` used to have `value: String` parameter in the function signature: `Input(value = "something",...)`.
It's not the case now. The `value` can be set only within `attrs` section:
```
Input(..., attrs = { value("something") })
```


web/core/src/jsMain/kotlin/androidx/compose/web/DomApplier.kt
```diff
@@ -45,16 +46,31 @@ open class DomNodeWrapper(open val node: Node) {
...
 
+    private fun Options.BooleanValue.toBoolean(): Boolean? = when (this) {
+        Options.BooleanValue.True -> true
+        Options.BooleanValue.False -> false
+        Options.BooleanValue.Default -> null
+    }
+
+    // This conversion helps to preserve the default values of options parameters (defaults can vary per browser)
+    private fun Options.toJsObject(): dynamic {
+        val obj = js("{}")
```
Skolotsky, line 67: we have `jsObject` util, use it and put into appropriate place.
common use-case 
```Kotlin
jsObject<SomeParams/* external interface*/> {
  param1 = ...
  param2 = ...
}
```

web/core/src/jsMain/kotlin/androidx/compose/web/attributes/InputAttrsBuilder.kt
```diff
@@ -0,0 +1,76 @@
...
+class InputAttrsBuilder<T>(val inputType: InputType<T>) : AttrsBuilder<HTMLInputElement>() {
+
+    fun onInput(options: Options = Options.DEFAULT, f: (SyntheticInputEvent<T, HTMLInputElement>) -> Unit) {
+        addEventListener(INPUT, options) {
+            val value = inputType.inputValue(it.nativeEvent)
+            f(SyntheticInputEvent(value, it.nativeEvent.target as HTMLInputElement, it.nativeEvent))
+        }
+    }
+
+    @Deprecated(
```
Skolotsky, line 32: When we will remove deprecated API ?

web/core/src/jsMain/kotlin/androidx/compose/web/attributes/PredefinedAttrValues.kt
```diff
@@ -1,26 +1,101 @@
...
-    object Search : InputType("search")
-    object Submit : InputType("submit")
-    object Tel : InputType("tel")
-    object Text : InputType("text")
-    object Time : InputType("time")
-    object Url : InputType("url")
-    object Week : InputType("week")
+import org.w3c.dom.events.Event
+
+sealed class InputType<T>(val typeStr: String) {
```
Skolotsky, line 5: Very sad that it will generate so much js code... Will think if we can do it less "fat"

Skolotsky (changes requested):

web/core/src/jsMain/kotlin/androidx/compose/web/DomApplier.kt
eymar: great! thanks, I'll use it

hfhbd:
Looks like #635 would be fixed too. Could you add a test with `dateTime-local` input?

web/core/src/jsMain/kotlin/androidx/compose/web/attributes/InputAttrsBuilder.kt
eymar: I think we'll remove it in a couple of weeks.

eymar:
> Looks like #635 would be fixed too. Could you add a test with `dateTime-local` input?

Indeed, #635 will be fixed.
This PR also updates the test cases here: https://github.com/JetBrains/compose-jb/blob/3c28dcba99278cdd22c1281a034796efe130f77d/web/integration-core/src/jsMain/kotlin/androidx/compose/web/sample/tests/InputsTests.kt#L135

So the `DateTimeLocal ` is also covered.

hfhbd:
@eymar Nice, I failed to see it.

web/core/src/jsMain/kotlin/androidx/compose/web/DomApplier.kt
eymar: fixed this

web/core/src/jsMain/kotlin/androidx/compose/web/attributes/PredefinedAttrValues.kt
```diff
@@ -1,26 +1,101 @@
...
+    }
+
+    object Checkbox : InputType<Boolean>("checkbox") {
+        override fun inputValue(event: Event): Boolean {
+            return event.target?.asDynamic()?.checked?.unsafeCast<Boolean>() ?: false
+        }
+    }
+
+    object Color : InputType<String>("color") {
+        override fun inputValue(event: Event) = valueAsString(event)
```
Schahen, line 18: why we need to override this all the time rather than have it somewhere higher in a hierarchy?

web/core/src/jsMain/kotlin/androidx/compose/web/attributes/PredefinedAttrValues.kt
```diff
@@ -1,26 +1,101 @@
...
+
+    object Date : InputType<String>("date") {
+        override fun inputValue(event: Event) = valueAsString(event)
+    }
+
+    object DateTimeLocal : InputType<String>("datetime-local") {
+        override fun inputValue(event: Event) = valueAsString(event)
+    }
+
+    object Email : InputType<String>("email") {
```
Schahen, line 29: it's a very stupid request but can we choose some order in which we keep homogenous entities? 
Not a big deal at all but it's easier to navigate through them if they are, say, in alphabetic order.

web/core/src/jsMain/kotlin/androidx/compose/web/attributes/TextAreaAttrsBuilder.kt
```diff
@@ -0,0 +1,29 @@
...
+
+package androidx.compose.web.attributes
+
+import org.jetbrains.compose.web.attributes.*
+import org.jetbrains.compose.web.events.WrappedTextInputEvent
+import org.w3c.dom.HTMLTextAreaElement
+
+class TextAreaAttrsBuilder : AttrsBuilder<HTMLTextAreaElement>() {
+
+    fun onInput(options: Options = Options.DEFAULT, f: (SyntheticInputEvent<String, HTMLTextAreaElement>) -> Unit) {
```
Schahen, line 14: let's have some more obvious param names or discuss why "f" would be the best choice here )

web/core/src/jsMain/kotlin/androidx/compose/web/elements/Elements.kt
```diff
@@ -357,17 +350,30 @@ fun A(
...
+    type: InputType<K>,
+    attrs: InputAttrsBuilder<K>.() -> Unit
 ) {
     TagElement(
         elementBuilder = ElementBuilder.Input,
         applyAttrs = {
-            type(type)
-            value(value)
-            attrs()
+            val iab = InputAttrsBuilder(type)
```
Schahen, line 360: let's not try to save bytes and call it al least attreBuilder )

web/core/src/jsTest/kotlin/elements/EventTests.kt
```diff
@@ -36,6 +39,165 @@ class EventTests {
...
+        composition {
+            Button(
+                {
+                    onClick { handled1 = true }
+                    onClick { handled2 = true }
+                }
+            ) {}
+        }
+
+        assertEquals(1, root.childElementCount)
```
Schahen, line 55: I mean, it's a test,  it still will fail if we don't have element for some reason - let's discuss whether we need this check at all - it's not like there's smartcast somewhere later on.

web/core/src/jsTest/kotlin/elements/EventTests.kt
```diff
@@ -36,6 +39,165 @@ class EventTests {
...
+        var count = 0
+        var composedTimes = 0
+
+        var recomposeScope: RecomposeScope? = null
+
+        composition {
+            recomposeScope = currentRecomposeScope
+            composedTimes++
+
+            Button(attrs = { onClick(options = Options(once = Options.BooleanValue.True)) { count++ } })
```
Schahen, line 111: From what I see this PR is about two things, not one - we are introducing syntax for one-time events for instance. I don't think it's safe enough to shove in two different big changes in one PR.  But let's discuss

web/integration-core/src/jsMain/kotlin/androidx/compose/web/sample/Sample.kt
```diff
@@ -258,23 +259,22 @@ fun MyInputComponent(text: State<String>, onChange: (String) -> Unit) {
...
                 onKeyUp {
                     println("On keyUp key = : ${it.getNormalizedKey()}")
                 }
             }
         )
     }
-    Div(
-        attrs = {
+    Div {
+        Input(type = InputType.Checkbox, attrs = {
```
Schahen, line 270: I suggest us to have standard definition for most popular inputs - I can clearly see one it will be more intuitive to write something like

```RadioInput({...})```

Schahen (commented):
It's only first iteration of reviewing things. Just want to move in small steps forward.
The job done is impressive, it's just that those changes are quite big and we should not rush.

web/core/src/jsMain/kotlin/androidx/compose/web/attributes/PredefinedAttrValues.kt
eymar: Aren't they in alphabetic order now? :)

web/integration-core/src/jsMain/kotlin/androidx/compose/web/sample/Sample.kt
eymar: Added them in `InputElements.kt` file.
Maybe we also want to add more attributes to the functions signatures to let set the attributes more easily.

web/core/src/jsTest/kotlin/elements/EventTests.kt
eymar: that's right.
the `Options` changes are now not a part of this PR

web/core/src/jsMain/kotlin/androidx/compose/web/elements/Elements.kt
eymar: fixed

web/core/src/jsMain/kotlin/androidx/compose/web/attributes/TextAreaAttrsBuilder.kt
eymar: fixed

web/core/src/jsMain/kotlin/androidx/compose/web/attributes/PredefinedAttrValues.kt
eymar: added an improvement for this.
let me know what you think

Issue #153, opened 2020-11-30
Support for animated gif images.


sureshg:
Trying to display the following animated gif image displays only the first frame.

```
Image(
     bitmap = imageFromResource("lottie-sample.gif"),
     modifier = Modifier.align(Alignment.BottomCenter).preferredSize(100.dp,100.dp)
)
```
![lottie-sample](https://user-images.githubusercontent.com/356994/100579048-4e006a80-3298-11eb-8ea0-a7205221f389.gif)

olonho:
Not sure if animated GIFs work even on Android. Maybe there's third-party library doing the same?

jimgoog:
it's worth noting that gifs are hugely bandwidth/latency inefficient (https://rigor.com/blog/optimizing-animated-gifs-with-html5-video/) which becomes a much bigger consideration for mobile devices.

For an animation like this one in particular, it might be worth generating/drawing the bubble explosion directly rather than having a pre-generated resource file (something like https://github.com/adibfara/composeclock).  That would allow you to programmatically control things like animation velocity, number of bubbles, etc.  It would also allow you to have an animation that could run continuously/forever without jumping, one that visually doesn't ever repeat, etc.  Just food for thought.

Dominaezzz:
This seems interesting I'll try and make a sample of this.

Dominaezzz:
This is a sample of naive GIF rendering with compose.

https://gist.github.com/Dominaezzz/cd51f8821162a149ee2a5fb69a702e7f

(Can PR a sample if maintainers are interested)

sureshg:
@Dominaezzz awesome.. Compiling it on JDK 15 gives the following error even though JDK has the `java.xml` module
```
Unresolved reference: NodeList
```
Should I explicitly add the XML API dependencies for `import org.w3c.dom.*` ?

Dominaezzz:
Ha, I guess so, I ran it with JDK 11 and the IntelliJ compose project setup and it just worked.

sureshg:
It's really weird that only fully qualified `NodeList` name is working with the compose plugin. Here is the Gradle compilation logs

```2020-12-03T00:20:05.814-0800 [DEBUG] [org.gradle.api.Task] v: Using Kotlin home directory <no_path>
2020-12-03T00:20:05.814-0800 [DEBUG] [org.gradle.api.Task] v: Configuring the compilation environment
[...]

PR #802, opened 2021-06-21
[WIP] Gif Renderer

Dominaezzz:
Add component for https://github.com/JetBrains/compose-jb/issues/153


olonho:
Do not think that this is GIF-only renderer, think it provide support for wider set of animations supported by Skia. So better rename it.

components/GifRenderer/common/src/desktopMain/kotlin/org/jetbrains/compose/gifrenderer/DesktopGifRenderer.kt
```diff
@@ -0,0 +1,50 @@
...
+import androidx.compose.ui.graphics.painter.Painter
+import org.jetbrains.skija.Bitmap
+import org.jetbrains.skija.Codec
+import org.jetbrains.skija.Data
+import java.net.URL
+
+@Composable
+internal actual fun rememberGifRendererImpl(url: String): State<Painter> {
+    val codec = remember(url) {
+        println("Gif renderer for $url")
```
olonho, line 16: Avoid debug printouts.

components/GifRenderer/common/src/desktopMain/kotlin/org/jetbrains/compose/gifrenderer/DesktopGifRenderer.kt
```diff
@@ -0,0 +1,50 @@
...
+        val bytes = URL(url).readBytes()
```
olonho, line 17: Use IO dispatcher to fetching data from network.

components/GifRenderer/common/src/commonMain/kotlin/org/jetbrains/compose/gifrenderer/GifRenderer.kt
```diff
@@ -0,0 +1,12 @@
+package org.jetbrains.compose.gifrenderer
+
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.State
+import androidx.compose.ui.graphics.painter.Painter
+
+@Composable
+fun rememberGifRenderer(url: String): State<Painter> {
```
igordmn, line 8: I think we need to provide a way to load gif from any `InputStream` and from any non-composable function (that can be called from `Dispatchers.IO`)

```
fun loadAnimation(inputStream: InputStream): Animation

// load synchronously from resources (see https://android-review.googlesource.com/c/platform/frameworks/support/+/1576433/2/compose/ui/ui/src/desktopMain/kotlin/androidx/compose/ui/res/DesktopSvgResources.desktop.kt#71)
@Composable
fun animationResource(resourcePath: String)

@Composable
fun Animation(animation: Animation) {

}

interface Animation {
  val frames: Int
  fun painterFor(frame: Int): Painter
}
```

P.S. `InputStream` is jvm-only, so it should be part of the `desktopMain`. 
Maybe we can implement something using https://github.com/Kotlin/kotlinx-io, but it is experimental, so our function should be experimental too.

components/GifRenderer/common/src/desktopMain/kotlin/org/jetbrains/compose/gifrenderer/DesktopGifRenderer.kt
igordmn: I think we can provide an universal function to load from any InputStream (see above). But we shouldn't decide for the users how they want to load an animation - synchronously or asynchronously.

So I suggest not to introduce the Composable API for loading from URL. In core Compose we also didn't introduce API for asynchronous loading (to be specific, we introduced it in the past, but removed [later](https://android-review.googlesource.com/c/platform/frameworks/support/+/1561297))

components/GifRenderer/common/src/desktopMain/kotlin/org/jetbrains/compose/gifrenderer/DesktopGifRenderer.kt
```diff
@@ -0,0 +1,50 @@
...
+    val frameIndex by transition.animateValue(
+        initialValue = 0,
+        targetValue = codec.frameCount - 1,
+        Int.VectorConverter,
+        animationSpec = infiniteRepeatable(
+            animation = keyframes {
+                durationMillis = 0
+                for ((index, frame) in codec.framesInfo.withIndex()) {
+                    index at durationMillis
+                    durationMillis += frame.duration
```
YeungKC, line 41: > If the delay time of a frame is less than 20ms, mainstream browsers will set 100ms by default.
https://github.com/flutter/flutter/issues/29130#issuecomment-897732165

if use https://mixin-images.zeromesh.net/rtCKv3Mng6KY7ydK9KhKJfD7NL1QkqSsBnMd-u4sS74XId7cjLBiWPVvIrJucKOhyp0tg3YZhqTM3t5hAEoAtRU will found, only first one frame have a duration, other frame’s duration is zero.

igordmn:
Closed as there is no activity. We still plan to make it as a separate component in https://github.com/JetBrains/compose-jb/tree/master/components, when we find time to do it.

PR #809, opened 2021-06-23
Release 226 workaround

Schahen:


Schahen:
Merged as a separate commit (since we don't want to update core version alongside with this particular changes)

PR #810, opened 2021-06-23
CSS animations

Skolotsky:
Introduce css animations like this
```kotlin
object AppStyleSheet : StyleSheet() {
   val bounce by keyframes {
        from {
            property("transform", "translateX(50%)")
        }

        to {
            property("transform", "translateX(-50%)")
        }
    }

    val myClass by style {
        animation(bounce) {
            duration(2.s)
            timingFunction(AnimationTimingFunction.EaseIn)
            direction(AnimationDirection.Alternate)
        }
    }
}
```

### DIFF
--- a/web/core/src/jsMain/kotlin/androidx/compose/web/css/CSSProperties.kt
+++ b/web/core/src/jsMain/kotlin/androidx/compose/web/css/CSSProperties.kt
@@ -237,3 +237,70 @@ fun StyleBuilder.padding(value: CSSNumeric) {
     // padding hasn't Typed OM yet
     property("padding", value)
 }
+
+@Suppress("EqualsOrHashCode")
+data class CSSAnimation(
+    val keyframesName: String,
+    var duration: List<CSSSizeValue<out CSSUnitTime>>? = null,
+    var timingFunction: List<AnimationTimingFunction>? = null,
+    var delay: List<CSSSizeValue<out CSSUnitTime>>? = null,
+    var iterationCount: List<Int?>? = null,
+    var direction: List<AnimationDirection>? = null,
+    var fillMode: List<AnimationFillMode>? = null,
+    var playState: List<AnimationPlayState>? = null
+) : CSSStyleValue {
+    override fun toString(): String {
+        val values = listOfNotNull(
+            keyframesName,
+            duration?.joinToString(", "),
+            timingFunction?.joinToString(", "),
+            delay?.joinToString(", "),
+            iterationCount?.joinToString(", ") { it?.toString() ?: "infinite" },
+            direction?.joinToString(", "),
+            fillMode?.joinToString(", "),
+            playState?.joinToString(", ")
+        )
+        return values.joinToString(" ")
+    }
+}
+
+inline fun CSSAnimation.duration(vararg values: CSSSizeValue<out CSSUnitTime>) {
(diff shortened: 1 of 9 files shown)


web/core/src/jsMain/kotlin/androidx/compose/web/css/CSSProperties.kt
```diff
@@ -237,3 +237,70 @@ fun StyleBuilder.padding(value: CSSNumeric) {
     // padding hasn't Typed OM yet
     property("padding", value)
 }
+
+@Suppress("EqualsOrHashCode")
+data class CSSAnimation(
```
eymar, line 242: I see that you usually avoid using `data` classes. 
Is it necessary in this case?

web/core/src/jsMain/kotlin/androidx/compose/web/css/StyleSheet.kt
```diff
@@ -39,17 +40,21 @@ class CSSRulesHolderState : CSSRulesHolder {
  * ```
  */
 open class StyleSheet(
-    private val rulesHolder: CSSRulesHolder = CSSRulesHolderState()
+    private val rulesHolder: CSSRulesHolder = CSSRulesHolderState(),
+    val usePrefix: Boolean = true,
```
eymar, line 44: I didn't find/see any cases when `usePrefix = false`.
What's the idea behind this parameter? When would one want to pass `false`?

web/core/src/jsMain/kotlin/androidx/compose/web/css/CSSProperties.kt
```diff
@@ -237,3 +237,70 @@ fun StyleBuilder.padding(value: CSSNumeric) {
...
+            iterationCount?.joinToString(", ") { it?.toString() ?: "infinite" },
+            direction?.joinToString(", "),
+            fillMode?.joinToString(", "),
+            playState?.joinToString(", ")
+        )
+        return values.joinToString(" ")
+    }
+}
+
+inline fun CSSAnimation.duration(vararg values: CSSSizeValue<out CSSUnitTime>) {
```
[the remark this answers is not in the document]
Skolotsky: It was intentionally because it isn't contract, just convenient way to modify structure. Ant it is very similar to how other properties added to CSS.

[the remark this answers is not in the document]
Skolotsky: You don't need chain here because it's intended to be called in 
```kotlin
animation(keyframe) {
   delay()
   duration()
}
```
So it's just a way how we are doing our DSL

web/core/src/jsMain/kotlin/androidx/compose/web/css/CSSProperties.kt
Skolotsky: It is necessary due to equals check. We discussed with @Schahen that later we will think about replacing data classes with some interface which needs equals to be implemented.

web/core/src/jsMain/kotlin/androidx/compose/web/css/StyleSheet.kt
Skolotsky: It's for the cases when user want integrate with some third party lib and it's classes.

PR #815, opened 2021-06-24
Add possibility to run integration tests in multiple browsers

Schahen:


web/integration-core/src/jvmTest/kotlin/org/jetbrains/compose/web/tests/integration/common/BaseIntegrationTests.kt
```diff
@@ -1,19 +1,78 @@
...
+@ParameterizedTest(name = "{displayName} [{0}]")
+@MethodSource("resolveDrivers")
+annotation class ResolveDrivers
+
+@DisplayNameGeneration(DisplayNameSimplifier::class)
+@ExtendWith(value = [StaticServerSetupExtension::class])
+abstract class BaseIntegrationTests() {
+    companion object {
+        @JvmStatic
+        fun resolveDrivers(): Array<Array<Any>> {
```
eymar, line 71: I'm just curious why does it need to be Array of Arrays? Is it a junit 5 requirement?

Schahen: It can not, you basically need to pass it this way in order to make parametrized test working. Each of subarrays is list of parameters for this particular call.

PR #820, opened 2021-06-25
Automatically add `-Dcompose.application.configure.swing.globals=true to run/package tasks

igordmn:
When users run/package application we will configure Compose to work as standalone application.

It means we will override some global Swing properties in Compose initialization function:
- sets system property `apple.laf.useScreenMenuBar` to true
- sets system property `sun.java2d.uiScale`/`sun.java2d.uiScale.enabled` automatically on Linux
- sets `UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName())`

When users don't use those tasks (when they use Compose in an existing Swing application, or in an IDEA plugin),
this property will not be set, and so we will not override Swing globals.

If users don't want to override Swing's globals, they can change the default behavior in build.gradle:
`jvmArgs("-Dcompose.application.configure.swing.globals=false")`

See [androidx CL](https://android-review.googlesource.com/c/platform/frameworks/support/+/1748236)

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureApplication.kt
@@ -24,6 +24,8 @@ import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
 import java.io.File
 import java.util.*
 
+private val defaultJvmArgs = listOf("-Dcompose.application.configure.swing.globals=true")
+
 // todo: multiple launchers
 // todo: file associations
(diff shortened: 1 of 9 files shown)


olonho (commented):
Shalln't we do the same for native distribution?

igordmn:
> Shalln't we do the same for native distribution?
We done this. The task "package" is for native distribution

olonho (approved):

AlexeyTsvetkov (commented):
I think it would be useful to check, that the property is set, in tests (from within the test process)

igordmn:
> I think it would be useful to check, that the property is set, in tests (from within the test process)
Great idea! Done

PR #828, opened 2021-06-28
Web: Use typed Element in ref to remove cast in usage

hfhbd:


web/core/src/jsMain/kotlin/androidx/compose/web/attributes/AttrsBuilder.kt
```diff
@@ -12,7 +12,7 @@ open class AttrsBuilder<TElement : Element> : EventsListenerBuilder() {
     val styleBuilder = StyleBuilderImpl()
 
     val propertyUpdates = mutableListOf<Pair<(Element, Any) -> Unit, Any>>()
```
hfhbd, line 14: Why does `propertyUpdates` not use `TElement` too?

Issue #334, opened 2021-01-29
SplitPane component


igordmn:
Consider implementing [SplitPane](https://docs.oracle.com/javase/tutorial/uiswing/components/splitpane.html) component available as a maven artifact:
https://github.com/JetBrains/compose-jb/tree/master/components

P.S. we implemented non-ideal version of it in:
https://github.com/JetBrains/compose-jb/blob/9e67cfc916f696220d630ceb6e5b67e9570777a3/examples/codeviewer/common/src/commonMain/kotlin/org/jetbrains/codeviewer/ui/CodeViewerView.kt#L56

See also:
https://github.com/JetBrains/compose-jb/issues/148

igordmn:
@LAP2 implemented SplitPane [here](https://github.com/JetBrains/compose-jb/pull/355)

Thanks!

Things that we need to do or think about:

1. Is it right that we change size in percent's, not in pixels?
Usual desktop programs don't work this way.
Columns in desktop apps have fixed size, and only change their size when it is not enough space.

(it is important and we need to think about it before publication)

2. How can we implement multiple columns?
Now we can put one SplitPane into another.
Is there a better way? In which cases nested SplitPane can cause undesirable behavior?

3. Is API for defining custom splitter minimal and powerful?

4. Write tests:
- try to coverage all cases in all if's in SplitPane.kt
- write tests for recomposition with different arguments for Composable

5. Configure maven publication

6. Try to use SplitPane in [codeviewer](https://github.com/JetBrains/compose-jb/tree/master/examples/codeviewer), [issues](https://github.com/JetBrains/compose-jb/tree/master/examples/issues) and [widgetsgallery](https://github.com/JetBrains/compose-jb/tree/master/examples/widgetsgallery).

altavir:
Just tried it. Several issues:
* Gives an error if I set Modifier.fillMaxSize() and minimum size of an element at the same time.
* The drag shape is missing.
* Minimum width of the first element is applied to the second element as well for some reason

SerVB:
I've just taken a look at the SplitPane usage. It's cool that it already works, thanks! At the same time, I have some ideas how it can be improved. Should I write it here or in separate tickets? For now, writing here.

> Is it right that we change size in percent's, not in pixels? Usual desktop programs don't work this way.

We have the whole window divided by SplitPane and it's a bit unexpected to see that the percent is persisted when the whole window is resized: both halves are resized.

[...]

PR #829, opened 2021-06-29
Desktop. Publish SplitPane component, refactor components structure

igordmn:
- publish SplitPane (https://github.com/JetBrains/compose-jb/issues/334)

- refactor components structure (desktop -> demo, common -> library)

- convert line endings in SplitPane to "\n"


igordmn:
Example of published components: 0.0.0-newci-316

components/buildSrc/src/main/kotlin/GradleUtils.kt
```diff
@@ -0,0 +1,19 @@
...
+import org.gradle.api.Project
+import org.gradle.api.publish.maven.MavenPublication
+import java.lang.UnsupportedOperationException
+
+inline fun <reified T> Project.configureIfExists(fn: T.() -> Unit) {
+    extensions.findByType(T::class.java)?.fn()
+}
+
+var MavenPublication.mppArtifactId: String
+    get() = throw UnsupportedOperationException()
```
olonho, line 15: Why so?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -149,6 +151,10 @@ class ComposePlugin : Plugin<Project> {
         }
     }
 
+    object DesktopComponentsDependencies {
+        val splitPane = composeDependency("org.jetbrains.compose.components:components-splitpane")
```
olonho, line 155: Cool!

olonho (approved):
Great change, thanks!

components/buildSrc/src/main/kotlin/GradleUtils.kt
igordmn: We don't use this, and implementation would not be trivial

PR #830, opened 2021-06-29
Compose as submodule, scripts for build/test Compose

igordmn:
- add run configurations for IDEA / Android Studio
- configure CI to use those scripts to build everything: `0.0.0-newci-316` (with desktop/web components)


compose/.run/androidx.benchmark.integration-tests.macrobenchmark-target.run.xml
```diff
@@ -0,0 +1,60 @@
+<component name="ProjectRunConfigurationManager">
```
igordmn, line 1: These configurations are predefined in Android Studio, but we hide them under the directory `android`

.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "compose/frameworks/support"]
+	path = compose/frameworks/support
+	url = https://github.com/JetBrains/androidx
+[submodule "compose/golden"]
+	path = compose/golden
+	url = https://android.googlesource.com/platform/frameworks/support-golden
+[submodule "compose/external/doclava"]
```
olonho, line 7: Why not https://github.com/JetBrains/doclava-mirror? It have simplified JDK8 tools.jar story.

compose/.run/androidx.benchmark.integration-tests.macrobenchmark-target.run.xml
olonho: What's the point of having those under VC?

compose/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+Composite build of [Compose-jb sources](https://github.com/JetBrains/androidx)
+
+## Download submodules after downloading the main project:
+```
+git submodule update --init
+```
+Set this property to always update submodules on git checkout/pull/reset:
+```
+git config --global submodule.recurse true
```
olonho, line 11: Shall it indeed be global, cannot we make it project local?

compose/README.md
```diff
@@ -0,0 +1,61 @@
...
+  - CMake 3.10.2.4988404 (in folder $androidSdk/cmake, not in $androidSdk/cmake/$version)
+
+## Requirements to develop in IDE
+- Android Studio Arctic Fox | 2020.3.1 Canary 15
+- Custom Gradle 7.0 specified in `Settings -> Build, Execution, Deployment -> Build Tools -> Gradle` (because Android Studio will pick the wrong Gradle in the subproject instead of the Gradle in the root project)
+- Specified Gradle JDK 11 in `... -> Build Tools -> Gradle`
+- Environment variables:
+```
+export ALLOW_PUBLIC_REPOS=1
+export JAVA_TOOLS_JAR=$PWD/external/tools.jar
```
olonho, line 29: Are `COMPOSE_DESKTOP_GITHUB_BUILD=true` and `OUT_DIR=$PWD/out` intentionally omitted?

compose/build.gradle.kts
```diff
@@ -0,0 +1,83 @@
...
+fun Task.dependsOnComposeTask(name: String) = dependsOn(composeBuild.task(name))
+
+gradle.startParameter.excludedTaskNames.add(":support:compose:ui:ui-inspection:assembleInspectorJarRelease")
+gradle.startParameter.excludedTaskNames.add(":support:compose:ui:ui-inspection:dexInspectorRelease")
+
+// To show all projects which use `xxx` task, run:
+// ./gradlew -p frameworks/support help --task xxx
+
+tasks.register("publishComposeJb") {
+    dependsOnComposeTask(":compose:compiler:compiler:publishMavenPublicationToMavenRepository")
```
olonho, line 11: Can we maintain this list in androidx repo, so that when refactoring they could be updated?

compose/scripts/downloadAndroidSdk
```diff
@@ -0,0 +1,48 @@
...
+    mkdir -p $1
+    pushd $1
+    git init
+    git config advice.detachedHead false
+    git fetch --depth=1 $2 $3
+    git checkout FETCH_HEAD
+    popd
+}
+
+downloadLinuxSDK() {
```
olonho, line 15: Cool!

compose/scripts/publishExtendedIcons
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+cd "$(dirname "$0")"
+. ./prepare
+
+pushd ..
+./gradlew publishComposeJbExtendedIcons $COMPOSE_DEFAULT_GRADLE_ARGS --max-workers=1 "$@" || exit 1
```
olonho, line 7: Why separate script just for that?

compose/.run/androidx.benchmark.integration-tests.macrobenchmark-target.run.xml
igordmn: Are you asking why even have the desktop configurations under VC?

They hard to find in the usual Gradle tasks.

When we develop in IDE - it is useful to have the most popular tasks in IDEA run configurations

compose/scripts/testDesktop
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+cd "$(dirname "$0")"
+. ./prepare
+
+pushd ..
+./gradlew testComposeJbDeskto $COMPOSE_DEFAULT_GRADLE_ARGS "$@" || exit 1
```
olonho, line 7: testComposeJbDeskto*p*

compose/README.md
igordmn: Unfortunately I haven't found a way to

compose/README.md
igordmn: Yes, we keep the default androidx structure, so everything work out of the box.

compose/build.gradle.kts
igordmn: We can't entirely, because it contains js tasks.

We can do it only for desktop, but I don't know what is the right place where it should be defined (maybe in root androidx/build.gradle, maybe in buildSrc).

Besides, most certainly we will be changing this ourselves, so better to keep this in compose-jb.

If androidx decides to publish MPP artifacts, they probably will write the scripts in a different way.

compose/scripts/publishExtendedIcons
igordmn: - it is very heavy task (on CI: 20min - publishComposeJbExtendedIcons, 6min - publish).
- it shouldn't be run in parallel (it requires a lot of RAM)
- Compose can work without extended icons. So on the local machine we don't have to wait 20min to build icons which we don't use in the local project

compose/scripts/testDesktop
igordmn: Done

.gitmodules
igordmn: tools.jar is not only needed for doclava, but also for [compiler-hosted](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/compiler/compiler-hosted/integration-tests/build.gradle;l=44;drc=4c006d02dcb96771d6948918b56031ed20830f28)

olonho: So we could reused for it as well, couldn't we?

compose/README.md
olonho: I meant just `git config submodule.recurse true`

.gitmodules
igordmn: We can, but instead of  `JAVA_TOOLS_JAR` we will come with `COMPOSE_DESKTOP_GITHUB_BUILD` again, which is the same thing, just not default for androidx :).

Besides we should publish CL to Google, which takes mind/time resources :)

compose/build.gradle.kts
olonho: Ok.

compose/scripts/publishExtendedIcons
olonho: Thanks, makes sense.

olonho (approved):

compose/README.md
igordmn: It only changes the local config for the local repo, if I checked correctly.

PR #831, opened 2021-06-29
Update tutorials check task to support individual targets checks

eymar:
This change will help investigate and fix inconsistencies in the tutorials a little faster since it's easier to run individual checks (for web/desktop or all)

`./gradlew check -PCHECK_TARGET=all` - `all` is the default, so can be omitted
`./gradlew check -PCHECK_TARGET=desktop`
`./gradlew check -PCHECK_TARGET=web`

This PR also changes the command used for web tutorials: `build` -> `compileKotlinJs` (it should be enough). It saves ~1 minute when running the check for all targets.

### DIFF
--- a/tutorials/checker/build.gradle.kts
+++ b/tutorials/checker/build.gradle.kts
@@ -50,16 +50,16 @@ fun cloneTemplate(template: String, index: Int, content: String): File {
   return tempDir
 }
 
-fun checkDirs(dirs: List<String>, template: String) {
+fun checkDirs(dirs: List<String>, template: String, buildCmd: String = "build") {
   val snippets = findSnippets(dirs)
   snippets.forEachIndexed { index, snippet ->
     println("process snippet $index at ${snippet.file}:${snippet.lineNumber} with $template")
     snippet.tempDir = cloneTemplate(template, index, snippet.content)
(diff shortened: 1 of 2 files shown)


olonho (approved):

tutorials/checker/build.gradle.kts
```diff
@@ -78,18 +78,48 @@ fun checkDirs(dirs: List<String>, template: String) {
 // NOTICE: currently we use a bit hacky approach, when "```kotlin" marks code that shall be checked, while "``` kotlin"
 // with whitespace marks code that shall not be checked.
 tasks.register("check") {
+  val checks = CheckSpec.createCheckSpecs(
+    checkTargets = (project.property("CHECK_TARGET")?.toString() ?: "all").toLowerCase()
```
olonho, line 82: toLowerCase is obsolete in Kotlin 1.5

eymar: Apparently , even latest gradle 7.1 is using kotlin stdlib 1.4.31 where `lowercase()` function is experimental, so it requires adding extra @OptIn annotation. I think we'll change it to `lowercase()` as soon as gradle starts using kotlin stdlib 1.5 or newer. 

Screenshot 2021-06-29 at 16 38 40

PR #852, opened 2021-07-05
Fix preview scaling

### LINKED ISSUE #837: Desktop. Preview. Graphical artifacts when the system DPI is fractional
Compose 0.0.0-custom-build322, Windows 10, System scale 1.25

1. Run IDEA with Compose IDEA plugin (`./gradlew runIde` in idea-plugin)
2. Open example `idea-plugin\examples\desktop-project`
3. Start preview
4. Result:
image
On the screenshot we can see that the image was scaled, and not rendered pixel-top-pixel

AlexeyTsvetkov:
Resolves #837

### DIFF
--- a/idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/PreviewStateService.kt
+++ b/idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/PreviewStateService.kt
@@ -34,7 +34,7 @@ class PreviewStateService : Disposable {
             val frameConfig = FrameConfig(
                 width = c.width,
                 height = c.height,
-                scale = null
+                scale = c.graphicsConfiguration.defaultTransform.scaleX
             )
             previewManager.updateFrameConfig(frameConfig)
         }


igordmn (approved):

idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/PreviewStateService.kt
```diff
@@ -34,7 +34,7 @@ class PreviewStateService : Disposable {
             val frameConfig = FrameConfig(
                 width = c.width,
                 height = c.height,
-                scale = null
+                scale = c.graphicsConfiguration.defaultTransform.scaleX
```
olonho, line 37: What will happen if we'll move IDE window to display with different DPI?

olonho (approved):

idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/PreviewStateService.kt
AlexeyTsvetkov: I have no idea. @igordmn @Rsedaikin do you guys have any ideas, if this is a problem with `graphicsConfiguration.defaultTransform.scaleX` Swing API?

AlexeyTsvetkov: Seems it should work, since swing should trigger resize event for all components (as @igordmn said during today's sync). I want to merge this PR now, will check DPI change later.

igordmn: I checked, it works

PR #854, opened 2021-07-05
Preview toolwindow improvements

AlexeyTsvetkov:


olonho (approved):

idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/RunPreviewAction.kt
```diff
@@ -14,16 +14,20 @@ import com.intellij.openapi.externalSystem.service.execution.ProgressExecutionMo
...
+import javax.swing.SwingUtilities
 
 class RunPreviewAction(
     private val fqName: String,
     private val modulePath: String
 ) : AnAction({ "Show non-interactive preview" }, PreviewIcons.RUN_PREVIEW) {
     override fun actionPerformed(e: AnActionEvent) {
         val project = e.project!!
+        val previewToolWindow = ToolWindowManager.getInstance(project).getToolWindow("Desktop Preview")
+        previewToolWindow?.setAvailable(true)
```
igordmn, line 29: Maybe we call it in `onSuccess`? Until then preview will be hidden and user will se running action in the statusbar

PR #857, opened 2021-07-06
Add a workaround for compose js targets to use compose with kotlinx.serialization

eymar:
Add a workaround for compose js targets to use compose with kotlinx.serialization

The workaround is to force the order of the plugins: compose compiler plugin will be executed in the end.

To enable the workaround `FIX_KOTLINX_SERIALIZATION_COMPOSE_ORDER_IN_JS` needs to be set in gradle.properties:
FIX_KOTLINX_SERIALIZATION_COMPOSE_ORDER_IN_JS=true


olonho:
Don't think we shall commit it. Maybe better add detection for cases where serialization plugin is applied and suggestion to split modules.

eymar:
Closing this, as we will proceed with a warning message: https://github.com/JetBrains/compose-jb/pull/900

PR #859, opened 2021-07-06
Web: Update autocomplete attr builder according to MDN

hfhbd:
https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete


eymar:
Good point! thanks a lot.

Would you like to try add more strict type for the value (instead of the String)?
```kotlin
fun AttrsBuilder<HTMLTextAreaElement>.autoComplete(value: String)
```

If you decided to do that, I think a good reference can be found here:
[external interface StylePropertyEnum](https://github.com/JetBrains/compose-jb/blob/e2dd2562355edf5a63e156e7a7a8bdbb050b102d/web/core/src/jsMain/kotlin/androidx/compose/web/css/CSSEnums.kt#L5)

and here ["fake enum implementation"](https://github.com/JetBrains/compose-jb/blob/e2dd2562355edf5a63e156e7a7a8bdbb050b102d/web/core/src/jsMain/kotlin/androidx/compose/web/css/CSSEnums.kt#L25)

PR #860, opened 2021-07-07
web: Add Event Listener Options

eymar:
There are 3 parameters in Options
```kotlin
class Options(
    val capture: BooleanValue = BooleanValue.Default,
    val once: BooleanValue = BooleanValue.Default,
    val passive: BooleanValue = BooleanValue.Default,
)
```

Usage example:
```kotlin
onClick(options = Options(capture = Options.BooleanValue.True)) {
      ...
}
```


Skolotsky:
Can we use `Boolean?` instead of `BooleanValue` ?

Schahen:
First thing we need to discuss why we need this. 
Exactly like we don't need to be able to removeClass - I see no reason why we should support adding event once. 
So, first, let's discuss )

eymar:
We discussed and decided to not merge it now as we can't see any need in Options for now.

PR #862, opened 2021-07-07
web: Use SyntheticMouseEvent as a type for a value in mouse events listeners

eymar:


web/core/src/jsMain/kotlin/org/jetbrains/compose/web/events/SyntheticEvent.kt
```diff
@@ -0,0 +1,26 @@
...
+    val target: Element,
+    val nativeEvent: Event
+) {
+    val bubbles: Boolean = nativeEvent.bubbles
+    val cancelable: Boolean = nativeEvent.cancelable
+    val composed: Boolean = nativeEvent.composed
+    val currentTarget: HTMLElement? = nativeEvent.currentTarget.unsafeCast<HTMLInputElement?>()
+    val eventPhase: Short = nativeEvent.eventPhase
+    val defaultPrevented: Boolean = nativeEvent.defaultPrevented
+    val timestamp: Number = nativeEvent.timeStamp
```
Skolotsky, line 18: I think it should be Int

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/events/SyntheticMouseEvent.kt
```diff
@@ -0,0 +1,89 @@
...
+
+    constructor(nativeEvent: WheelEvent): this(
+        target = nativeEvent.target as HTMLElement,
+        nativeEvent = nativeEvent
+    )
+
+    val deltaX: Double = nativeEvent.deltaX
+    val deltaY: Double = nativeEvent.deltaY
+    val deltaZ: Double = nativeEvent.deltaZ
+    val deltaMode: Int = nativeEvent.deltaMode
```
Skolotsky, line 69: Is it sort of enum?

Skolotsky (changes requested):

Schahen (approved):

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/events/SyntheticMouseEvent.kt
eymar: yes, sort of:

here is its defenition in kotlin/js
```kotlin
 companion object {
        val DOM_DELTA_PIXEL: Int
        val DOM_DELTA_LINE: Int
        val DOM_DELTA_PAGE: Int
        val NONE: Short
        val CAPTURING_PHASE: Short
        val AT_TARGET: Short
        val BUBBLING_PHASE: Short
    }
```

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/events/SyntheticEvent.kt
eymar: in kotlin/js nativeEvent the type is Number, so I followed its example.

Did you want to suggest that it's better to cast to Int?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/events/SyntheticMouseEvent.kt
Skolotsky: it's better to use another type than `Int`, to prompt available values and use safe `when(deltaMode)`

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/events/SyntheticEvent.kt
Skolotsky: It's bad typing, timestamp is never Double

PR #869, opened 2021-07-09
Proof of concept for newer CSS API

Schahen:
The idea is to get rid of existing API calls where we try to mimick one-slot string union type values via enum.
Instead it is supposed to use specialized methods.

This is proof of concept, we also have inline vals as in CSSEnums and the idea is to choose the best syntax based on what will generate smaller bundles. 

However it give idea what we are up to (regardless of what syntax we'll choose - it won't affect the API we are providing for users).


Schahen:
Decided to move in a differente direction

PR #870, opened 2021-07-09
Web/add compose plugin integration cases for kotlinjs

eymar:


web/compose-compiler-integration/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
+fun cloneTemplate(templateName: String, contentMain: String, contentLib: String): File {
```
Schahen, line 1: If I got what's going on here right (and I apologize if not) it seems to me that just using copy task would be a better thing to invest into. Reason is that Copy task, among other things, sets inputs and outputs - that said, the task will be triggered only when relevant changes happened.

web/compose-compiler-integration/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+    } else {
+        ProcessBuilder("bash", "./gradlew", *buildCmd)
+    }
+    val proc = procBuilder
+        .directory(directory)
+        .redirectOutput(ProcessBuilder.Redirect.PIPE)
+        .redirectError(ProcessBuilder.Redirect.PIPE)
+        .start()
+
+    proc.waitFor(5, TimeUnit.MINUTES)
```
Schahen, line 34: why we need this?

web/compose-compiler-integration/build.gradle.kts
```diff
@@ -0,0 +1,134 @@
...
+
+    "(COMPOSE_INTEGRATION_VERSION=\\[.*\\])".toRegex().find(
+        proc.inputStream.bufferedReader().readText()
+    )?.also {
+        println(it.groupValues[1])
+    }
+
+    println(proc.errorStream.bufferedReader().readText())
```
Schahen, line 42: should we print this all the time or only when exit code is not null?

web/compose-compiler-integration/build.gradle.kts
eymar: `proc.waitFor(5, TimeUnit.MINUTES)`
Causes the current thread to wait, if necessary, until the subprocess represented by this Process object has terminated, or the specified waiting time elapses.

We can poll the `exitValue` value or `isAlive()` in the loop instead. But anyway we need to set some timeout. So `waitFor` seems to be the most concise solution for the task.

web/compose-compiler-integration/build.gradle.kts
eymar: in case when exitCode == 0, it will simply print an empty string.
So yes, it's not necessary to always print, although it doesn't hurt. This printed message is only for our convenience when we read the output.

PR #879, opened 2021-07-12
Web: Add Form onSubmit Event Listener

hfhbd:
Is this the right file/place?


eymar:
Hi @hfhbd ,
thanks for pointing out the missing event listener. 
Yes, it's the right place for the current version. 

The thing is that the events are being refactored at the moment. I think merging this PR won't hurt anyone, but the change will be overridden quite soon by the refactoring.  
So I prefer keeping this PR open (as as reminder) till the refactoring gets merged and the onSubmit event is also present. 
Does it work for you?

hfhbd:
@eymar Sure, this will not hurt me (hard) :D

eymar:
onSubmit and onReset were added in https://github.com/JetBrains/compose-jb/pull/887

Issue #738, opened 2021-06-02
Compose Web: Can't use Kotlinx.Serialization 1.5.10 / 1.2.1 with Kotlin 1.5.10: IllegalStateException: IdSignature clash


SebastianAigner:
In a project with the following build file:

```

plugins {
    kotlin("multiplatform") version "1.5.10"
    kotlin("plugin.serialization") version "1.5.10"
    id("org.jetbrains.compose") version "0.0.0-web-dev-14"
}

repositories {
    mavenCentral()
    maven("https://maven.pkg.jetbrains.space/public/p/compose/dev")
}

kotlin {
    js(IR) {
        browser()
        binaries.executable()
    }
    sourceSets {
        val jsMain by getting {
            dependencies {
                implementation(compose.web.core)
                implementation(compose.runtime)
                implementation("org.jetbrains.kotlinx:kotlinx-serialization-json:1.2.1")
            }
        }
    }
}

```

...any attempt to serialize by adding the `@Serializable` annotation an object results in the following error messages:

```
java.lang.IllegalStateException: IdSignature clash: public /TodoItem.$serializer|null[0]; Existed declaration CLASS OBJECT name:$serializer modality:FINAL visibility:public superTypes:[kotlinx.serialization.internal.GeneratedSerializer<<root>.TodoItem>] clashed with new CLASS OBJECT name:$serializer modality:FINAL visibility:public superTypes:[kotlinx.serialization.internal.GeneratedSerializer<<root>.TodoItem>]
	at org.jetbrains.kotlin.ir.backend.js.lower.serialization.ir.JsUniqIdClashTracker.commit(JsDeclarationTable.kt:35)
	at org.jetbrains.kotlin.backend.common.serialization.GlobalDeclarationTable.computeSignatureByDeclaration(DeclarationTable.kt:46)
	at org.jetbrains.kotlin.backend.common.serialization.DeclarationTable.computeSignatureByDeclaration(DeclarationTable.kt:74)
	at org.jetbrains.kotlin.backend.common.serialization.DeclarationTable.signatureByDeclaration(DeclarationTable.kt:83)
	at org.jetbrains.kotlin.backend.common.serialization.IrFileSerializer.protoIdSignature(IrFileSerializer.kt:237)
	at org.jetbrains.kotlin.backend.common.serialization.IrFileSerializer.serializeIrSymbol(IrFileSerializer.kt:291)
[...]

PR #900, opened 2021-07-16
Add a check that compose is used along with kotlinx.serialization

eymar:
There's a known issue that both plugins can't be used in the same module (specifically for kotlin/js targets).

We decided to make the problem more explicit by showing a message.
The compose gradle plugin will check for `kotlinx.serialization` and will show a warning.

@AlexeyTsvetkov  What do you think? Is there a more idiomatic way to solve such kind of task?
Thanks!

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/WarnAboutComposeWithSerialization.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/WarnAboutComposeWithSerialization.kt
@@ -1,28 +1,18 @@
 package org.jetbrains.compose.internal
 
 import org.gradle.api.Project
-import org.gradle.configurationcache.extensions.serviceOf
-import org.gradle.internal.logging.text.StyledTextOutput
-import org.gradle.internal.logging.text.StyledTextOutputFactory
 
 internal fun Project.checkAndWarnAboutComposeWithSerialization() {
-    afterEvaluate {
-        val usesKotlinxSerialization = configurations.names.asSequence().filter {
-            it.startsWith("kotlinCompilerPluginClasspath")
-        }.any { configurationName ->
-            configurations.getByName(configurationName).dependencies.any { dependency ->
-                dependency.name.contains("kotlin-serialization")
-            }
-        }
+    project.plugins.withId("org.jetbrains.kotlin.plugin.serialization") {
+        val warningMessage = """
 
-        if (usesKotlinxSerialization) {
-            val out = serviceOf<StyledTextOutputFactory>().create("COMPOSE_PLUGIN")
-            out.style(StyledTextOutput.Style.FailureHeader)
-                .text("WARNING! Both Compose and kotlinx.serialization plugins are used in the module '${project.name}'")
-                .println()
-                .style(StyledTextOutput.Style.Failure)
-                .text("Consider using these plugins in separate modules to avoid compilation errors")
-                .println()
-        }
+            >>> COMPOSE WARNING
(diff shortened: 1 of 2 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/WarnAboutComposeWithSerialization.kt
```diff
@@ -0,0 +1,28 @@
+package org.jetbrains.compose.internal
+
+import org.gradle.api.Project
+import org.gradle.configurationcache.extensions.serviceOf
+import org.gradle.internal.logging.text.StyledTextOutput
+import org.gradle.internal.logging.text.StyledTextOutputFactory
+
+internal fun Project.checkAndWarnAboutComposeWithSerialization() {
+    afterEvaluate {
```
AlexeyTsvetkov, line 9: A [recommended way](https://github.com/Kotlin/kotlinx.serialization#using-the-plugins-block) to set up serialization is to use the dedicated Gradle plugin, so instead of checking a plugin classpath, you could just add a callback on the serialization plugin use:
```
project.plugins.withId("org.jetbrains.kotlin.plugin.serialization") {
     // ...
}
```

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/WarnAboutComposeWithSerialization.kt
```diff
@@ -0,0 +1,28 @@
...
+        val usesKotlinxSerialization = configurations.names.asSequence().filter {
+            it.startsWith("kotlinCompilerPluginClasspath")
+        }.any { configurationName ->
+            configurations.getByName(configurationName).dependencies.any { dependency ->
+                dependency.name.contains("kotlin-serialization")
+            }
+        }
+
+        if (usesKotlinxSerialization) {
+            val out = serviceOf<StyledTextOutputFactory>().create("COMPOSE_PLUGIN")
```
AlexeyTsvetkov, line 19: Please don't use internal apis for simple logging, `project.logger.warn()` or `project.logger.error()` are more than enough

AlexeyTsvetkov (changes requested):

hfhbd:
@eymar Maybe you could specify the link to an issue or readme, proving more details and the workaround, instead referring to this repo only.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/WarnAboutComposeWithSerialization.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.gradle.api.Project
+
+internal fun Project.checkAndWarnAboutComposeWithSerialization() {
+    project.plugins.withId("org.jetbrains.kotlin.plugin.serialization") {
+        val warningMessage = """
+
+            >>> COMPOSE WARNING
+            >>> Project `${project.name}` has `compose` and `kotlinx.serialization` plugins applied!
+            >>> Consider using these plugins in separate modules to avoid compilation errors
+            >>> Check more details here: https://github.com/JetBrains/compose-jb
```
olonho, line 12: Please use proper link to an issue.

eymar: Updated the link: https://github.com/JetBrains/compose-jb/issues/738

olonho (approved):

Issue #896, opened 2021-07-15
Move from java-like enum-like String values to a flat hierarchy of light-weight inline values


Schahen:
This is an experimental issue and there's non-zero probability that, after it will be done we'll still decide to stick with current API. It depends on lot of factors, such as size of generated code and UX in IDE.

Schahen:
We've completely reimplemented API with shorter (and flatter) syntax and came to conclusion that we don't want t rush and shove in such radical changes to the alpha release. The main reason for this is that IDE experience is actually arguably worser. I'm closing this one and on further improvements we'll work in separate proposals / tasks

PR #906, opened 2021-07-19
Shorter syntax for keywords in CSS API

Schahen:
The goal of this PR is more or less described in #896 

basically we want to write instead of this:
```kotlin
display(DisplayStyle.Block)
```
just this
```
display(block)
```


Schahen:
CSS API will evolve in a different direction

PR #909, opened 2021-07-19
Add previews to TodoApp example

arkivanov:


olonho (approved):
You'll likely need recent Compose build for preview to work with up to date IDE plugin.

PR #954, opened 2021-07-29
Sync samples

Schahen:
Goal of this particular PR is always keep web examples in consistent state. 
For that I copied all sources to web/samples and introduced `generateExamples` task. 
Samples will be always built against master and examples will be generated on demand whenever we want to update the version of compose artifacts we use in examples.


web/settings.gradle.kts
```diff
@@ -28,13 +28,22 @@ fun module(name: String, path: String) {
...
 
+module(":samples:falling_balls_with_web", "samples/falling_balls_with_web")
+module(":samples:compose-web-lp", "samples/web_landing")
+module(":samples:web-compose-bird", "samples/web-compose-bird")
+module(":samples:web-with-react", "samples/web-with-react")
+module(":samples:web-getting-started", "samples/web-getting-started")
+
+
+
 if (extra["compose.web.buildSamples"]!!.toString().toBoolean() == true) {
```
eymar, line 47: Is this flag still needed? 

As I see all the example modules are included unconditionally above.

eymar (approved):

web/settings.gradle.kts
Schahen: Those are two kind of beasts - one is build agains master, other one agains published artifact. That said, it's still convenient to be able to build them in a single command

examples/falling_balls_with_web/build.gradle.kts
```diff
@@ -1,6 +1,8 @@
 import org.jetbrains.compose.compose
 import org.jetbrains.compose.desktop.application.dsl.TargetFormat
```
olonho, line 2: Maybe rename to `falling-balls-with-web` to make it uniform?

web/build.gradle.kts
```diff
@@ -1,16 +1,67 @@
...
+}
+
 subprojects {
     apply(plugin = "maven-publish")
 
     group = "org.jetbrains.compose.web"
     version = COMPOSE_WEB_VERSION
 
+    if (isSampleProject()) {
+
```
olonho, line 24: Extra space.

olonho:
Not sure if this is overall right approach, could we discuss?

Issue #908, opened 2021-07-19
The Preview feature does not work in a KMP module with Compose enabled


arkivanov:
STR:

1. Add `implementation(compose.uiTooling)` to `desktopMain` source set of the `todoapp/common/compose-ui` module
2. Sync Gradle
3. Add any `@Preview @Composable` function to `desktopMain` source set
4. Run the preview

AR: The `configureDesktopPreview` Gradle task is not added for the module and Preview feature fails with the following error:
```
FAILURE: Build failed with an exception.

* What went wrong:
Task 'configureDesktopPreview' not found in project ':common:compose-ui'.
```

ER: The Preview feature works

slavonnet:
I have another issue)) Preview migrate to cwd up and AS does not find class ActivittPreview in classpath

akurasov:
I took the application and was not able to use uiTooling. Seems compose version is a bit old. I can't where compose version is setup there. Could you point to it?

arkivanov:
@akurasov What application do you mean? If it is the TodoApp example, then it's [here](https://github.com/JetBrains/compose-jb/blob/e071b2889b86b7008f03719e96b2fa46c96cc63a/examples/todoapp/buildSrc/buildSrc/src/main/kotlin/Deps.kt#L16).

arkivanov:
@akurasov I updated Compose to `0.5.0-build262` and Kotlin to `1.5.21`, and pushed a reproducer [commit](https://github.com/arkivanov/compose-jb/commit/3f0ec9012df109208cd06b24f9f9a519d1059d4c) to a separate [branch](https://github.com/arkivanov/compose-jb/tree/preview-kmp-issue). The not working preview can be found [here](https://github.com/arkivanov/compose-jb/blob/preview-kmp-issue/examples/todoapp/common/compose-ui/src/desktopMain/kotlin/example/todo/common/ui/PreviewIssue.kt).

akurasov:
So with build262 everything works fine? 

The issue could be in incompatible versions of Compose and IdeaCompose plugin. A check for it have been added recently.

arkivanov:
@akurasov No, the issue is still there. I have updated the Compose IDEA plugin to match the Compose version.

akurasov:
Reproduced(build262). configureDesktopPreview present in Gradle window, however if run manually the following error occurs:
> No value has been specified for property 'idePort$compose'.
> No value has been specified for property 'previewTarget$compose'.

[...]

PR #1002, opened 2021-08-04
Move TodoApp previews to compose-ui module

arkivanov:
The issue #908 is fixed and so previews can be moved to `compose-ui` module.

### DIFF
--- a/examples/todoapp/common/compose-ui/build.gradle.kts
+++ b/examples/todoapp/common/compose-ui/build.gradle.kts
@@ -1,5 +1,3 @@
-import org.jetbrains.compose.compose
-
 plugins {
     id("multiplatform-compose-setup")
     id("android-setup")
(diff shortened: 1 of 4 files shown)


examples/todoapp/common/compose-ui/build.gradle.kts
```diff
@@ -14,5 +16,11 @@ kotlin {
                 implementation(Deps.ArkIvanov.Decompose.extensionsCompose)
             }
         }
+
+        named("desktopMain") {
+            dependencies {
```
olonho, line 21: Is it really needed with alpha3 and later?

arkivanov: Seems like it works without the `ui-tooling` dependency. Should I remove it?

igordmn: Let's remove it, as we transitively depend on it [here](https://github.com/JetBrains/compose-jb/blob/696d20c931feb848f113de1e67fe9446adef538f/examples/todoapp/buildSrc/src/main/kotlin/multiplatform-compose-setup.gradle.kts#L31).

We have to [explicitly](https://github.com/JetBrains/compose-jb/blob/696d20c931feb848f113de1e67fe9446adef538f/templates/multiplatform-template/common/build.gradle.kts#L20) add it only if we don't depend on the aggregated dependency `compose.desktop.common`

arkivanov: Done

PR #1031, opened 2021-08-09
Web/controlled inputs

eymar:


web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/InputElements.kt
```diff
@@ -13,54 +13,86 @@ private fun InputAttrsBuilder<String>.applyAttrsWithStringValue(
     attrs()
 }
 
+/**
+ * It's a controlled Input of [InputType.Checkbox].
```
Schahen, line 17: I suggest us to link to the relevant piece of code where we explain what controlled actually means.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Elements.kt
```diff
@@ -650,22 +650,51 @@ fun Section(
...
+ * ```
+ */
 @Composable
 fun TextArea(
-    attrs: (TextAreaAttrsBuilder.() -> Unit)? = null,
-    value: String
-) = TagElement(
-    elementBuilder = TextArea,
-    applyAttrs = {
-        val  taab = TextAreaAttrsBuilder()
```
Schahen, line 660: **(I do realize that this is the removed code, I just want to use it as a chance to discuss this)**
honestly, let's not use "c-like" variable names,
1) we are using  IDE
2) the benefits of immediately understanding what this variable is about over-weighs the perks of short name)

eymar: agree

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/InputElements.kt
eymar: improved

Issue #1036, opened 2021-08-10
Exception throw while using compose.desktop.components.splitPane


sonder-joker:
```
java.lang.IllegalStateException: CompositionLocal LocalAppWindow not provided
	at androidx.compose.desktop.AppWindow_desktopKt$LocalAppWindow$1.invoke(AppWindow.desktop.kt:50)
	at androidx.compose.desktop.AppWindow_desktopKt$LocalAppWindow$1.invoke(AppWindow.desktop.kt:49)
	at kotlin.SynchronizedLazyImpl.getValue(LazyJVM.kt:74)
	at androidx.compose.runtime.LazyValueHolder.getCurrent(ValueHolders.kt:29)
	at androidx.compose.runtime.LazyValueHolder.getValue(ValueHolders.kt:31)
	at androidx.compose.runtime.ComposerImpl.resolveCompositionLocal(Composer.kt:1780)
	at androidx.compose.runtime.ComposerImpl.consume(Composer.kt:1750)
```
The reason is obviously that use deprecated LocalAppWindow. 
```kotlin
private fun Modifier.cursorForHorizontalResize(
    isHorizontal: Boolean
): Modifier = composed {
    var isHover by remember { mutableStateOf(false) }

    if (isHover) {
        LocalAppWindow.current.window.cursor = Cursor(
            if (isHorizontal) Cursor.E_RESIZE_CURSOR else Cursor.S_RESIZE_CURSOR
        )
    } else {
        LocalAppWindow.current.window.cursor = Cursor.getDefaultCursor()
    }
    pointerMoveFilter(
        onEnter = { isHover = true; true },
        onExit = { isHover = false; true }
    )
}
```
It's use a LocalAppWindow to get it. How about use 
```kotlin
internal val LocalWindow = compositionLocalOf<Window?> { null }
```
to instead it?
If you can give some advice about how to fix it , I will make a pull request on it.

akurasov:
Now we have new API for cursor changing - Modifier.pointerIcon. Take a look at https://github.com/JetBrains/compose-jb/blob/242ec6bbf9b9df085e9d6e56e63e14ff82e4a52d/examples/widgets-gallery/common/src/desktopMain/kotlin/org/jetbrains/compose/demo/widgets/platform/Mouse.kt as an example. I think it should work here.

sonder-joker:
Already create a pull request ! Wait for your review

okushnikov:
[...]

PR #1037, opened 2021-08-10
Fix component SplitPane

sonder-joker:
#1036 
Use the new way to set pointer and remove unnecessary import

### DIFF
--- a/components/SplitPane/demo/src/jvmMain/kotlin/org/jetbrains/compose/splitpane/demo/Main.kt
+++ b/components/SplitPane/demo/src/jvmMain/kotlin/org/jetbrains/compose/splitpane/demo/Main.kt
@@ -1,26 +1,18 @@
 package org.jetbrains.compose.splitpane.demo
 
 import androidx.compose.desktop.DesktopTheme
-import androidx.compose.desktop.LocalAppWindow
-import androidx.compose.desktop.Window
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.width
 import androidx.compose.material.MaterialTheme
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
-import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.composed
 import androidx.compose.ui.graphics.Color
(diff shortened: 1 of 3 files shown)


components/SplitPane/library/src/desktopMain/kotlin/org/jetbrains/compose/splitpane/DesktopSplitter.kt
```diff
@@ -1,39 +1,28 @@
...
 private fun Modifier.cursorForHorizontalResize(
     isHorizontal: Boolean
 ): Modifier = composed {
     var isHover by remember { mutableStateOf(false) }
 
     if (isHover) {
-        LocalAppWindow.current.window.cursor = Cursor(
-            if (isHorizontal) Cursor.E_RESIZE_CURSOR else Cursor.S_RESIZE_CURSOR
-        )
+        pointerIcon(PointerIcon(Cursor(if (isHorizontal) Cursor.E_RESIZE_CURSOR else Cursor.S_RESIZE_CURSOR)))
```
igordmn, line 23: `pointerIcon` has no effect, because we still return pointerMoveFilter instead.

The right fix would be
```
@OptIn(ExperimentalComposeUiApi::class)
private fun Modifier.cursorForHorizontalResize(
    isHorizontal: Boolean
): Modifier = composed {
    pointerIcon(PointerIcon(Cursor(if (isHorizontal) Cursor.E_RESIZE_CURSOR else Cursor.S_RESIZE_CURSOR)))
}
```

components/SplitPane/demo/src/jvmMain/kotlin/org/jetbrains/compose/splitpane/demo/Main.kt
```diff
@@ -13,26 +13,31 @@ import androidx.compose.runtime.getValue
...
 import java.awt.Cursor
 
+@OptIn(ExperimentalComposeUiApi::class)
 private fun Modifier.cursorForHorizontalResize(
 ): Modifier = composed {
     var isHover by remember { mutableStateOf(false) }
 
     if (isHover) {
-        LocalAppWindow.current.window.cursor = Cursor(Cursor.E_RESIZE_CURSOR)
+        pointerIcon(PointerIcon(Cursor(Cursor.E_RESIZE_CURSOR)))
```
igordmn, line 38: The same as above:
```
@OptIn(ExperimentalComposeUiApi::class)
private fun Modifier.cursorForHorizontalResize(): Modifier = composed {
    pointerIcon(PointerIcon(Cursor(Cursor.E_RESIZE_CURSOR)))
}
```

igordmn (changes requested):

sonder-joker:
Already repair and now it work well in my environment

igordmn (approved):

PR #1044, opened 2021-08-11
Add transparency to multiplatform widgets Color class

### LINKED ISSUE #943: Add transparency to Color in multiplatform widgets
Currently, the Color class in the multiplatform widgets only supports red, green and blue values, and adding an alpha value would open up the entire range of colors.

andyburris:
Should resolve #943 

Adds an alpha value to the multiplatform widgets `Color` class (from the `compose.web.widgets` artifact). 

Right now the class is structured like the CSS implementation of `rgba()` (the red, green, and blue components are 0-255 `Int` values and the alpha value is a 0-1 `Float` value) and not the desktop implementation (all `Float` values), but if the other API is preferred, I can change it. 

Also, by adding top-level functions in both `commonMain/Color.kt` and `jvmMain/Color.kt`, the compiler throws a `Duplicate JVM class name` error (since it creates two classes named `ColorKt`). This seems like the first time this issue has been dealt with in this repository, so I've solved by adding `@file:JvmName("PlatformColor")` to the file in `jvmMain/Color.kt`, but was wondering if there was a convention for the `JvmName` that I should use, or alternatively if I should just extract any top-level functions into a helper file to avoid this problem entirely.


web/widgets/src/commonMain/kotlin/Color.kt
```diff
@@ -1,18 +1,21 @@
 package org.jetbrains.compose.common.core.graphics
 
-public data class Color(val red: Int, val green: Int, val blue: Int) {
+public data class Color(val red: Int, val green: Int, val blue: Int, val alpha: Float) {
```
eymar, line 3: Do you see any concerns if `val alpha: Float` would have a default value `1.0`?

andb3: Seems good to me, plus that would get rid of the top-level function issue (for now, though eventually there would probably be some helper functions like `Color(value: Long)` that would reintroduce that issue)

eymar:
> This branch has conflicts that must be resolved

no need to resolve conflicts.
there is an update regarding these widgets - https://github.com/JetBrains/compose-jb/pull/1453 
since widgets will be deprecated, i think it makes sense to close this PR.

PR #1048, opened 2021-08-12
web: Update test utils api to prepare it for publishing

eymar:


web/core/src/jsTest/kotlin/TestUtils.kt
```diff
@@ -1,68 +1,167 @@
...
-        }
+    @Deprecated(
+        message = "The intention of this function was to wait for recompositions. Now there is a more specific function",
+        replaceWith = ReplaceWith(expression = "waitForRecompositionComplete()"),
+        level = DeprecationLevel.WARNING
+    )
+    // TODO: remove this function. It looks like there is no need to keep it
+    internal suspend fun waitForAnimationFrame() {
+        waitForRecompositionComplete()
+//        suspendCoroutine<Unit> { continuation ->
```
Schahen, line 118: Let's just remove commented code if we don't need it.

web/core/src/jsTest/kotlin/TestUtils.kt
```diff
@@ -1,68 +1,167 @@
...
 
-If tests get broken, then DefaultMonotonicFrameClock need to be checked if it still
-uses window.requestAnimationFrame */
-internal suspend fun waitForAnimationFrame() {
-    suspendCoroutine<Unit> { continuation ->
-        window.requestAnimationFrame {
-            continuation.resume(Unit)
-        }
+    @Deprecated(
+        message = "The intention of this function was to wait for recompositions. Now there is a more specific function",
```
Schahen, line 111: My understanding is that it's only us who used this code - so do we actually need to deprecate it?

Schahen (approved):
Apart from two - very minor - things - I don't see anything I should comment about

web/core/src/jsTest/kotlin/TestUtils.kt
eymar: since it's going to be public, I decided to deprecate it for us as well.
Actually, I believe there is no reason to keep this function. So I replace it's usages by  `waitForRecompositionComplete` and remove deprecated `waitForAnimationFrame `

web/core/src/jsTest/kotlin/TestUtils.kt
eymar: removed the function completely (it was deprecated)

Issue #922, opened 2021-07-22
Support transform functions in CSS API


Schahen:
Basically we've decided to make it possible to write typed versions of any transformation, so instead of
```kotlin
 property("transform", "perspective(500px) translate(10px, 0, 20px) rotateY(3deg)")
```

one will be able to form arguments without necessity to form a string

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-6316">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1084, opened 2021-08-20
Support transformation in CSS API

Schahen:
This PR is dedicated to https://github.com/JetBrains/compose-jb/issues/922

After long discussion we decided to have following approach - whenever CSS property is actually a stacked of consequently applied functions, we have builder syntax instead of "simple" one.

That said, one invokes transform the following way:
```kotlin
 Div({
    style {
       transform {
             skew(4.deg)
        }
    }
 }) 
```


eymar (approved):

Schahen:
Merged already

PR #1088, opened 2021-08-21
Fix VideoPlayer

theapache64:
# Changes

This PR fixes two issues with the `VideoPlayer` sample.

## 🐛   No Video Issue

**Problem:**
- Previously, we were using  `vlcj`'s `EmbeddedMediaPlayerComponent` to play the video. This component doesn't work on macOS due to insufficient support from the `AWT` side. [[1]](https://stackoverflow.com/a/59346015/4370279)[[2]](https://github.com/caprica/vlcj/issues/887#issuecomment-503288294). This caused the player not to play the video on MacOS.

**Solution:**

- Check the OS type and provide `EmbeddedMediaPlayerComponent`  or `CallbackMediaPlayerComponent`(for macOS) accordingly. (here's the commit https://github.com/JetBrains/compose-jb/commit/1e396c79e01d7162e522c5fc3a92e52b1560e238)

## 🐛   Local file path Issue

**Problem:**
Previously, the URL was a local file path that wouldn't exist in every system.

**Solution:**
Update it to a public URL. Here's the commit (4366a6ea1aae6393879e7c9589fa9386176ee1fa)

# Output

https://user-images.githubusercontent.com/9678279/130316868-19d78267-db95-4bd3-be7d-52037935b8ae.mp4

### DIFF
--- a/components/VideoPlayer/demo/src/jvmMain/kotlin/org/jetbrains/compose/videoplayer/demo/Main.kt
+++ b/components/VideoPlayer/demo/src/jvmMain/kotlin/org/jetbrains/compose/videoplayer/demo/Main.kt
@@ -13,7 +13,11 @@ fun main() {
     ) {
         MaterialTheme {
             DesktopTheme {
-                VideoPlayer("/System/Library/Compositions/Yosemite.mov", 640, 480)
+                VideoPlayer(
+                    url = "http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4",
+                    width = 640,
+                    height = 480
+                )
             }
         }
     }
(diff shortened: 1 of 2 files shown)


igordmn:
Found issues:

1. Fullscreen is too slow on macOs on my machine (MacBook Pro 2019 16" i9), but not slow on the `theapache64` machine.
2. I noticed a frame tearing (as happened when vsync is disabled)
3. Encountered:
```
 java.io.IOException: Native library (darwin/libvlc.dylib) not found in resource path
```
until I installed VLC:
```
brew install vlc
```
I thinks we need to bundle VLC as a part of VideoPlayer.

theapache64:
@igordmn I am using the same MBP, but with a dual monitor setup. I was running the demo on the external monitor. Moving the output window to MBC's display also causing the same issue.

PR #1098, opened 2021-08-23
Let `Functional` inherit from `Nth`

EdwarDDay:
I'm not sure about this, but it looks like this is missing.


akurasov:
@Schahen will you take a look, please?

EdwarDDay:
Fixed by #1609

PR #1112, opened 2021-08-25
Web: Align StyleSheetBuilder parameter with other @Composable ContentBuilder

hfhbd:


Schahen:
@hfhbd let's discuss more specifically why in your opinion this should be done? 
If we accidentally broke some existing behavior - what exactly behavior and what should be the test case.

If it's something new, a proposal - let's discuss what the goal of that proposal.

hfhbd:
Yes, the latest change broke existing code. 
In d2d4971f the inline modifier of
```kotlin
@Composable
inline fun Style(
    noinline applyAttrs: (AttrsBuilder<HTMLStyleElement>.() -> Unit)? = null,
    rulesBuild: StyleSheetBuilder.() -> Unit
) {
```
was removed, so it is not possible to call `@Composable` in `rulesBuild` anymore.

Usecase::

https://github.com/hfhbd/bootstrap-compose/blob/751945f55b64ed966a4d215e40962acd0b5d8818/src/jsMain/kotlin/app/softwork/bootstrapcompose/GridBox.kt#L94
```kotlin
Style {
            val classname = remember { "_${UUID()}" } // creating dynamically a style

            CSSSelector.CSSClass(classname) style {
                gap?.let {
                    property("gap", gap.toString())
                }
                justifyContent?.let {
                    property("justify-content", it.toString())
                }
                alignContent?.let {
                    property("align-content", it.toString())
                }
            }
```

hfhbd:
@Schahen Added a test. I hope it is okay :D

PR #1165, opened 2021-09-15
web: add `testUtils` alias to gradle plugin

eymar:
This will allow to use the dependency like this:
`compose.web.testUtils`


eymar:
@Schahen This a minor change. Anyway, please let me know if you think the alias name is fine.

PR #1233, opened 2021-10-06
Basic animations test

Skolotsky:


web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Style.kt
```diff
@@ -81,6 +81,38 @@ private fun fillRule(
...
+            }
+        }
+        is CSSGroupingRuleDeclaration -> {
+            cssRuleDeclaration.rules.forEach { childRuleDeclaration ->
+                stringBuilder.append("$delimiter${childRuleDeclaration.stringPresentation(baseIndent + indent, indent, delimiter)}")
+            }
+        }
+        is CSSKeyframesRuleDeclaration -> {
+            cssRuleDeclaration.keys.forEach { childRuleDeclaration ->
+                stringBuilder.append("$delimiter${childRuleDeclaration.stringPresentation(baseIndent + indent, indent, delimiter)}")
```
Schahen, line 108: let's just use regular good old joinToString

Skolotsky: ok

Issue #705, opened 2021-05-27
Support scrollbars for text fields


igordmn:
0.4.0-build211

Currently there is no way to add scrollbars to the text field because text fields have their own scroll state that is not accessible outside of the text field.

```
import androidx.compose.desktop.Window
import androidx.compose.foundation.VerticalScrollbar
import androidx.compose.foundation.layout.Box
import androidx.compose.foundation.layout.fillMaxHeight
import androidx.compose.foundation.layout.fillMaxSize
import androidx.compose.foundation.rememberScrollState
import androidx.compose.foundation.rememberScrollbarAdapter
import androidx.compose.foundation.text.BasicTextField
import androidx.compose.ui.Alignment
import androidx.compose.ui.Modifier

fun main() = Window {
    val scrollState = rememberScrollState()
    Box(
        modifier = Modifier.fillMaxSize()
    ) {
        BasicTextField((1..10000).joinToString { "x" }, {}, Modifier.fillMaxSize())
        VerticalScrollbar(
            rememberScrollbarAdapter(scrollState),
            modifier = Modifier.align(Alignment.CenterEnd).fillMaxHeight()
        )
    }
}
```

igordmn:
See also another related issue with scroll and text field: https://github.com/JetBrains/compose-jb/issues/26

m-sasha:
Fixed by https://github.com/JetBrains/compose-multiplatform-core/pull/384

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-5398">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1274, opened 2021-10-19
Add Code Editor component

n34t0:
Hi!

I am developing a simple code editor using Jetpack Compose for Desktop and IntelliJ Platform: https://github.com/n34t0/compose-code-editor. I wonder if Jetbrains would be interested in accepting such component contributions to the compose-jb repository.

A few explanations:

- Because of #705 I had to do a CoreTextField fork to get the scroll offset for correct positioning.
- I used IntelliJ Platform with some modifications to implement the code completion and go-to-declaration functions. Scripts from `buildSrc` are used to retrieve and prepare platform code. Since Gradle does not support `buildSrc` in subprojects, I had to put the platform code into a separate module and include it through `includeBuild`.
- To keep the client code from depending on these platform build scripts, I put all the necessary platform code into some fat jars.  But I'm not quite sure how to properly publish these jars.


olonho:
Thanks for the great code! @n34t0 this code seems too large for `components` folder in this repo. I would suggest to keep it in the separate repo. We could consider smaller contribution with the rich editor component, as this is more likely widely reusable.

n34t0:
Yes, I agree, this code is probably too large for components. Thanks!

PR #1406, opened 2021-11-16
Fix compatibility with Intellij 2021.3

### LINKED ISSUE #1373: Add IJ2021.3 Beta support for compose multi-platform plugin
Compose multi-platform plugin is not working on IJ2021.3 which is in beta now.

AlexeyTsvetkov:
Resolves #1373

### DIFF
--- a/idea-plugin/build.gradle.kts
+++ b/idea-plugin/build.gradle.kts
@@ -18,7 +18,6 @@ repositories {
 
 dependencies {
     implementation("org.jetbrains.compose:preview-rpc")
-    implementation(kotlin("reflect"))
 }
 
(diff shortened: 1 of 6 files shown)


idea-plugin/build.gradle.kts
```diff
@@ -18,6 +18,7 @@ repositories {
 
 dependencies {
     implementation("org.jetbrains.compose:preview-rpc")
+    implementation(kotlin("reflect"))
```
olonho, line 21: Cannot we just use JVM reflection, not Kotlin's?

olonho (approved):

PR #1418, opened 2021-11-18
Fix css selectors: combine, sibling

Skolotsky:


eymar:
would you like to add tests for these changes with `getComputedStyle()`? It's good to check how styles get applied to elements.

Skolotsky:
> would you like to add tests for these changes with `getComputedStyle()`? It's good to check how styles get applied to elements.

I think that there should be (and i will make it) test which checks that rules apply to DOM correctly. But here i want to test that selector applied to rule correctly. And bug was in applying to rule. So it's kinda unrelated.

eymar (approved):

Issue #1405, opened 2021-11-16
Option to specify license for deb packages


sebkur:
It seems to be possible to specify a license for RPM linux packages, but there is no option for deb. I think there should probably be such an option. Existing rpm option:
```
rpmLicenseType = "TYPE_OF_LICENSE" — a type of license for the rpm package;
```
proposed deb option:
```
debLicenseType = "TYPE_OF_LICENSE" — a type of license for the deb package;
```
At the moment I do get this file located at `share/doc/copyright` when installing a `*.deb` file produced using `./gradlew packageDeb`:
```
Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/

Files: *
Copyright: Copyright (C) 2021
License: Unknown
```

AlexeyTsvetkov:
@sebkur does existing `copyright` property not work for you?

```
compose.desktop {
    application {
        nativeDistributions {
            packageName = "ExampleApp"
            version = "0.1-SNAPSHOT"
            description = "Compose Example App"
            copyright = "© 2020 My Name. All rights reserved."
            vendor = "Example vendor"
        }
    }
}
```

https://github.com/JetBrains/compose-jb/tree/master/tutorials/Native_distributions_and_local_execution#customizing-metadata

sebkur:
thanks @AlexeyTsvetkov for your quick response. Copyright does help a bit indeed for changing the content after the `Copyright: ` line, however I am more concerned about the content after the `License: ` line and my goal would be to change "Unknown" to something like "MIT License"

sebkur:
I gave it a try to fix this, see the PR above

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-6208">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1420, opened 2021-11-18
Linux packaging: use licenseType for deb package

sebkur:
Rename option 'rpmLicenseType' to 'licenseType' and use also for deb packaging

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/PlatformSettings.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/PlatformSettings.kt
@@ -59,7 +59,7 @@ open class LinuxPlatformSettings @Inject constructor(objects: ObjectFactory): Pl
     var appCategory: String? = null
     var debMaintainer: String? = null
     var menuGroup: String? = null
-    var rpmLicenseType: String? = null
+    var licenseType: String? = null
     var debPackageVersion: String? = null
     var rpmPackageVersion: String? = null
 }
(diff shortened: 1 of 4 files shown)


sebkur:
Related to #1405

sebkur:
I have no idea how to actually test what I've done here does the trick. I guess I'd need to build my regular project using the plugin from this branch instead of the one pulled from the Maven repo and see if the results are as expected. Any guidance on how to do that?

sebkur:
OK, tested like this:

Publish plugin locall: `./gradlew publish`

In the project to test, add the local build repository to the buildscript stanza:
```
buildscript {
    reopsitories {
        maven("/path/to/.../github/JetBrains/compose-jb/gradle-plugins/build/repo/")
    }
}
```

Result: doesn't work. Turns out this actually depends on `jpackage` and the CLI options that tool has. It does have a `--linux-rpm-license-type` option but but no such thing as a `--linux-license-type` option

sebkur:
Aha! There is however a `--license-file` option for `jpackage` and the contents of the file specified will be put in place of `Unknown` in `/opt/<application-name>/share/doc/copyright`. Using the Apache license file from the compose-jb repo, I get this as a result:

```
Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/

Files: *
Copyright: © 2021 Blah blah.
License:                                  Apache License
                           Version 2.0, January 2004
                        http://www.apache.org/licenses/

   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION

   1. Definitions.
…
```

sebkur:
Closing this in favor of #1421

PR #1421, opened 2021-11-18
Expose 'licenseFile' option for native packaging

sebkur:
Related to #1405 and #1420


sebkur:
As described in #1405, when I build a Debian package using `./gradlew packageDeb`, after installing that I do get this a file located at `/opt/<app-name>/share/doc/copyright` with this content:
```
Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/

Files: *
Copyright: Copyright (C) 2021
License: Unknown
```

and there's no way to specify the License to something else than "Unknown".

`jpackage` has an option `--license-file` which appends the content of that file to that part of the `copyright` file. The `AbstractJPackageTask` also already passes the option to `jpackage`, it's just that the DSL doesn't define a property to set this which gets passed on to the packaging task.

In this PR, I'm adding this option to the DSL and update the packaging tutorial to mention the new option.

AlexeyTsvetkov:
Looks good, thanks for the contribution!

sebkur:
Thanks for merging and anyway thanks a lot for your awesome work on Compose multiplatform!

PR #1453, opened 2021-11-24
deprecate compose.web.web-widgets in source code and gradle plugin 

eymar:


Schahen (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -205,6 +205,7 @@ class ComposePlugin : Plugin<Project> {
             composeDependency("org.jetbrains.compose.web:web-svg")
         }
 
+        @Deprecated("compose.web.web-widgets API is not deprecated")
```
hfhbd, line 208: is not?

eymar: what a funny typo. Thanks!

Issue #1338, opened 2021-10-31
Add TextInput inputMode attribute and enum


brianguertin:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

E.g.:

```kotlin
TextInput {
   inputMode(InputMode.url)
}
```

eymar:
should be available in the future build

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-5206">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1461, opened 2021-11-25
web: add `inputmode` global attribute

eymar:
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

closes https://github.com/JetBrains/compose-jb/issues/1338

added `fun inputMode(...)` to common AttrsBuilder because it's global according to docs, even though it most likely to be used with inputs, although can be used with other contenteditable elements.


Schahen (approved):
LGTM )

PR #1462, opened 2021-11-25
Throw exception for `String + CSSSelector` with `self` cases.

Skolotsky:


web/core/src/jsTest/kotlin/CSSStylesheetTests.kt
```diff
@@ -311,4 +309,22 @@ class CSSVariableTests {
             "Sibling selector isn't generated correctly"
         )
     }
+
+    @Test
+    fun testStringPlusCSSSelectorConcatenation() {
+        assertTrue("Concatenation of String + CSSSelector should be restricted.") {
```
eymar, line 315: you can use assertFailsWith() in such cases if you like

Skolotsky: didn't know, thanks

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/selectors/CSSSelectors.kt
```diff
@@ -23,13 +23,18 @@ sealed class Nth {
...
     }
 
     internal open fun contains(other: CSSSelector, strict: Boolean = false): Boolean {
         return if (strict) this === other else this == other
     }
 
+    // This method made for workaround because of possible concatenation of `String + CSSSelector`,
+    // so `toString` is called for such operator, but we are calling `asString` for instantiation.
+    // `toString` is reloaded for CSSSelfSelector
+    internal open fun asString(): String = toString()
```
eymar, line 36: not a suggestion, but to understand better:

can CSSSelector be abstract so this asString() is abstract too?
It seems that in this case `toString()` wont' return anything meaningful. Is it right?

eymar: anyway, what do you think about making CSSSelector abstract?
Can CSSSelector be instantiated on its own? If yes, It seems I don't know about use cases when it's needed. Could you please tell me?

eymar: of course this question is not directly related to this PR. just asking

Skolotsky: it is good idea to make CSSSelector abastract

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheet.kt
```diff
@@ -80,7 +80,8 @@ open class StyleSheet(
 
     @Suppress("EqualsOrHashCode")
     class CSSSelfSelector(var selector: CSSSelector? = null) : CSSSelector() {
-        override fun toString(): String = selector.toString()
+        override fun toString(): String = throw IllegalStateException("You can't concatenate `String + CSSSelector` which contains `self` or `root`. Use `selector(<your string>)` to convert `String` to `CSSSelector` for proper work.")
+        override fun asString(): String = selector?.asString() ?: throw IllegalStateException("You can't instantiate self without actual selector")
```
Schahen, line 84: Two question here
1) (just to see a bigger picture if we can not have nullable selector in any practical scenario why should we allow passing nullable selector at all? When we can end up in such situation

2) Even if we'll have null there for any reason - can we just cast  it with !! so that we fail - we'll let user that something is null why we are not expecting it to be null but user will see the standard message they got used to see.

web/core/src/jsTest/kotlin/CSSStylesheetTests.kt
Schahen: +1 for assertFailsWith

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/selectors/CSSSelectors.kt
```diff
@@ -23,13 +23,18 @@ sealed class Nth {
 
 open class CSSSelector {
     override fun equals(other: Any?): Boolean {
-        return this === other || toString() == other.toString()
+        return this === other || asString() == (other as? CSSSelector)?.asString()
     }
```
Schahen, line 27: This is irrelevant to the PR itself but I want to use this as a chance to refresh a bit the context of having this equals at all - do we anywhere in code rely on its existence?

Schahen (commented):
I'm very not sure about the 
`You can't instantiate self without actual selector` exception as a new part of contract (but this is subject to discussion) and join Sasha in request to use standard API for checking an exception.

Also  - may be we even can add link in exception to the actual issue on github to make it even clearer for users what's going on. Apart from that, has nothing more to add.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/selectors/CSSSelectors.kt
Skolotsky: It's because i had plan to make our css reactive - so we can compare selectors equality. At least.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheet.kt
Skolotsky: 1. It's bacause some complicated experimantal scenario with "auto" css, which doesn't know instance of self until instantiation and generating class for it. I just reused this class for this case too.
2. It's more for us, not for users... In Space we have another policy - do not use `!!`, but write custom understandable exceptions

eymar (approved):

PR #1470, opened 2021-11-26
web: remove Options for event listeners

eymar:
Options were never actually applied and there was no way to use anything beside Options.DEFAULT

We had a discussion about this some time ago. The conclusion was that it's likely redundant parameter for most of the cases.
Anyway It's still possible to set options using k/js `addEventListener` on the element's reference

Here is a related PR that was not merged: https://github.com/JetBrains/compose-jb/pull/860#issuecomment-875685792


Skolotsky (approved):

Schahen (approved):
This is an example of something that in theory a breaking (but needed and totally justified) change that on practice won't break anything. 

Thank you for this change!

PR #1475, opened 2021-11-26
web: add test-utils tutorial

eymar:
Please have a look at the tutorial for test-utils for compose-dom components.
These utils allow users to write unit-tests for their `@Composable` dom components.


eymar:
Merging it now, but please let me know if anything needs to be improved

PR #1504, opened 2021-11-29
web (breaking change): move top level functions for css selectors to new SelectorsScope

eymar:
These functions don't need to be on top level because their purpose is to be used within StyleSheet


Skolotsky (approved):

Schahen (approved):
LGTM

PR #1505, opened 2021-11-29
web (breaking change): move CSSUnit convertors functions and CSS operations (on units) to their new scopes

eymar:
The intention is to avoid having these functions on top level. They are usually used within specific scopes, so it makes sense to limit the number of places where it's possible to call them.


eymar:
Not likely to merge now. This change prevents such cases:
```
// somewhere not in the scope of CSSUnits
val a = baseMargin + 8.px
```

this can be solved by something like:
```
val a = cssUnits { baseMargin + 8.px }
```

We'll consider this and possible other options but not for 1.0

Skolotsky (commented):
I think we need to postpone such change until understanding of state with skiko.

Schahen:
I don't think we should have this one to be honest. Yep, some thing don't have much sense in wrong context - but this is true about anything. 

Suggesting to use 
```kotlin
val a = cssUnits { baseMargin + 8.px }

```

instead of 
```
val a = baseMargin + 8.px
```

Is not justified enough

eymar:
closing this for now

PR #1510, opened 2021-11-30
web (breaking change): make constructors of CSSSelectors classes internal

eymar:
this is a follow for #1504


web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/selectors/CSSSelectors.kt
```diff
@@ -136,96 +136,139 @@ abstract class CSSSelector {
 
         companion object {
             // Location pseudo-classes
+            @Deprecated("Consider using a property from SelectorsScope")
```
Schahen, line 139: let's put this message to a const val

eymar: done

PR #1511, opened 2021-11-30
Modify replace.sh for replacing Compose version

igordmn:
Remove:
```
__LATEST_COMPOSE_RELEASE_VERSION__
__KOTLIN_COMPOSE_VERSION__
System.getenv("COMPOSE_TEMPLATE_COMPOSE_VERSION")
```

They pollute templates/examples.

Now, all paths where we need to change the version are hardcoded in the script.

Usage:
```
./replace.sh 1.0.0-rc6
```

In the future I will add support for changing Kotlin version


examples/falling-balls-web/build.gradle.kts
```diff
@@ -4,7 +4,7 @@ import org.jetbrains.compose.desktop.application.dsl.TargetFormat
 
 plugins {
     kotlin("multiplatform") version "1.5.31"
-    id("org.jetbrains.compose") version "1.0.0-beta5"
```
olonho, line 7: What's this change?

igordmn: My mistake, thanks ).

Fixed

examples/falling-balls/settings.gradle.kts
```diff
@@ -1 +1,8 @@
+pluginManagement {
+    repositories {
+        gradlePluginPortal()
+        maven("https://maven.pkg.jetbrains.space/public/p/compose/dev")
```
olonho, line 4: I intentionally removed that earlier to have a test that works without custom settings. Why it was needed?

examples/intellij-plugin/build.gradle.kts
```diff
@@ -4,8 +4,7 @@ plugins {
     id("org.jetbrains.intellij") version "1.1.4"
     java
     kotlin("jvm") version "1.5.31"
-    // __LATEST_COMPOSE_RELEASE_VERSION__
```
olonho, line 7: Please do not remove tag, it's intended to ensure that we never miss an update.

experimental/lwjgl-integration/build.gradle.kts
```diff
@@ -2,10 +2,8 @@ import org.jetbrains.compose.compose
 import org.jetbrains.compose.desktop.application.dsl.TargetFormat
 
 plugins {
-    // __KOTLIN_COMPOSE_VERSION__
     kotlin("jvm") version "1.5.31"
-    // __LATEST_COMPOSE_RELEASE_VERSION__
-    id("org.jetbrains.compose") version (System.getenv("COMPOSE_TEMPLATE_COMPOSE_VERSION") ?: "1.0.0-alpha4-build411")
+    id("org.jetbrains.compose") version "1.0.0-alpha4-build411"
```
olonho, line 6: Better use another version and keep tag.

tools/replace.sh
```diff
@@ -1,13 +1,57 @@
...
-APPCOMPAT_NEWVER=1.3.0-beta01
-find -E $ROOT  -regex '.*\.(kts|properties|kt)' -exec sed -i '' -e "s/$APPCOMPAT_OLDVER/$APPCOMPAT_NEWVER/g" {} \;
-KOTLIN_OLDVER=1.5.10
-KOTLIN_NEWVER=1.5.10
-find -E $ROOT  -regex '.*\.(kts|properties|kt)' -exec sed -i '' -e "s/$KOTLIN_OLDVER/$KOTLIN_NEWVER/g" {} \;
-git grep -C 1 __KOTLIN_COMPOSE_VERSION__ $ROOT
-git grep -C 1 __LATEST_COMPOSE_RELEASE_VERSION__ $ROOT
+
+declare -a files=(
+  "templates/desktop-template/build.gradle.kts"
```
olonho, line 8: How we ensure that all files are in this list?

examples/intellij-plugin/build.gradle.kts
igordmn: Now we only specify folders, and iterate through all files in it. If we would introduce a new pattern (which is rare), we would add it to `replace.sh`

jamiecrowther (approved):

examples/intellij-plugin/build.gradle.kts
olonho: Yep, approach with properties seems right and this way tag is not needed.

olonho (approved):

Issue #1490, opened 2021-11-28
Fix background for widgets example (dark theme)


igordmn:
Compose 1.0.0-rc5, Windows 11

When the system in dark theme, the background is still white.

Because of that some elements are invisible:

![image](https://user-images.githubusercontent.com/5963351/143783237-a21bfc35-e897-4d5b-9cd4-0ce1628e4269.png)
![image](https://user-images.githubusercontent.com/5963351/143783241-94c549c7-cfda-40f1-a6cd-30564165f52e.png)

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-5140">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1536, opened 2021-12-02
Fix dark theme in widgets example

igordmn:
Fixes https://github.com/JetBrains/compose-jb/issues/1490

image


olonho (approved):
Interesting why this change fixes theming?

igordmn:
> Interesting why this change fixes theming?

Material theme applies only to the material components.

In widgets example we didn't have "background" component, so it was always white.

PR #1549, opened 2021-12-06
Plugin path resolution customization

Schahen:
The goal of following commit is to be able to read a list of local maven locations for plugin resolution. For example, if one needs to look for plugin location in common maven folder and  in, say, `/androidx-main/out/.gradle/.m2/repository`, this can be achieved by putting following code to the local.properties:

```
compose.web.repos = /home/shagen/.m2/repository/:/androidx-main/out/.gradle/.m2/repository
```

### DIFF
--- a/web/settings.gradle.kts
+++ b/web/settings.gradle.kts
@@ -1,23 +1,49 @@
 
 pluginManagement {
     val COMPOSE_CORE_VERSION: String by settings
+    println("[build] compose core version: $COMPOSE_CORE_VERSION")
+
+    // pluginManagement section won't see outer scope, hence the FQ names
+    fun properties(path: String): java.util.Properties? {
+        val localPropertiesFile = File(path)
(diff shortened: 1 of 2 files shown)


web/settings.gradle.kts
```diff
@@ -1,22 +1,50 @@
...
             url = uri("https://packages.jetbrains.team/maven/p/ui/dev")
         }
+
         google()
     }
 
     resolutionStrategy {
         eachPlugin {
             if (requested.id.id == "org.jetbrains.compose") {
-                println("[build] compose core version: ${extra["COMPOSE_CORE_VERSION"]}")
```
eymar, line 18: is it okay to keep this println? sometimes it helps  to understand that I'm using correct version.

Schahen: This message just moved upper (and printed once) since we have common plugin resolution for all subprojects.

eymar (approved):

PR #1590, opened 2021-12-14
release 1.0.1-rc2

eymar:
uses kotlin 1.6.10


hfhbd:
Wow, nice.
But what about updating to 1.6.10 release version? There should no differences between RC and release version.

eymar:
@hfhbd 
yes, I noticed that 1.6.10 is out only after I created this PR. We'll build with 1.6.10 soon

PR #1592, opened 2021-12-14
web: increase timeout for tests (in karma.config.common.d)

eymar:


Schahen (changes requested):
My only request is to add semicolons since it's still js. 
And I apologize for being picky )

Schahen (approved):

Issue #1339, opened 2021-10-31
Issue with transparency on Linux


DevSrSouza:
I'm having couple issues with transparency support on Linux
- At my second monitor (4k one) it fails to setup SkiaLayer with a exception (in thread).
- When It get open at my primary monitor (2k) it have a poor performance, when I switch my second monitor to 2k the performance get a lot better. (the application is still being displayed at the first monitor)
- When I try to move the window to the second monitor (SUPER + SHIFT + LEFT) it gives my the exception in the first case and stop working.
- When moving from extend to mirror screens, the issue does not appear.
- When moving from single monitor to extend screen with the application is open, the issue do not appear.
- Using `WindowDraggableArea` also breaks the application if it was running previously.

System information:
- Pop!_OS 21.04 x86_64
- JDK 16
- Gnome version 3.38.5
- X11

Exception:
```
Cannot create Linux GL context
Exception in thread "AWT-EventQueue-0" java.lang.IllegalStateException: SkiaLayer is disposed
	at org.jetbrains.skiko.SkiaLayer.update$skiko(SkiaLayer.jvm.kt:408)
	at org.jetbrains.skiko.redrawer.AbstractDirectSoftwareRedrawer.redrawImmediately(AbstractDirectSoftwareRedrawer.kt:36)
	at org.jetbrains.skiko.redrawer.LinuxSoftwareRedrawer.redrawImmediately(LinuxSoftwareRedrawer.kt:24)
	at org.jetbrains.skiko.SkiaLayer.paint(SkiaLayer.jvm.kt:317)
	at androidx.compose.ui.awt.ComposeLayer$ComponentImpl.paint(ComposeLayer.desktop.kt:100)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:907)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1083)
	at java.desktop/javax.swing.JLayeredPane.paint(JLayeredPane.java:586)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:907)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1083)
	at java.desktop/javax.swing.JComponent.paintChildren(JComponent.java:907)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1083)
	at java.desktop/javax.swing.JLayeredPane.paint(JLayeredPane.java:586)
[...]

PR #1601, opened 2021-12-15
Adding transparent windows information to the Window API tutorial

akurasov:
Also closes https://github.com/JetBrains/compose-jb/issues/1339

### DIFF
--- a/tutorials/Window_API_new/README.md
+++ b/tutorials/Window_API_new/README.md
@@ -607,7 +607,7 @@ private fun WindowScope.AppWindowTitleBar() = WindowDraggableArea {
 <img alt="Draggable area" src="draggable_area.gif" height="239" />
 
 ## Transparent windows (e.g. allows to make windows of a custom form)
-To create a transparent window it is enough to pass two parameners to the Window function: transparent=true and undecorate=true (it is not possible to decorate a transparent Window). Common scenario is to combine transparent window with a Surface of a custom form. Below is an example of a round-cornered Window. 
+To create a transparent window it is enough to pass two parameners to the Window function: `transparent=true` and `undecorate=true` (it is not possible to decorate a transparent Window). Common scenario is to combine transparent window with a Surface of a custom form. Below is an example of a round-cornered Window. 
 
 ```kotlin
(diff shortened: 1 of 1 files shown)


tutorials/Window_API_new/README.md
```diff
@@ -605,3 +605,46 @@ private fun WindowScope.AppWindowTitleBar() = WindowDraggableArea {
...
+                shape = RoundedCornerShape(20.dp) //window has round corners now
+            ) {
+                Text("Hello World!", color = Color.White)
+            }
+        }
+    }
+}
+```
+
+_Important note: Window transparency is implemented based on JDK implementation, that contains known issue in case of moving a Window between two monitors with different density. So when you move an App, the Window stops being transparent. And it seems nothing can be done with this situation on Compose side.
```
Rsedaikin, line 649: This is a Linux only issue. Everything is fine on macOS and Windows.

tutorials/Window_API_new/README.md
```diff
@@ -605,3 +605,46 @@ private fun WindowScope.AppWindowTitleBar() = WindowDraggableArea {
 }
 ```
 <img alt="Draggable area" src="draggable_area.gif" height="239" />
+
+## Transparent windows (e.g. allows to make windows of a custom form)
+To create a transparent window it is enough to pass two parameners to the Window function: transparent=true and undecorate=true (it is not possible to decorate a transparent Window). Common scenario is to combine transparent window with a Surface of a custom form. Below is an example of a round-cornered Window.
```
Rsedaikin, line 610: I think we should quote parameters/methodы/classes and other appropriate names. Like
` transparent=true`

Rsedaikin (changes requested):

tutorials/Window_API_new/README.md
akurasov: fixed

tutorials/Window_API_new/README.md
akurasov: fixed

Rsedaikin (approved):

Issue #1606, opened 2021-12-15
Sample TodoApp Web app does not compile 


arkivanov:
```
Could not determine the dependencies of task ':common:utils:jsTestPackageJson'.
> Could not resolve all dependencies for configuration ':common:utils:jsTestNpm'.
   > Could not find org.jetbrains.kotlin:kotlin-test-common:null.
```

Looks like it was broken in #1511. I will try to fix it in the coming days.

igordmn:
Should be fixed in https://github.com/JetBrains/compose-jb/pull/1607#issuecomment-995253256

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-5119">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1607, opened 2021-12-15
Fix Todo example Deps

igordmn:
task `gradlew jsBrowserDevelopmentRun` didn't work

### DIFF
--- a/examples/todoapp/buildSrc/gradle.properties
+++ b/examples/todoapp/buildSrc/gradle.properties
@@ -1,2 +1,3 @@
+# TODO can we get rid of duplication with root gradle.properties?
 kotlin.version=1.6.10
 compose.version=1.0.1-rc2
--- a/examples/todoapp/gradle.properties
+++ b/examples/todoapp/gradle.properties
@@ -22,3 +22,6 @@ kotlin.code.style=official
 org.gradle.parallel=true
 org.gradle.caching=true
 kotlin.native.disableCompilerDaemon=true
+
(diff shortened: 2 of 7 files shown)


igordmn:
сс @arkivanov

igordmn:
Fixes https://github.com/JetBrains/compose-jb/issues/1606

examples/todoapp/buildSrc/gradle.properties
```diff
@@ -1,2 +1,3 @@
+# TODO can we get rid of duplication with root gradle.properties?
```
arkivanov, line 1: Perhaps migrating to "Version catalogs" may help: https://docs.gradle.org/current/userguide/platforms.html

examples/todoapp/gradle.properties
```diff
@@ -22,3 +22,6 @@ kotlin.code.style=official
 org.gradle.parallel=true
 org.gradle.caching=true
 kotlin.native.disableCompilerDaemon=true
+
+kotlin.version=1.6.10
```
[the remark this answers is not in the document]
igordmn: Yes, but this change also simplifies accessing Deps.

examples/todoapp/buildSrc/gradle.properties
igordmn: But it is not easy to override the version in version catalog via an argument in command line.

arkivanov:
In this case LGTM

Issue #1538, opened 2021-12-03
CSS nth-child(an+b) cannot be used


mee1080:
Nth.Functional doesn't inherit Nth, and `nthChild(Nth.Functional(2,1))` is an error.

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-6145">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1609, opened 2021-12-16
Fix for Nth.Functional

Schahen:
see #1538

### DIFF
--- a/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/selectors/CSSSelectors.kt
+++ b/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/selectors/CSSSelectors.kt
@@ -8,7 +8,7 @@ internal const val webCssSelectorsDeprecationMessage = "Consider using a propert
 private val selectorScope = object : SelectorsScope {}
 
 sealed class Nth {
-    data class Functional(val a: Int? = null, val b: Int? = null) {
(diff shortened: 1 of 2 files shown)


web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/selectors/CSSSelectors.kt
```diff
@@ -7,19 +7,19 @@ import org.jetbrains.compose.web.css.SelectorsScope
 internal const val webCssSelectorsDeprecationMessage = "Consider using a property from SelectorsScope"
 private val selectorScope = object : SelectorsScope {}
 
-sealed class Nth {
-    data class Functional(val a: Int? = null, val b: Int? = null) {
+sealed interface Nth {
+    data class Functional(val a: Int? = null, val b: Int? = null) : Nth {
```
eymar, line 11: Not for this PR, but might be good to decide before next build:

since previous API was not very useful (didn't work as expected), can we make it internal and provide corresponding public functions in SelectorsScope?

Schahen: I suggest not to do it in this PR since it introduce breaking changes (while this one just fix current API) but I totally agree-  it's just that such selctor API steel need to be designed.

eymar (approved):

PR #1612, opened 2021-12-16
Add option to set macOS build version

### LINKED ISSUE #1611: Add option to set macOS build version
Add option to set macOS build version in addition to the release version.

Currently you can set the version using `packageVersion`, which will set both `CFBundleShortVersionString` and `CFBundleVersion`. There is no way to set the build version (`CFBundleVersion`) to a different number.

The build version is useful for TestFlight. This way you can increment the build version (`CFBundleVersion`) without incrementing the release (public) version (`CFBundleShortVersionString`) for updates on TestFlight. (And once there is a stable build it can be pushed to the App Store and users only see the release version).

I will create a PR which adds support for setting the build version to a different version.

Thomas-Vos:
Fixes #1611

Currently, the build version is set to the same version as the `packageVersion` property. This PR adds a new option, called `packageBuildVersion` which can be used to customise the build version. If this option is not specified, it will fallback to `packageVersion` for backwards compatibility.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
@@ -129,6 +129,10 @@ abstract class AbstractJPackageTask @Inject constructor(
 
     @get:Input
     @get:Optional
+    val packageBuildVersion: Property<String?> = objects.nullableProperty()
+
+    @get:Input
+    @get:Optional
     val winConsole: Property<Boolean?> = objects.nullableProperty()
 
     @get:Input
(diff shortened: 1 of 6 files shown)


olonho:
@AlexeyTsvetkov please review

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
```diff
@@ -477,7 +481,8 @@ abstract class AbstractJPackageTask @Inject constructor(
         val packageVersion = packageVersion.get()!!
         plist[PlistKeys.CFBundleShortVersionString] = packageVersion
         plist[PlistKeys.LSApplicationCategoryType] = "Unknown"
-        plist[PlistKeys.CFBundleVersion] = packageVersion
+        val packageBuildVersion = packageBuildVersion.get()!!
```
AlexeyTsvetkov, line 484: I think `packageBuildVersion.get()!!` should be replaced with `packageBuildVersion.orNull`, since the property is marked as optional (which is good, as it allows the task to be used separately from the default plugin configuration without bothering with setting `packageBuildVersion`)

AlexeyTsvetkov (commented):
In general the PR looks good to me.

The docs could have some more details 
on when to use this new property,
@Thomas-Vos what do you think?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
Thomas-Vos: Sounds good, I did not know this package task could be used standalone, see latest commit. It now uses `packageBuildVersion.orNull` and has `packageVersion` as a fallback (for backwards compatibility and to make packageBuildVersion really optional). What do you think about this?

Thomas-Vos:
Updated the PR with a fix in `AbstractJPackageTask`, added some more documentation about the build version, and fixed some typos in the existing documentation.

AlexeyTsvetkov (approved):

AlexeyTsvetkov:
@Thomas-Vos thanks for the contribution!

Issue #1599, opened 2021-12-15
Upload to Apple App Store (Universal macOS Binary required, signing issues)


Thomas-Vos:
To be able to upload a Compose for Desktop app to the Apple App Store, there are multiple issues that need to be resolved.

- [ ] Universal macOS Binary (optional)
- [x] Signing for Mac App Store (required) #1613 
- [x] Fix TestFlight builds (optional) #1812

# Universal macOS Binary (optional)

To be able to support both Intel and ARM64 Macs natively, there needs to be a way to build a Universal pkg (Intel + ARM64). That is needed to be able to upload it to the Apple App Store. The only alternative for now is to upload an Intel-only app.

<img width="502" alt="Screenshot 2021-12-15 at 19 56 49" src="https://user-images.githubusercontent.com/11077217/146248046-e7b19ec7-ba1e-4167-8590-eba5e6a231f1.png">

I tried adding both `compose.desktop.macos_arm64` and `compose.desktop.macos_x64` dependencies as suggested on Slack. However, the produced app is still not Universal. 

I am building the project on an M1 Mac with an ARM64 JDK. When doing a right click and get info on the produced `.app` it shows Apple Silicon instead of Universal.

Furthermore, I am not sure what to do about the bundled JRE. Would I need to include it twice (for Intel and for ARM64) or so?

Slack discussion: https://kotlinlang.slack.com/archives/C01D6HTPATV/p1639570539349100

Found some information on this here: https://bugs.openjdk.java.net/browse/JDK-8266259

Thomas-Vos:
Just wanted to add that it is not technically mandatory for the AppStore, as I think you can still upload Intel only apps at this time. However, if you want the app to run natively on M1 Macs, you are required to build a Universal app with support for both architectures.

Thomas-Vos:
I am working on support for the Apple App Store. Could use some guidance on this if possible. See #1613

Thomas-Vos:
Does anyone have any idea how to fix the univeral binary issue? Intel and ARM64 in one binary.

Thomas-Vos:
It is now possible to upload apps to the App Store now that https://github.com/JetBrains/compose-jb/pull/1613 is merged. However, TestFlight is not working and universal binaries are not supported. Any suggestions?

Thomas-Vos:
I opened a TSI with Apple and am working with them to get the TestFligth issues resolved. They were able to provide me with more detailed error message, so I am working on that.

[...]

PR #1613, opened 2021-12-16
Apple App Store

Thomas-Vos:
This is a work-in-progress PR which adds support for publishing to the Apple App Store. For more information see #1599.

This PR adds the following:
- `appStore` option which passes the `--mac-app-store` option to jpackage.
- `appCategory` option which passes the `--mac-app-category category` option to jpackage and sets the `LSApplicationCategoryType` value in `Info.plist`.
- Possibly a fix for the `--app-image` option? I needed to change this to get the build working.

# Notes
* At least JDK 17 is required (https://bugs.openjdk.java.net/browse/JDK-8248904).
* With JDK 17 I ran into some signing issues (codesign reports `libskiko-macos-x64.dylib: is already signed`). Still need to look into this further. For me JDK 18 worked fine, so I just used that for now. Any ideas? Edit: issue here: https://github.com/JetBrains/compose-jb/issues/1666
* There is a bug in macOS Montery 12.0 which causes `productbuild` to fail with `'*** -[NSMutableOrderedSet unionOrderedSet:]: ordered set argument is not an NSOrderedSet’`. Solution: update to macOS Montery 12.1. See for more information: https://github.com/xamarin/xamarin-macios/issues/13197.

# TODO
The following errors were still reported by the App Store when using Transporter, with an app built with this PR:
* > WARNING ITMS-90889: ""Cannot be used with TestFlight because the bundle at “HueEssentials.app” is missing a provisioning profile. Main bundles are expected to have provisioning profiles in order to be eligible for TestFlight.""

I would appreciate some help with the error above.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
@@ -135,6 +135,11 @@ abstract class AbstractJPackageTask @Inject constructor(
     @get:Optional
     val macAppCategory: Property<String?> = objects.nullableProperty()
 
+    @get:InputFile
+    @get:Optional
+    @get:PathSensitive(PathSensitivity.ABSOLUTE)
+    val macEntitlementsFile: RegularFileProperty = objects.fileProperty()
+
     @get:Input
     @get:Optional
     val packageBuildVersion: Property<String?> = objects.nullableProperty()
@@ -344,6 +349,7 @@ abstract class AbstractJPackageTask @Inject constructor(
                 cliArg("--mac-package-identifier", nonValidatedMacBundleID)
                 cliArg("--mac-app-store", macAppStore)
                 cliArg("--mac-app-category", macAppCategory)
+                cliArg("--mac-entitlements", macEntitlementsFile)
 
                 macSigner?.let { signer ->
                     cliArg("--mac-sign", true)
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
@@ -508,7 +508,7 @@ abstract class AbstractJPackageTask @Inject constructor(
         val packageVersion = packageVersion.get()!!
         plist[PlistKeys.CFBundleShortVersionString] = packageVersion
         // If building for the App Store, use "utilities" as default just like jpackage.
-        val category = macAppCategory.orNull ?: (if (macAppStore.get()!!) "utilities" else null)
+        val category = macAppCategory.orNull ?: (if (macAppStore.orNull == true) "utilities" else null)
         plist[PlistKeys.LSApplicationCategoryType] = category?.let { "public.app-category.$it" } ?: "Unknown"
         val packageBuildVersion = packageBuildVersion.orNull ?: packageVersion
         plist[PlistKeys.CFBundleVersion] = packageBuildVersion
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
@@ -284,7 +284,7 @@ abstract class AbstractJPackageTask @Inject constructor(
 
         if (targetFormat != TargetFormat.AppImage) {
             // Args, that can only be used, when creating an installer
-            cliArg("--app-image", appImage)
+            cliArg("--app-image", appImage.dir("${packageName.get()}.app"))
             cliArg("--install-dir", installationPath)
             cliArg("--license-file", licenseFile)
 
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
@@ -441,7 +441,7 @@ abstract class AbstractJPackageTask @Inject constructor(
                 .also { setInfoPlistValues(it) }
                 .writeToFile(jpackageResources.ioFile.resolve("Info.plist"))
 
-            if (macAppStore.get()!!) {
+            if (macAppStore.orNull == true) {
                 val productDefPlistXml = """
                     <key>os</key>
                     <array>
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
@@ -291,6 +291,7 @@ abstract class AbstractJPackageTask @Inject constructor(
             cliArg("--app-image", appImage.dir("${packageName.get()}.app"))
             cliArg("--install-dir", installationPath)
             cliArg("--license-file", licenseFile)
+            cliArg("--resource-dir", jpackageResources)
 
             when (currentOS) {
                 OS.Linux -> {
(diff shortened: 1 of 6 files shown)


olonho:
Looks interesting, but let's wait for @AlexeyTsvetkov opinion.

Thomas-Vos:
Could someone please review this PR?

olonho:
We're looking on that change and @AlexeyTsvetkov will comment on what's going on there.
@Thomas-Vos could you please rebase the change to current master?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
```diff
@@ -276,9 +284,10 @@ abstract class AbstractJPackageTask @Inject constructor(
 
         if (targetFormat != TargetFormat.AppImage) {
             // Args, that can only be used, when creating an installer
-            cliArg("--app-image", appImage)
+            cliArg("--app-image", appImage.dir("${packageName.get()}.app"))
```
AlexeyTsvetkov, line 287: This change looks suspicious. With non app store packaging previous value clearly worked.
I'm not sure, why this is required and whether it could break cases, that worked previously.
I don't doubt, that you've had a reason to add it, but I would like to double check it before merging.

AlexeyTsvetkov (changes requested):
Overall LGTM
The only thing that really concerns me is the change to the `--app-image` parameter value.
I can either merge everything else now and check that line separately, or check that line 
slightly later and merge everything together

AlexeyTsvetkov:
@Thomas-Vos 

> is missing a provisioning profile. Main bundles are expected to have provisioning profiles in order to be eligible for TestFlight

Might be related to https://bugs.openjdk.java.net/browse/JDK-8274346 which is fixed in JDK 18

> With JDK 17 I ran into some signing issues

I will push a fix soon

> Possibly a fix for the --app-image option? I needed to change this to get the build working.

Could you provide some more details on the exact issue?

Thomas-Vos:
Rebased with no changes, will look at your feedback now.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
Thomas-Vos: Without this change I get the following error:
```
...
[16:18:13.997] Running /usr/bin/productbuild
[16:18:14.015] Command [PID: 6897]:
    /usr/bin/productbuild --resources /var/folders/61/dv805tx974q7kldwx81tf9b80000gn/T/jdk.jpackage12469847631489678324/config --sign 3rd Party Mac Developer Installer: Thomas Vos (4J6Y3WBQB6) --product /var/folders/61/dv805tx974q7kldwx81tf9b80000gn/T/jdk.jpackage12469847631489678324/config/product-def.plist --component /var/folders/61/dv805tx974q7kldwx81tf9b80000gn/T/jdk.jpackage12469847631489678324/root-6402241906218587642/Hue Essentials.app /Applications /Users/thomas/AndroidStudioProjects/HueEssentials/desktop/build/compose/binaries/main/pkg/Hue Essentials-1.0.0.pkg
[16:18:14.015] Output:
    productbuild: error: The component at "/var/folders/61/dv805tx974q7kldwx81tf9b80000gn/T/jdk.jpackage12469847631489678324/root-6402241906218587642/Hue Essentials.app" does not exist or is not a directory.
[16:18:14.015] Returned: 1
...
```

By adding packageName.app to the --app-image option it works fine. Not sure why this is needed.

I have not tested the code on other platforms, but this will fail because ".app" is macOS specific. I can add this within a condition to only add this for macOS and macAppStore.

Thomas-Vos: What do you think of https://github.com/JetBrains/compose-jb/pull/1613/commits/b24cd0ddaba5ff4810f94b25caa9d7b92602aa84?

Are you able to test this? You could still exclude this change of course, but I think that would make the PR pointless as for me this is needed for it to work. Let me know what you think.

tutorials/Signing_and_notarization_on_macOS/README.md
```diff
@@ -167,19 +175,19 @@ uniquely identifies an application in Apple's ecosystem.
 macOS {
     signing { 
          sign.set(true)
-         identity.set("John Doe")
+         identity.set("John Doe (123)")
```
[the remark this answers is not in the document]
Thomas-Vos: You are right, will revert.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
```diff
@@ -420,6 +431,17 @@ abstract class AbstractJPackageTask @Inject constructor(
             InfoPlistBuilder(macExtraPlistKeysRawXml.orNull)
                 .also { setInfoPlistValues(it) }
                 .writeToFile(jpackageResources.ioFile.resolve("Info.plist"))
+
+            if (macAppStore.get()!!) {
```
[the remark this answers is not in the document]
Thomas-Vos: Thanks, fixed.

Thomas-Vos:
> > is missing a provisioning profile. Main bundles are expected to have provisioning profiles in order to be eligible for TestFlight
> 
> Might be related to https://bugs.openjdk.java.net/browse/JDK-8274346 which is fixed in JDK 18

Seems like it, however with JDK 18 I still get the error. I think you may need to include a provisioning profile manually and include it using the JDK 18 option `--app-content`. Not familiar with this, will have to look into it.

> 
> > Possibly a fix for the --app-image option? I needed to change this to get the build working.
> 
> Could you provide some more details on the exact issue?

https://github.com/JetBrains/compose-jb/pull/1613#discussion_r784064784

Thomas-Vos:
Found a solution for the provisioning profile, will add support for it to the PR and update the documentation.

Thomas-Vos:
@AlexeyTsvetkov Added an option to set the provisioning profile. Please let me know what you think about this.

No errors anymore when uploading to the App Store:

Screenshot 2022-01-13 at 19 03 29


However, TestFlight still shows "Not Available for Testing" (see screenshot). Not sure why, am looking into it. I guess something is still wrong with the provisioning profile or so, even though there were no errors or warnings uploading.

Screenshot 2022-01-13 at 19 07 06

Thomas-Vos:
I was not able to figure out the error in TestFlight, any suggestions?

I think that publishing an app to production on the App Store should work now with this PR, but haven't tried that.

For my project I need to use TestFlight.

Also added another option so the entitlements file can be customised.

@AlexeyTsvetkov Let me know if you have any feedback, thanks for reviewing!

Thomas-Vos:
By the way, if you want to test this on an M1 Mac, you need to build an Intel binary for the App Store (TODO: find a way to build universal binaries). To do this, you need to use a macOS / x64 (Intel) build of the JDK. And for building your Compose project use the following terminal command:
```sh
arch -x86_64 ./gradlew packagePkg
```

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
Thomas-Vos: @AlexeyTsvetkov FYI, when creating a pkg using packagePkg, without this PR, after installation on my mac it is installed like this:
Screenshot 2022-01-14 at 20 33 57

Think there is indeed an issue in compose. But maybe it's only for the newer JDK version, not sure. This was with JDK 17.

AlexeyTsvetkov (approved):

AlexeyTsvetkov:
> I was not able to figure out the error in TestFlight, any suggestions?

Sadly I don't have any at the moment. I will try to look into it, but no ideas so far.

@Thomas-Vos 
Overall the PR looks good to me now.
Again, sorry for all the waiting.

This is a great and welcome contribution,
thank you for your efforts!

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
Thomas-Vos: Created issue here: https://github.com/JetBrains/compose-jb/issues/1712

Thomas-Vos:
Thanks a lot for the feedback and merging this PR, @AlexeyTsvetkov. Will keep looking into the TestFlight issue, maybe I will open a TSI with Apple for this and ask them what is wrong with the uploaded package. For the incorrect app image path, see https://github.com/JetBrains/compose-jb/issues/1712.

PR #1624, opened 2021-12-20
Adding obfuscation info

akurasov:


tutorials/Native_distributions_and_local_execution/README.md
```diff
@@ -544,3 +544,9 @@ fun main() {
 ```
 3. Run `./gradlew runDistributable`.
 4. Links like `compose://foo/bar` are now redirected from a browser to your application.
+
+## Obfuscation
+    
+There's nothing special with obfuscation of CfD programs - it's done the same was as for any other JVM program. Compose plugin provides task
```
olonho, line 550: Please rewrite like: 

To obfuscate Compose Multiplatform JVM applications standard approach for JVM application works.  Using task `packageUberJarForCurrentOS` one could generate  JAR file which could be later obfuscated using ProGuard or R8, see for example https://stackoverflow.com/questions/64355998/proguard-example-for-gradle-java-application.

Also example using Kotlin DSL would be better, as above link uses Groovy.

akurasov: done

PR #1636, opened 2021-12-23
Minimal demo of assigning style properties via  just passing js object

Schahen:


web/core/build.gradle.kts
```diff
@@ -31,6 +31,7 @@ kotlin {
             dependencies {
                 implementation(project(":internal-web-core-runtime"))
                 implementation(kotlin("stdlib-js"))
+                implementation("org.jetbrains.kotlin-wrappers:kotlin-csstype:3.0.10-pre.283-kotlin-1.6.10")
```
eymar, line 34: do we need this dependency in web/core?

I thought it's going to be added by users in their projects.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsBuilder.kt
```diff
@@ -43,6 +48,14 @@ open class AttrsBuilder<TElement : Element> : EventsListenerBuilder() {
         styleBuilder.apply(builder)
     }
 
+    fun style(propertyHolder: Any) {
```
eymar, line 51: maybe adding one of our @Experimental annotation can be beneficial in this case.

web/core/build.gradle.kts
Schahen: Your understanding is 100% correct - just like I said at the meeting it's a fake PR - nobody is going to merge this one - it was just to illustrate so that we check whether we need this (and this dependency was added for test)

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsBuilder.kt
```diff
@@ -43,6 +48,14 @@ open class AttrsBuilder<TElement : Element> : EventsListenerBuilder() {
...
+        styleBuilder.apply {
+            Object.entries(propertyHolder).forEach { (k, v) ->
```
eymar, line 53: Did you want to try with `Object.assign()`? 
I seems that with `Object.assign()` we can avoid having `forEach` loop.

web/core/build.gradle.kts
eymar: got it, thanks! :) 
i forgot that it's a fake PR

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsBuilder.kt
eymar: UPD: not now, since this PR won't be merged for now

PR #1645, opened 2021-12-28
Fix MPP CI issues

AlexeyTsvetkov:


olonho (commented):
Let's merge for now, but better move version setter to Compose Gradle plugin.

olonho (approved):

Issue #1602, opened 2021-12-15
Docs/Samples for adding Compose Web into a legacy JS/TS app migration


StokeMasterJack:
Greenfields apps are great. But many users (myself included) have large legacy JavaScript (or TypeScript) apps.

I have a large legacy app with webpack, typescript, react and mui on the client. 

It would be great if there were docs or samples that show how one might gradually introduce Compose Web into this existing web app.

eymar:
@StokeMasterJack 

We've added an example here https://github.com/JetBrains/compose-jb/tree/master/examples/web-compose-in-js 
Please let us know if it helps or not.

akurasov:
Closing as done

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-5123">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1677, opened 2022-01-10
web: add example for dom-based-composables usage in plain JS

eymar:
this example should help with https://github.com/JetBrains/compose-jb/issues/1602


akurasov:
Shouldn't it be in the examples folder?

akurasov:
Maybe it also worth to make html a bit more complicated, to use Compose not from the root, but inside specific div - this is the basic migration scenario I suppose

eymar:
> Shouldn't it be in the examples folder?

yes, it should be in examples. I'll move it

akurasov (approved):

PR #1690, opened 2022-01-12
Add mavenCentral() to pluginManagement

avdim:
Before this changes:
On clean build error occurs: Could not get resource "...kotlin-compiler-embeddable/1.5.10/...jar"
(Reproduced on MacOS, jvm11)

Add mavenCentral() to fix problem

### DIFF
--- a/gradle-plugins/settings.gradle.kts
+++ b/gradle-plugins/settings.gradle.kts
@@ -1,6 +1,7 @@
 pluginManagement {
     repositories {
         gradlePluginPortal()
+        mavenCentral()
     }
 }
 


avdim:
To reproduce:
rm -rf ~/.gradle
rm -rf ~/.m2

in project dir:
cd compose (follow instructions on README.md)
./scripts/publishToMavenLocal -Pcompose.platforms=js

avdim:
I think the problem is related to disabling jcenter.
The Gradle plugin portal uses jcenter as a mirror.
You can read more here https://blog.gradle.org/jcenter-shutdown.
But If you explicitly specify mavenCentral, then the problem is solved.

AlexeyTsvetkov:
I find it weird, that in the comment you're mentioning `./scripts/publishToMavenLocal`, but the change is made to the gradle plugin settings.gradle.kts, which is not used by `publishToMavenLocal`.
Anyway the change is fine

AlexeyTsvetkov:
Thanks for the contribution!

avdim:
> I find it weird, that in the comment you're mentioning `./scripts/publishToMavenLocal`, but the change is made to the gradle plugin settings.gradle.kts, which is not used by `publishToMavenLocal`. Anyway the change is fine

Sorry, I mean ./scripts/publishGradlePluginToMavenLocal

PR #1694, opened 2022-01-13
AttrsScope interface extraction

### LINKED ISSUE #1651: [Web] Make AttrsBuilder<T> extendable
Currently `AttrsBuilder<T>` is declared as an `open class` as opposed to an `interface` like `ElementScope<T>`. Not only this breaks consistency, but also prevents from extending `AttrsBuilder<T>` via delegated implementation to inject your own custom scope.

Consider this scenario, where as a compose library developer, I want to make certain extensions available in `attrs = {}` argument for my custom composable which is then proxied into an HTML composable. My extensions should only be available for my custom composable as opposed to all `Button()` composables.
```kotlin
class MyCustomAttrsBuilder: AttrsBuilder<HTMLButton>() {
  fun onCustomEvent(listener: () -> Unit) {
    addEventListener("MyCustomEventName") { listener() }
  }
}

@Composable
fun MyCustomComponent(attrs: MyCustomAttrsBuilder.() -> Unit) {
  Button(
    attrs = {
      val custom = MyCustomAttrsBuilder().apply(attrs) // Consumer's attrs are applied to my own instance of AttrsBuilder with no way to proxy them onto `this` AttrsBuilder instance.
      this.copyFrom(custom) // <------- Potential solution
    }
  )
}
```

Now since `ElementScope<T>` is an interface, the above problem can be solved quite easily with delegated implementation
```kotlin
class MyCustomElementScope(originalScope: ElementScope<HTMLButton>): ElementScope<HTMLButton> by originalScope {
  @Composable
  fun CustomSubComponent() {
    Div()
  }
}

@Composable
fun MyCustomComponent(content: @Composable MyCustomElementScope.() -> Unit) {
  Button(
    content = {
      MyCustomElementScope(this).content() // Works just fine, because all base methods will be delegated to `this`
    }
  )
}
```

Ideally, both, `AttrsBuilder` and `ElementScope` should be exposed to HTML composable argument builder lambdas as interfaces.

On a side note, `AttrsBuilder` should probably be renamed to `AttrsScope` to be more discoverable and consistent with `ElementScope`. This rename would also allow keeping the current `AttrsBuilder` class untouched (except for making it implement the new interface) and in-use behind the scenes, only changing the types at the HTML composables, therefore making all old code using `AttrsBuilder` still valid as it implements the new `AttrsScope` interface.

mpetuska:
Closes #1651 

I'm sure this needs some refinement as I had to expose some internal properties as public, but I'd appreciate some initial feedback on this. Note that extracting AttrsBuilder into an interface took me down the rabbit hole and I ended up making ScopeBuilder and EventListenerBuilder into interfaces too.


web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Elements.kt
```diff
@@ -60,7 +60,12 @@ import org.w3c.dom.HTMLVideoElement
 import org.w3c.dom.Text
 import org.w3c.dom.css.CSSStyleSheet
 
-typealias AttrBuilderContext<T> = AttrsBuilder<T>.() -> Unit
+@Deprecated(
+    message = "Renamed to AttrsBuilder<T>",
+    replaceWith = ReplaceWith("AttrsBuilder<T>", "org.jetbrains.compose.web.dom.AttrsBuilder")
+)
+typealias AttrBuilderContext<T> = AttrsBuilder<T>
+typealias AttrsBuilder<T> = AttrsScope<T>.() -> Unit
```
eymar, line 68: I think I see your idea is to have consistent naming: AttrsBuilder and ContentBuilder are both lambdas with corresponding context.

at the same time, shouldn't `typealias AttrsBuilder<T>` be an alias for a new AttrsScope (or even AttrsScopeBuilder)? so it still can be instantiated/used in cases when someone already uses old AttrsBuilder directly in their code? 

And if yes, then it would need to be in `org.jetbrains.compose.web.attributes` package to keep it compatible with existing codebases (in external projects). And then typealias AttrsBuilder can be marked Depreacted. 

The idea is to make a change as less breaking as possible.

mpetuska: Good point, I'll revisit this.

web/benchmark-core/build.gradle.kts
```diff
@@ -9,8 +11,7 @@ kotlin {
         browser() {
             testTask {
                 useKarma {
-                    useChromeHeadless()
-                    //useFirefox()
+                    standardConf()
```
eymar, line 14: this one actually was intentionally kept unchanged, as it uses nonStandard karma config to increase the timeouts.
with this change, the build fails due to timeouts in tests.

mpetuska: Ah, I see. Will revert.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/builders/TextAreaAttrsScope.kt
```diff
@@ -6,44 +6,46 @@
...
 import org.jetbrains.compose.web.attributes.*
+import org.jetbrains.compose.web.attributes.EventsListenerScope.Companion.BEFOREINPUT
+import org.jetbrains.compose.web.attributes.EventsListenerScope.Companion.INPUT
 import org.jetbrains.compose.web.events.SyntheticChangeEvent
 import org.jetbrains.compose.web.events.SyntheticSelectEvent
 import org.jetbrains.compose.web.events.SyntheticInputEvent
 import org.w3c.dom.HTMLTextAreaElement
 
-class TextAreaAttrsBuilder : AttrsBuilder<HTMLTextAreaElement>() {
+class TextAreaAttrsScope : AttrsScopeBuilder<HTMLTextAreaElement>() {
```
eymar, line 16: can we have Deprecated typealiases for such cases too?

even though I think it's a rare case (that one have TestAreaAttrsBuilder instantiated in their code), it would be good if we try eliminate most cases which can break existing codebases.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
```diff
@@ -153,7 +162,7 @@ open class StyleBuilderImpl : StyleBuilder, StyleHolder {
         } else false
     }
 
-    internal fun copyFrom(sb: StyleBuilderImpl) {
```
eymar, line 156: as for exposing internal functions...
One thing we can do is to mark them `@ComposeWebInternalApi` 
I'm not sure we want to make any stability guarantees about this and similar functions which were internal, so adding an annotation is an option.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
```diff
@@ -3,34 +3,30 @@ package org.jetbrains.compose.web.attributes
...
-    internal val propertyUpdates = mutableListOf<Pair<(Element, Any) -> Unit, Any>>()
-    internal var refEffect: (DisposableEffectScope.(TElement) -> DisposableEffectResult)? = null
-
-    internal var inputControlledValueSet = false
-    internal var inputDefaultValueSet = false
-    internal var inputControlledCheckedSet = false
-    internal var inputDefaultCheckedSet = false
-
+interface AttrsScope<TElement : Element>: EventsListenerScope {
+    val attributesMap: Map<String, String>
```
eymar, line 22: Do these public val need to be in the interface? 
Can they be only in the implementation? 

they were internal, so it seems all of them are exposed to external code via functions in a controllable manner. Am I missing something?

mpetuska: I'll double check, but I only moved things into the interface one by one when I've discovered some outside usage of them (i.e. extension functions withing compose-web).

web/benchmark-core/build.gradle.kts
Schahen: This, by the way, can noticed if one will run `./gradlew clean build` ;)

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/EventsListenerScope.kt
```diff
@@ -5,191 +5,197 @@ import androidx.compose.web.events.SyntheticEvent
...
  * Its purpose is to add events to the element. For all most frequently used events there
  * exist dedicated method. In case you need to support event that doesn't have such method,
  * use [addEventListener]
  */
-open class EventsListenerBuilder {
-
-    protected val listeners = mutableListOf<SyntheticEventListener<*>>()
-
+interface EventsListenerScope {
+    val listeners: List<SyntheticEventListener<*>>
```
Schahen, line 27: A question similar to @eymar's question about attributesMap - do we need to leave ` val listeners` as part of public contract?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
Schahen: @mpetuska @eymar I suggest to leave this as exclusively StyleBuilderImpl method and `override val styleScope = StyleScopeBuilder()`

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Elements.kt
mpetuska: Ok after reviewing this I think nothing needs to change.
* Old `typealias AttrBuilderContext<T> = AttrsBuilder<T>.() -> Unit` was in this file, so it needs to stay here deprecated to maintain comatability
* New `AttrsBuilder` references the `AttrsScope` interface for consistency with `ContentBuilder` leaving `AttrsScopeBuilder` implementation hidden as an implementation detail

web/benchmark-core/build.gradle.kts
mpetuska: Reverted

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/builders/TextAreaAttrsScope.kt
mpetuska: Added deprecation typealiases for the following:
* InputAttrsBuilder
* SelectAttrsBuilder
* TextAreaAttrsBuilder

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
mpetuska: That would be ideal, however it's used in the AttrsScope interface scope ([example](http://github.com/mpetuska/compose-jb/blob/28b1dabb848d28fbc7d4c4b52e4e725b4ab57a77/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Elements.kt#L1019-L1019)) and as such needs to be declared in the interface.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
mpetuska: I've checked, and because they were internal, but on the concrete implementation class used by all the builder lambda arguments elsewhere they were fine. Now that all builders are using the interface, some internal code still needs them accessible and as such they cannot just be internal vals on the implementation as builders do not see it.

mpetuska: However such usage only occurs in your internal test suites so far.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/EventsListenerScope.kt
mpetuska: https://github.com/JetBrains/compose-jb/pull/1694#discussion_r790290835

mpetuska:
Made some tweaks and checks. Please have a look at the previous threads as well as new changes.

PR #1699, opened 2022-01-14
add BouncingBalls example to falling-balls-mpp

eymar:
Decided to add it as a separate Composable function (which can be optionally used when needed) in existing falling-balls-mpp example, to avoid increasing the number of example projects.

Screenshot 2022-01-14 at 12 52 50

### DIFF
--- a/examples/falling-balls-mpp/src/commonMain/kotlin/bouncingBalls/BouncingBalls.kt
+++ b/examples/falling-balls-mpp/src/commonMain/kotlin/bouncingBalls/BouncingBalls.kt
@@ -47,7 +47,7 @@ fun BouncingBallsApp(initialBallsCount: Int = 5) {
     val items = remember {
         val list = mutableStateListOf<BouncingBall>()
         list.addAll(generateSequence {
-            BouncingBall.nextRandomBall()
+            BouncingBall.createBouncingBall()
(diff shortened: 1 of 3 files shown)


examples/falling-balls-mpp/src/commonMain/kotlin/bouncingBalls/BouncingBalls.kt
```diff
@@ -164,13 +172,12 @@ private class BouncingBall(
         private val angles = listOf(PI / 4, -PI / 3, 3 * PI / 4, -PI / 6, -1.1 * PI)
         private val colors = listOf(Color.Red, Color.Black, Color.Green, Color.Magenta)
 
-        fun nextRandomBall(): BouncingBall {
+        fun nextRandomBall(offset: Offset? = null): BouncingBall {
```
olonho, line 175: Maybe better use another name.

olonho (approved):

examples/falling-balls-mpp/src/commonMain/kotlin/bouncingBalls/BouncingBalls.kt
eymar: done

PR #1729, opened 2022-01-21
Update Compose for Intellij plugin name and description

AlexeyTsvetkov:


tooling/compose-intellij-platform/src/main/resources/META-INF/plugin.xml
```diff
@@ -1,10 +1,17 @@
 <idea-plugin>
     <id>org.jetbrains.compose.intellij.platform</id>
-    <name>Compose for IDE Base</name>
+    <!--suppress PluginXmlValidity ("plugin" is usually not approved in the IJ marketplace, but it's OK for US) -->
```
olonho, line 3: Why this line?

AlexeyTsvetkov: Otherwise the name will be underscored red by IDE

> "plugin" is usually not approved in the IJ marketplace, but it's OK for US

In general Intellij verifier prohibits any name with the word "plugin", but I talked to the marketplace team, that we can ignore that rule in this case, because the word plugin actually makes sense here

olonho (approved):

PR #1732, opened 2022-01-21
Reuse rendercomposable in tests

Schahen:
The main goal of this PR is to reuse "regular" renderComposable  in composition helper function from tests.  I left renderTestComposable (since technically it's the part of public API) but let's discuss whether we need it.

### DIFF
--- a/web/test-utils/build.gradle.kts
+++ b/web/test-utils/build.gradle.kts
@@ -30,6 +30,7 @@ kotlin {
         val jsMain by getting {
             dependencies {
                 implementation(project(":internal-web-core-runtime"))
+                implementation(project(":web-core"))
                 implementation(kotlin("stdlib-js"))
             }
         }
--- a/web/core/src/jsTest/kotlin/DomSideEffectTests.kt
+++ b/web/core/src/jsTest/kotlin/DomSideEffectTests.kt
@@ -2,7 +2,6 @@ package org.jetbrains.compose.web.core.tests
 
 import androidx.compose.runtime.*
 import org.jetbrains.compose.web.dom.Div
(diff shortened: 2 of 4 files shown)


web/test-utils/build.gradle.kts
```diff
@@ -30,6 +30,7 @@ kotlin {
         val jsMain by getting {
             dependencies {
                 implementation(project(":internal-web-core-runtime"))
+                implementation(project(":web-core"))
```
Schahen, line 33: This is the only part I have concerns about. Technically we need this dependency because of a single (crucial one but nevertheless) function.

olonho: What is this function?

eymar:
1) What's the benefit of using 

> "regular" renderComposable

in tests?

Is the code duplication (in renderTestComposable) a concern?

2) All public functions are `@ComposeWebExperimentalTestsApi`, so I think we can make changes if needed even to existing methods.

Schahen:
@eymar it's more not about benefits but rather the code duplication - otherwise at some point in time we can forget about it and change renderTestComposable so that it won't be equivalent to "regular" one. Of course, we have integration tests as well, but after all why to have two entities when without effort we can have one?

web/test-utils/build.gradle.kts
Schahen: @olonho it's about renderComposable - it's an important function though - otherwise we were testing in unit tests equivalent but separate construct.

eymar:
web:core also depends on "implementation(project(":test-utils"))"

does it work like that when you have a circular dependency?

Schahen:
@eymar it does because it depends on it as on a test dependency

Issue #1493, opened 2021-11-28
web-compose-bird example. Impossible to win the game


igordmn:
Compose 1.0.0-rc5

There are too many obstacles. We need to reduce their count.

https://user-images.githubusercontent.com/5963351/143785099-b7047b6c-dfac-4c64-8cf8-d9bef9841a8c.mp4

igordmn:
Fixed

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-5137">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1747, opened 2022-01-25
Fix web-compose-bird example too many obstacles(#1493)

xuankaicat:


akurasov:
It is just refactoring and doesn't fix anything, right?

xuankaicat:
> It is just refactoring and doesn't fix anything, right?

it will set the redio button checked to false in order to fix the bug in #1493

olonho (approved):

PR #1752, opened 2022-01-25
Treat warning as errors globally in web

Schahen:
The only exception is web-widgets, which are all deprecated but suppressing this deprecation via annotation won't do a good service to users (and there's no way in modern Kotlin tooling to ignore this warning only).

The other change  useExperimentalAnnotation in build (which is going to be deprecated) with recommended optIn option

### DIFF
--- a/web/compose-compiler-integration/src/jsMain/kotlin/CrossmoduleTestsDependencies.kt
+++ b/web/compose-compiler-integration/src/jsMain/kotlin/CrossmoduleTestsDependencies.kt
@@ -37,7 +37,7 @@ class ClassSavesComposableIntoVar(c: @Composable () -> Unit) {
 }
 
 class ClassSavesComposableIntoLateinitVar(c: @Composable () -> Unit) {
-    lateinit var composableVar: @Composable () -> Unit
+    var composableVar: @Composable () -> Unit
 
     init {
         composableVar = c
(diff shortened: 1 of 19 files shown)


web/compose-compiler-integration/src/jsMain/kotlin/CrossmoduleTestsDependencies.kt
```diff
@@ -37,7 +37,7 @@ class ClassSavesComposableIntoVar(c: @Composable () -> Unit) {
 }
 
 class ClassSavesComposableIntoLateinitVar(c: @Composable () -> Unit) {
-    lateinit var composableVar: @Composable () -> Unit
```
eymar, line 40: maybe it's better to suppress the warning here, because this a test and it needs to test lateinit var with Composable

Schahen: just for self-education - why it need to be lateinit? I mean, I am a simple man, I see tests passing, I'm joyous.

eymar: it's a part of a test case.  Some time ago we had such vars broken and this test had been failing. 
Now it's fixed. So by keeping `lateinit` we ensure we catch it if it breaks again.

eymar:
since https://github.com/JetBrains/compose-jb/pull/1803 was merged,  can this pr be closed?

PR #1753, opened 2022-01-25
Add `checkAndWarnAboutUsingJsPlugin` to gradle plugin

### LINKED ISSUE #1693: Publication of not multiplatform project
It is required to add some warning or even error when user trying to publish compose library which is configured as (for example) js project.

eymar:
Tried to use `project.plugins.whenObjectAdded {  }`, but it receives only plugins added after our plugin, so order matters.
At the same time `project.plugins.all { }` receives all plugins (added before and after our one) 

closes #1693

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
@@ -25,9 +25,11 @@ import org.jetbrains.compose.desktop.application.internal.currentTarget
 import org.jetbrains.compose.desktop.preview.internal.initializePreview
 import org.jetbrains.compose.experimental.dsl.ExperimentalExtension
 import org.jetbrains.compose.experimental.internal.configureExperimental
+import org.jetbrains.compose.internal.COMPOSE_PLUGIN_ID
+import org.jetbrains.compose.internal.KOTLIN_JS_PLUGIN_ID
+import org.jetbrains.compose.internal.KOTLIN_MPP_PLUGIN_ID
 import org.jetbrains.compose.web.WebExtension
 import org.jetbrains.kotlin.gradle.plugin.KotlinDependencyHandler
-import org.jetbrains.kotlin.gradle.plugin.KotlinJsPluginWrapper
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 internal val composeVersion get() = ComposeBuildConfig.composeVersion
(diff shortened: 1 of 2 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -134,6 +135,21 @@ class ComposePlugin : Plugin<Project> {
...
+        checkAndWarnAboutUsingJsPlugin(project)
+    }
+
+    private fun checkAndWarnAboutUsingJsPlugin(project: Project) {
+        val msg = "Project has kotlin(\"js\") plugin applied.\n" +
+                "It's necessary to apply kotlin(\"multiplatform\") plugin " +
+                "in order to use \"org.jetbrains.compose\" plugin with k/js targets.\n" +
+                "See https://github.com/JetBrains/compose-jb/blob/master/tutorials/Web/README.md"
+
+        project.plugins.all {
```
AlexeyTsvetkov, line 148: Usually `plugins.withId` is used for that case: 
```
plugins.withId("org.jetbrains.kotlin.js") {
    // ...
}
```

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -134,6 +135,21 @@ class ComposePlugin : Plugin<Project> {
                 }
             }
         }
+
+        checkAndWarnAboutUsingJsPlugin(project)
+    }
+
+    private fun checkAndWarnAboutUsingJsPlugin(project: Project) {
+        val msg = "Project has kotlin(\"js\") plugin applied.\n" +
```
AlexeyTsvetkov, line 143: I think `kotlin(...)` shortcuts are Kotlin DSL only.
`org.jetbrains.kotlin.js` and `org.jetbrains.kotlin.multiplatform` should be used instead.

Also consider using constants instead of literals https://github.com/JetBrains/compose-jb/blob/master/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/constants.kt#L8

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -134,6 +135,21 @@ class ComposePlugin : Plugin<Project> {
...
+                "It's necessary to apply kotlin(\"multiplatform\") plugin " +
```
AlexeyTsvetkov, line 144: The warning is quite long.
What do you think about `'org.jetbrains.compose' plugin is not compatible with 'org.jetbrains.kotlin.js' plugin (use 'org.jetbrains.kotlin.multiplatform' instead)`?

AlexeyTsvetkov (changes requested):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
eymar: thanks. fixed

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
eymar: fixed

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
eymar: fixed

AlexeyTsvetkov (approved):

olonho (approved):

PR #1757, opened 2022-01-26
Added link to Gradle tool window﻿

AnimeAllstar:
As someone new to Compose, I did not know how I'm supposed to "Open `build.gradle.kts` as a project". It'd be helpful to add a link since this is a tutorial.

### DIFF
--- a/tutorials/Getting_Started/README.md
+++ b/tutorials/Getting_Started/README.md
@@ -144,7 +144,7 @@ fun main() = application {
 ```
 ## Running the project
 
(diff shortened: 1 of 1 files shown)


akurasov:
I think this link only confuses, but not clarifies how to do it. It is about gradle support in IDEA, not about opening file as a project.

olonho (approved):

olonho:
Think it is useful, as it shows where to look for Gradle.

PR #1770, opened 2022-01-31
App API prototype

olonho:


components/Kapp/library/src/commonMain/kotlin/org/jetbrains/compose/kapp/Kapp.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import androidx.compose.runtime.Composable
+
+interface KAppScope
+
+@Composable
+expect fun KAppScope.Frame(content: @Composable () -> Unit)
+
+// Complex multi-frame application
+fun kapp(name: String = "application", title: String = "Compose Application", content: @Composable KAppScope.() -> Unit) {
```
VeselovAlex, line 11: Does name mean application name, and title mean window title? In this case I think it is reasonable to:
1. Remove default argument for name, let's make it mandatory
2. Make title a property in some WindowScope, with initial value set to application name
3. simpleKapp callback should be in WindowScope, and app callback in some window management (router?) scope

Issue #1528, opened 2021-12-01
Web: CheckboxInput cannot be used with MDC Checkbox component


OliverO2:
Using the controlled `CheckboxInput` to create a wrapper for an [MDC Checkbox](https://github.com/material-components/material-components-web/tree/v13.0.0/packages/mdc-checkbox) produces this exception: 

```
Uncaught Error: Checkbox component requires a .mdc-checkbox__native-control element
MDCCheckbox.getNativeControl                    	component.js:233
hasNativeControl                                	component.js:136
MDCCheckboxFoundation.transitionCheckState      	foundation.js:121
MDCCheckboxFoundation.handleChange              	foundation.js:118
set                                             	component.js:186
_no_name_provided__184.invoke_8ohztc_k$         	sandbox.js:64912
eval                                            	sandbox.js:64941
updateProperties                                	sandbox.js:68539
_no_name_provided__190.invoke_7q0vyf_k$         	sandbox.js:68648
l                                               	sandbox.js:68802
_no_name_provided__85.invoke_472icx_k$          	sandbox.js:40578
eval                                            	sandbox.js:42928
CompositionImpl.applyChanges_sv8swh_k$          	sandbox.js:44082
_no_name_provided__117.invoke_wiekkq_k$         	sandbox.js:45878
eval                                            	sandbox.js:48457
_no_name_provided__155.invoke_okcceq_k$         	sandbox.js:63086
eval                                            	sandbox.js:63194
Async call from requestAnimationFrame
$withFrameNanosCOROUTINE$4.doResume_0_k$        	sandbox.js:63125
[...]
```

This is caused by the following code invoking `checked(checked)` before applying attributes:
https://github.com/JetBrains/compose-jb/blob/db53b95a3174f985f067ef0620dd0ca8369ecb95/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/InputElements.kt#L25-L33

MDC Checkbox requires a class "mdc-checkbox__native-control" to be present before dealing with state changes. Reversing the order to `this.attrs()`, then `checked(checked)` would fix it:

```kotlin
        attrs = {
            this.attrs()
            checked(checked)
        }
```

eymar:
Hi @OliverO2 !
Do you have a branch somewhere where we can reproduce this?

if the order really matters, then you can try:

```kotlin
Input(
    type = InputType.CheckBox,
    attrs = {
         yourAttrs()
         checked(checked)
    }
)
```

[...]

PR #1788, opened 2022-02-03
web: apply `classes` and `style` before all other attributes and properties

eymar:
This mitigates "Flash of unstyled content"

The order is following:

1) Everything passed into `classes(...)` applied before everything else
2) Everything passed into `style { ... }` applied before everything else 
3) apply all `attr(...)` calls (including all custom attrs (like `id`, `title`, etc) which use `attr()` under the hood). 
    So `attr("class", "someClass")` can completely override calls to `classes(...)`
    And `attr("style", "color: red;")` can completely override everything set in `style { }` blocks
4) apply all `prop(...)` calls

Please see new tests which try to check the order. 

Fixes https://github.com/JetBrains/compose-jb/issues/1528

### DIFF
--- a/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
+++ b/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
@@ -58,7 +58,7 @@ interface AttrsScope<TElement : Element>: EventsListenerScope {
      *  In the ideology of Composable functions and their recomposition one just don't need to remove classes,
      *  since if your classList is, for instance, condition-dependent, you can always just call this method conditionally.
      */
-    fun classes(vararg classes: String)= prop(setClassList, classes)
+    fun classes(vararg classes: String)
 
     fun id(value: String) = attr(ID, value)
     fun hidden() = attr(HIDDEN, true.toString())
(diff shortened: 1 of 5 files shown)


web/svg/src/jsTest/kotlin/svg/SvgTests.kt
```diff
@@ -132,16 +134,19 @@ class SvgTests {
...
-                })
+                }
             }
         }
 
-        assertEquals(
-            "<svg><text x=\"20\" y=\"30\" class=\"small\">some text</text></svg>",
-            nextChild<SVGCircleElement>().outerHTML
-        )
+        with(nextChild<SVGElement>().firstChild!! as SVGTextElement) {
```
Schahen, line 143: Why can not we still check agains the whole string?

web/core/src/jsTest/kotlin/elements/AttributesTests.kt
```diff
@@ -387,18 +434,31 @@ class AttributesTests {
             }
         }
 
-        assertEquals(
-            expected = "<button class=\"a\" style=\"color: red;\">Button</button>",
-            actual = nextChild().outerHTML
-        )
+        with(nextChild()) {
```
Schahen, line 437: the same question as in SVG test - I can not see why this is better (I apologise if I miss something)

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Base.kt
```diff
@@ -81,14 +79,25 @@ private class DomElementWrapper(override val node: Element): DomNodeWrapper(node
 
     fun updateAttrs(attrs: Map<String, String>) {
         node.getAttributeNames().forEach { name ->
-            if (name == "style") return@forEach
-            node.removeAttribute(name)
+            when (name) {
+                "style", "class" -> {
```
Schahen, line 83: What will happen if we have, say, uppercase STYLE or CLASS - for some reason? 
I don't remember that it's valid to be honest, just asking.

Schahen (commented):
I suggest to discuss it later yet another time but as of now nothing more to add from my side.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
```diff
@@ -144,6 +144,18 @@ open class AttrsScopeBuilder<TElement : Element> : AttrsScope<TElement>, EventsL
     override val propertyUpdates = mutableListOf<Pair<(Element, Any) -> Unit, Any>>()
     override var refEffect: (DisposableEffectScope.(TElement) -> DisposableEffectResult)? = null
 
+    internal var classes: MutableList<String> = arrayListOf()
```
Schahen, line 147: let's make this val

Schahen (approved):

PR #1789, opened 2022-02-03
Move renderComposable to internal-web-core-runtime

Schahen:
This way we won't have pseudo-circular  project dependency which causes us troubles and won't allow to build in stricter warnig-as-errors mode (see https://youtrack.jetbrains.com/issue/KT-51110)


web/test-utils/src/jsMain/kotlin/org/jetbrains/compose/web/testutils/TestUtils.kt
```diff
@@ -32,13 +21,18 @@ import kotlin.time.DurationUnit
 import kotlin.time.ExperimentalTime
 import kotlin.time.toDuration
 
+interface TestComposition {
```
eymar, line 24: I don't see any problem with adding this interface. Just want to understand its purpose.
What problem does it solve?

web/internal-web-core-runtime/src/jsMain/kotlin/org/jetbrains/compose/web/renderComposable.kt
```diff
@@ -1,21 +1,31 @@
...
+import org.jetbrains.compose.web.dom.DOMScope
+import org.jetbrains.compose.web.internal.runtime.ComposeWebInternalApi
+import org.jetbrains.compose.web.internal.runtime.DomApplier
+import org.jetbrains.compose.web.internal.runtime.DomNodeWrapper
+import org.jetbrains.compose.web.internal.runtime.GlobalSnapshotManager
+import org.jetbrains.compose.web.internal.runtime.JsMicrotasksDispatcher
 import org.w3c.dom.Element
 import org.w3c.dom.HTMLBodyElement
 import org.w3c.dom.get
+
```
eymar, line 28: before this PR "internal-web-core-runtime" had only declarations which were not supposed to be used by external users and they all were marked with `@ComposeWebInternalApi`.

`renderComposable ` is a public function and it's used by external users. 
Does it belong to this module?

After giving it some thought:
It looks like this function is quite independent from the rest of composable-dom-API. And it's even possible to use only this module + a different implementation of html composables. 

The only thing bothering me is the name of this module. It's not internal anymore, as it contains the API intended for everyone's use. Should we rename this module?

Another question:
Is it ensured that when users add `web-core` dependency, all the APIs from "internal-web-core-runtime" will be available without adding an explicit dependency?

eymar: should "renderComposable" have its tests in this module now?

eymar (approved):

PR #1799, opened 2022-02-06
Fix jnilib files not signed

Thomas-Vos:
You can see here that only `dylib` files are signed:

https://github.com/openjdk/jdk/blob/f7814c120bf84d7e9b459f81a6ce19b44fa122ec/src/jdk.jpackage/macosx/classes/jdk/jpackage/internal/MacAppImageBuilder.java#L673-L677

So that means that `jnilib` files still need to be signed manually.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/files/MacJarSignFileCopyingProcessor.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/files/MacJarSignFileCopyingProcessor.kt
@@ -31,9 +31,22 @@ internal class MacJarSignFileCopyingProcessor(
                      * so we need to remove signature before running jpackage.
                      *
                      * JDK 18 processes signed libraries fine, so we don't have to do anything.
+                     *
+                     * Note that the JDK only signs dylib files and not jnilib files,
+                     * so jnilib files still need to be signed here.
                      */
-                    jvmRuntimeVersion == 17 -> signer.unsign(target)
-                    else -> {}
+                    jvmRuntimeVersion == 17 -> {
+                        if (source.name.endsWith(".jnilib")) {
+                            signer.sign(target)
+                        } else {
+                            signer.unsign(target)
+                        }
+                    }
+                    else -> {
+                        if (source.name.endsWith(".jnilib")) {
+                            signer.sign(target)
+                        }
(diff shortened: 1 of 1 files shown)


Thomas-Vos:
Without the PR you get the following:
Screenshot 2022-02-08 at 00 17 19

These libs were added using `fromFiles` in Gradle compose configuration.

AlexeyTsvetkov:
I don't think the fix works ATM, because `jnilib` extension is checked, after a file is determined to have a `dylib` extension:
```
if (source.name.isDylibPath) {
    // ...
            if (source.name.endsWith(".jnilib")) {
                signer.sign(target)
            } else {
                signer.unsign(target)
            }
        }
        else -> {
            if (source.name.endsWith(".jnilib")) {
                signer.sign(target)
            }
        }
    // ...
}
```

Overall, the idea looks good to me, but I would prefer something like:
```
if (source.name.isDylibPath) {
     // ... same as before
} else if (source.name.isJnilibPath) {
    signer.sign(target)
}
```

AlexeyTsvetkov (changes requested):
I don't think the fix works ATM, because `jnilib` extension is checked, after a file is determined to have a `dylib` extension:
```
if (source.name.isDylibPath) {
    // ...
            if (source.name.endsWith(".jnilib")) {
                signer.sign(target)
            } else {
                signer.unsign(target)
            }
        }
        else -> {
            if (source.name.endsWith(".jnilib")) {
                signer.sign(target)
            }
        }
    // ...
}
```

Overall, the idea looks good to me, but I would prefer something like:
```
if (source.name.isDylibPath) {
     // ... same as before
} else if (source.name.isJnilibPath) {
    signer.sign(target)
}
```

Thomas-Vos:
@AlexeyTsvetkov actually the fix does work, because the `isDylibPath` function (which already existed), checks for both extensions. See:

https://github.com/JetBrains/compose-jb/blob/a1c4c8a41ee37c2f56ced2598ffb124decc28c6d/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/files/MacJarSignFileCopyingProcessor.kt#L76-L77

However, I can change it because I agree that it is not that clear.

AlexeyTsvetkov:
@Thomas-Vos whoops, my bad :) LGTM

AlexeyTsvetkov (approved):

PR #1807, opened 2022-02-08
Improve reporting of preview errors

AlexeyTsvetkov:


gradle-plugins/preview-rpc/src/main/kotlin/org/jetbrains/compose/desktop/ui/tooling/preview/rpc/commands.kt
```diff
@@ -20,7 +20,9 @@ internal fun RemoteConnection.receiveAttach(
         if (type == Command.Type.ATTACH) {
             val version = args.firstOrNull()?.toIntOrNull() ?: 0
             if (PROTOCOL_VERSION != version) {
-                listener?.onIncompatibleProtocolVersions(PROTOCOL_VERSION, version)
+                listener?.onError(
```
olonho, line 23: Shall/can we report the versions in question?

idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/ui/PreviewPanel.kt
```diff
@@ -0,0 +1,89 @@
...
+import java.awt.image.BufferedImage
+import java.util.concurrent.atomic.AtomicReference
+import javax.swing.SwingUtilities
+
+internal class PreviewPanel(private val myProject: Project) : JBPanel<PreviewPanel>() {
+    sealed class PreviewPanelState {
+        data class Image(val image: BufferedImage, val dimension: Dimension) : PreviewPanelState()
+        class Error(val error: String) : PreviewPanelState()
+    }
+    private val myState = AtomicReference<PreviewPanelState>()
```
olonho, line 25: Maybe `myState` -> `previewPanelState`?

idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/ui/PreviewPanel.kt
```diff
@@ -0,0 +1,89 @@
...
+    private val myStatusText = object : StatusText(this) {
+        override fun isStatusVisible(): Boolean {
+            return myState.get() is PreviewPanelState.Error
+        }
+    }
+
+    init {
+        SwingUtilities.invokeLater {
```
olonho, line 33: Shall we do that on some particular event AWT init event instead to make sure everything is properly inited?

olonho (approved):
Overall looks good, few nits. Also do we have tests for that?

idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/ui/PreviewPanel.kt
AlexeyTsvetkov: I used Intellij `my*` convention + short name, which in my opinion makes sense for a private property.
I think it is obvious from a type and the context, what state is meant.
If it would become public, we can refactor later.

gradle-plugins/preview-rpc/src/main/kotlin/org/jetbrains/compose/desktop/ui/tooling/preview/rpc/commands.kt
AlexeyTsvetkov: Sure, but it's a little bit tricky at the moment, because right now a protocol version is received from a preview host, not from the Gradle plugin. So we need to pass the Gradle plugin version to the IDE, which is itself incompatible change of protocol. I need to think how to implement that change better, so lets postpone

idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/ui/PreviewPanel.kt
AlexeyTsvetkov: Honestly I have no idea :)
But it seems that `createToolWindowContent` in which we create PreviewPanel is called after Intellij is fully initialized, so I would assume it is not necessary.

PR #1808, opened 2022-02-08
web: make ComposeDomNode inline

eymar:


web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Base.kt
```diff
@@ -134,7 +136,9 @@ fun <TElement : Element> TagElement(
             }
         },
         elementScope = scope,
-        content = content
+        content = {
```
Schahen, line 139: won't this create unecessary lambdas?

Schahen (commented):
Apart from a single question I had I have no other questions. Let's discuss that particular question however.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Base.kt
eymar: no, this lambda will be inlined.

PR #1814, opened 2022-02-09
web: make `applyAttrs` in the second overloaded `TagElement` nullable to be consistent with the first one

ShreckYe:
From what I understand, for non-inline functions passing `null` is a bit more efficient than passing an empty lambda, and that's why most functions in this library have the signature `attrs: AttrBuilderContext<HTMLElement>? = null`. Then there is no reason the second overloaded `TagElement` function shouldn't be like that, so I guess this is a small mistake that should be fixed.


hfhbd:
I accidentally created the same PR #1822

eymar:
thanks! lgtm

PR #1843, opened 2022-02-14
Expose jpackage's --resource-dir in Gradle DSL 

### LINKED ISSUE #1766: [linux] Scripts (preinst, postinst) inserted on installation resources don't execute
I'm trying to run script (postinst) after installing my application.
I'm able to put the scripts in the resources folder of the installation directory, but after the complete installation, they are not executed.

Follow the script instructions from this link:
http://www.embeddedlinux.org.cn/OEManual/recipes_installation_scripts.html

When I create the installer through jpackage via terminal, the scripts are executed normally, however, when I create the installer through the compose plugin, they are not executed.

I created a repository to simulate this situation, follow the link:
https://github.com/rodrigoknob/test_compose

If you run the create_package.sh script inside the jpackage_manual folder, it will create the .deb installer, and when you run the installer, the postinst scripts will run normally.

Now, if you are going to create the installer from the ./gradlew package task, the installer will be generated, however, if it is executed, no script is executed.

I don't know if some configuration is missing, or the plugin is not executing the scripts. 
Then, how to run the .deb package scripts via installer created by the plugin?

AltNico:
jpackage provides the option `--resource-dir` that allows to override
> resources used by jpackage, such as background images and template files for properties and scripts.

https://docs.oracle.com/en/java/javase/17/jpackage/override-jpackage-resources.html

This includes files like `preinst` and `postinst` that are used by .deb files to allow modifications of the installation process.

When looking at the .deb file generated by jpackage, the default `preinst` and `postinst` scripts land in the _control.tar.xz_ archive of the .deb file.

However, with the .deb files produced by the `packageDeb` task, one can see that the files from the source's `appResourcesRootDir` (specified in `build.gradle.kts`) land in _/opt/briar-desktop/lib/app/resources/_ inside the _data.tar.xz_
archive.

Looking at the jpackage parameters used during `packageDeb`, one can see that it uses _compose/tmp/resources_:
```
--resource-dir
"/home/dev/Documents/briar-desktop/build/compose/tmp/resources"
```
[project root]/build/compose/tmp/packageDeb.args.txt

Looking at that directory, it is empty. However, inside _compose/tmp/main/resources_ the files provided in source's `appResourcesRootDir` are found:
```
$ ls -l compose/tmp/main/resources/
total 8
-rwxr-xr-x 1 dev dev   9 Feb 14 13:02 postinst
-rwxr-xr-x 1 dev dev 618 Feb 14 13:02 preinst
$ ls -l compose/tmp/resources/
total 0
```

This MR therefore exposes `jpackageResources` to be specified in Gradle build files  by
Compose projects.

https://github.com/JetBrains/compose-jb/blob/v1.0.1/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt#L248

Fixes #1766.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/NativeDistributions.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/NativeDistributions.kt
@@ -27,6 +27,7 @@ open class NativeDistributions @Inject constructor(
     var vendor: String? = null
     var packageVersion: String? = null
     val appResourcesRootDir: DirectoryProperty = objects.directoryProperty()
+    val packagingResourcesRootDir: DirectoryProperty = objects.directoryProperty()
     val licenseFile: RegularFileProperty = objects.fileProperty()
 
     val outputBaseDir: DirectoryProperty = objects.directoryProperty().apply {
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureApplication.kt
@@ -94,6 +94,20 @@ internal fun Project.configurePackagingTasks(apps: Collection<Application>) {
             into(destDir)
         }
 
+        val preparePackagingResources = tasks.composeTask<Sync>(
+            taskName("preparePackagingResources", app)
+        ) {
+            val packagingResourcesRootDir = app.nativeDistributions.packagingResourcesRootDir
+            if (packagingResourcesRootDir.isPresent) {
+                from(packagingResourcesRootDir.dir("common"))
+                from(packagingResourcesRootDir.dir(currentOS.id))
+                from(packagingResourcesRootDir.dir(currentTarget.id))
+            }
+
+            val destDir = project.layout.buildDirectory.dir("compose/tmp/resources")
+            into(destDir)
+        }
+
         val createRuntimeImage = tasks.composeTask<AbstractJLinkTask>(
             taskName("createRuntimeImage", app)
         ) {
@@ -113,6 +127,7 @@ internal fun Project.configurePackagingTasks(apps: Collection<Application>) {
                 app,
                 createRuntimeImage = createRuntimeImage,
                 prepareAppResources = prepareAppResources,
+                preparePackagingResources = preparePackagingResources,
                 checkRuntime = checkRuntime
             )
         }
@@ -133,6 +148,7 @@ internal fun Project.configurePackagingTasks(apps: Collection<Application>) {
                         app,
                         createRuntimeImage = createRuntimeImage,
                         prepareAppResources = prepareAppResources,
+                        preparePackagingResources = preparePackagingResources,
                         checkRuntime = checkRuntime
                     )
                 } else {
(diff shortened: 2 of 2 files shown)


AlexeyTsvetkov:
@AltNico I'm not sure I like the idea of mixing packaging resources and app resources. Just to be clear, I'm not strongly against it either. That's why there are two `AbstractJPackageTask` properties: `jpackageResources` and `appResourcesDir`. 
However, `jpackageResources` is not exposed in the DSL. The reason for that was that at the time I was not sure, if users actually need it, and I try to avoid adding public APIs unless I understand use cases.
Now it seems that definitely there are users, who need it, so instead of merging `jpackageResources` and `appResourcesDir`, we could just expose `jpackageResources` in the DSL similarly to `appResourcesDir` (e.g. via `packagingResourcesRootDir` or `jpackageResourcesRootDir` property). @AltNico what do you think about this alternative option?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/NativeDistributions.kt
```diff
@@ -27,6 +27,7 @@ open class NativeDistributions @Inject constructor(
     var vendor: String? = null
     var packageVersion: String? = null
     val appResourcesRootDir: DirectoryProperty = objects.directoryProperty()
+    val packagingResourcesRootDir: DirectoryProperty = objects.directoryProperty()
```
AltNico, line 30: I'm now adding `packagingResourcesRootDir` to the DSL similar to how `appResourcesRootDir` is implemented.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureApplication.kt
```diff
@@ -211,6 +228,12 @@ internal fun AbstractJPackageTask.configurePackagingTask(
         appResourcesDir.set(resourcesDir)
     }
 
+    preparePackagingResources?.let { prepareResources ->
+        dependsOn(prepareResources)
+        val resourcesDir = project.layout.dir(prepareResources.map { it.destinationDir })
+        jpackageResources.set(resourcesDir)
+    }
+
```
AltNico, line 236: Here I'm not sure whether `resourcesDir` contains the right path.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureApplication.kt
```diff
@@ -94,6 +94,20 @@ internal fun Project.configurePackagingTasks(apps: Collection<Application>) {
...
+            if (packagingResourcesRootDir.isPresent) {
+                from(packagingResourcesRootDir.dir("common"))
+                from(packagingResourcesRootDir.dir(currentOS.id))
+                from(packagingResourcesRootDir.dir(currentTarget.id))
+            }
+
+            val destDir = project.layout.buildDirectory.dir("compose/tmp/resources")
+            into(destDir)
+        }
+
```
AltNico, line 110: This allows to specify different files per OS/arch combination, just like with `appResourcesRootDir`.

AltNico (commented):
@AlexeyTsvetkov Exposing `jpackageResources` via the DSL sounds good to me! I gave it a try and implemented `packagingResourcesRootDir` analogously to how `appResourcesDir` is implemented. I'm adding some comments in the code, but just want to mention that I did not add the `preparePackagingResources` functions to the `configurePackagingTask()` calls of `run composeTask` and `JavaExec.configureRunTask` because there we don't need those packaging resources, I guess.

Still unsure on how to test it (with Briar Desktop), though.

Thomas-Vos:
Will this line not delete the provided directory?

https://github.com/JetBrains/compose-jb/blob/4fb11bc52fb48b6b861dbe407cc6689b34743a26/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt#L470

krizzu:
Hey @AlexeyTsvetkov I know it's been a while, but what you think about the PR? I think it'd be super useful to get in

PR #1886, opened 2022-02-22
Some CSS cleanup fixes that will make further relocation of css files an easier task

Schahen:


web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheet.kt
```diff
@@ -79,49 +78,32 @@ open class StyleSheet(
...
-        }
-    }
-
     protected class CSSHolder(private val usePrefix: Boolean, private val cssBuilder: CSSBuilder.() -> Unit) {
         operator fun provideDelegate(
             sheet: StyleSheet,
             property: KProperty<*>
         ): ReadOnlyProperty<Any?, String> {
             val sheetName = if (usePrefix) "${sheet::class.simpleName}-" else ""
-            val selector = CSSSelector.CSSClass("$sheetName${property.name}")
```
eymar, line 117: was it incorrect? why can't we use `CSSSelector.CSSClass` here?

eymar: I see now that it was removed. Can method `className ` be used then?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheetBuilder.kt
```diff
@@ -241,66 +242,54 @@ interface SelectorsScope {
...
-}
-
-private data class Raw(val selector: String) : CSSSelector() {
+private data class RawSelector(val selector: String) : CSSSelector() {
     override fun toString(): String = selector
 }
 
 private data class Combine(val selectors: MutableList<CSSSelector>) : CSSSelector() {
-    override fun contains(other: CSSSelector, strict: Boolean): Boolean =
-        contains(this, other, selectors, strict)
```
eymar, line 262: just to be sure: was `strict` unused at all?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/selectors/CSSSelectors.kt
```diff
@@ -35,31 +35,21 @@ sealed interface Nth {
...
 
 abstract class CSSSelector internal constructor() {
-    override fun equals(other: Any?): Boolean {
-        return this === other || asString() == (other as? CSSSelector)?.asString()
-    }
 
-    internal open fun contains(other: CSSSelector, strict: Boolean = false): Boolean {
-        return if (strict) this === other else this == other
+    internal open fun contains(other: CSSSelector): Boolean {
+        return this === other
```
eymar, line 40: Just to understand the motivation to remove `strict`.

We never passed `true` for it? right? so it's a redundant check in 100% cases?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheet.kt
Schahen: className is unfortunately a context method but actually I see your point - let me think what can be done.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheetBuilder.kt
Schahen: @eymar correct, it was never used and was true in 100% of cases (actually there's a single point where it was used) - and keeping something just in case is a wrong thing to do (it also was quite confusing)

eymar (approved):

PR #1973, opened 2022-03-22
Provide relative path to jpackage --main-jar. Fixes #1578

### LINKED ISSUE #1578: [desktop] package.cfg contains unwanted classpath
### **Notice: This happened in every Compose desktop app, and it shouldn't exist in any release version.**

After generating a native distribution, it will create a [packageName].cfg in `build\compose\binaries\main\app\[packageName]\app`, which contains a source code real path.
For example, for `compose-jb\examples\todoapp\desktop`, In my PC, the project is in `E:\Git\JetBrains\compose-jb`.
It will generate `compose-jb\examples\todoapp\desktop\build\compose\binaries\main\app\ComposeDesktopTodo\app\ComposeDesktopTodo.cfg`
image

### We can also see @theapache64 's `stackzy`

After install [Stackzy-1.1.1.msi](https://github.com/theapache64/stackzy/releases/download/1.1.1/Stackzy-1.1.1.msi)

image

### Or my project [genshin-imapct-server-switcher.zip](https://github.com/Omico/genshin-impact-server-switcher/releases/download/v1.0.0/genshin-imapct-server-switcher.zip)

image

mcpiroman:
Per jpackage doc:

> --main-jar main jar file
> The main JAR of the application; containing the main class ___(specified as a path relative to the input path)___.
> 
> Either --module or --main-jar option can be specified but not both.

Tested only on Windows.

Fixes #1578

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
@@ -287,7 +287,8 @@ abstract class AbstractJPackageTask @Inject constructor(
 
             val mappedJar = libsMapping[launcherMainJar.ioFile]?.singleOrNull()
                 ?: error("Main jar was not processed correctly: ${launcherMainJar.ioFile}")
-            cliArg("--main-jar", mappedJar)
+            val mainJarRelative = (if (currentTarget.os == OS.Windows) "\\" else "/") + mappedJar.relativeTo(libsDir.ioFile).toString()
+            cliArg("--main-jar", mainJarRelative)
             cliArg("--main-class", launcherMainClass)
 
             when (currentOS) {


Omico:
@mcpiroman This works for me. Thanks!

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
```diff
@@ -287,7 +287,8 @@ abstract class AbstractJPackageTask @Inject constructor(
 
             val mappedJar = libsMapping[launcherMainJar.ioFile]?.singleOrNull()
                 ?: error("Main jar was not processed correctly: ${launcherMainJar.ioFile}")
-            cliArg("--main-jar", mappedJar)
+            val mainJarRelative = (if (currentTarget.os == OS.Windows) "\\" else "/") + mappedJar.relativeTo(libsDir.ioFile).toString()
```
jasonsparc, line 290: I believe that the leading slash (`"\\"` or `"/"`) is unnecessary and in fact, on Windows, that leading slash is causing `jpackage` to fail. See screenshot:

Capture

Notice in the above screenshot that the first command failed, but the second one succeeded (since the leading slash was omitted).

jasonsparc:
@AlexeyTsvetkov @igordmn I just wanted to know, when will this PR be approved? I know compose multiplatform is still in alpha, but it would really be nice if this PR gets approved. Currently, the main issue is that it's leaking sensitive information (i.e., the build machine's path) when people look into the `app/*.cfg` file (which is present even through the generated installer).

It's harmless at runtime though, given that the actual main `jar` is already fed via `--input`, so I wish we could just omit the `--main-jar` option, but sadly, `jpackage` requires it.

Putting your project in a directory with a non-sensitive information is a workaround, but I wish we could just build our projects anywhere without worrying about this. Also, very likely, a lot of people aren't even aware of this issue.

AlexeyTsvetkov:
Merged the PR manually with some additional changes on top

Issue #1975, opened 2022-03-24
Splitpanel should not depend on Material


rock3r:
The SplitPanel library depends on the Material Compose library, but only to get a default colour. This makes it impossible to use the component in a project that doesn't include Material.

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-1975">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #1979, opened 2022-03-25
SplitPane. Remove material dependency

igordmn:
SplitPane by design shouldn't have default visual representation, because it mostly depends on the design system (material, material3, etc)

Fixes https://github.com/JetBrains/compose-jb/issues/1975

RelNote:
SplitPane no longer depends on compose.material. Now, by default, SplitPane has invisible 0px splitter, not 1px black/white splitter as it was before. If you want to preserve the previous behaviour, specify a splitter explicitly:
```
HorizontalSplitPane(splitterState) {
    splitter {
        visiblePart {
            Box(
                Modifier
                    .height(1.dp)
                    .fillMaxHeight()
                    .background(MaterialTheme.colors.background)
            )
        }
    }
}
```

### DIFF
--- a/components/SplitPane/library/build.gradle.kts
+++ b/components/SplitPane/library/build.gradle.kts
@@ -15,7 +15,6 @@ kotlin {
             dependencies {
                 api(compose.runtime)
                 api(compose.foundation)
-                api(compose.material)
             }
         }
         named("desktopMain") {}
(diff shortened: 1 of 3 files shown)


dima-avdeev-jb (approved):

rock3r:
Ace, thanks ✨

PR #1982, opened 2022-03-29
Fix parameter names of StyleScope.borderWidth functions

EdwarDDay:
Reference: https://developer.mozilla.org/en-US/docs/Web/CSS/border-width#syntax

This will break code, if it used the wrong parameter names. Does this need to go through a deprecation cycle?


EdwarDDay:
I could only think of deprecate the old functions and have new functions with other names. But the current names are already so good, that this would decrease usability in my opinion.

EdwarDDay:
@eymar any update here? Is there anything I can improve in the PR?

eymar:
Hi @EdwarDDay ! I apologise for a long waiting time. 
We tried to tackle the issue with wrong parameter names and solve without breaking changes. Here's what we ended up with https://github.com/JetBrains/compose-jb/pull/2297

Thank you for raising this issue and for your help!

PR #1984, opened 2022-03-29
Excluding multiplatform common sources on xcodegen yaml

DevSrSouza:
When running `./gradlew iosDeployIPhone8Debug` in a module multiplatform that have all platforms in the same place, the task currently fails when other targets (`android` `jvm` `desktop`) have files that is not allowed in a iOS build.

For example: `androidMain/resource` with files `webp` for launcher icon.

Reproduce: https://github.com/adrielcafe/voyager/pull/52


dima-avdeev-jb (approved):
Awesome, Thanks!

PR #1988, opened 2022-04-02
Remove unnecessary parameter to required

veyndan:
As [`required` is a boolean attribute](https://html.spec.whatwg.org/multipage/input.html#the-required-attribute), `required(false)` would confusingly make the input required.

### DIFF
--- a/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/Attrs.kt
+++ b/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/Attrs.kt
@@ -194,6 +194,14 @@ fun AttrsScope<HTMLInputElement>.placeholder(value: String) =
 fun AttrsScope<HTMLInputElement>.readOnly() =
     attr("readonly", "")
 
+@Deprecated(
+    message = "Please use `required()` without parameters. Use if..else.. if conditional behaviour required.",
+    replaceWith = ReplaceWith("required()", "org.jetbrains.compose.web.attributes.required"),
+    level = DeprecationLevel.WARNING
+)
+fun AttrsScope<HTMLInputElement>.required(value: Boolean = true) =
+    attr("required", value.toString())
+
 fun AttrsScope<HTMLInputElement>.required() =
     attr("required", "")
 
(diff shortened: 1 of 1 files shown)


eymar:
@veyndan Thank you for your contribution.

Would you like to make some adjustments to ensure more smooth migration?

```
Index: web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/Attrs.kt
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/Attrs.kt b/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/Attrs.kt
--- a/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/Attrs.kt	(revision a027149b9527838798f8b590d7d9fa539b151473)
+++ b/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/Attrs.kt	(date 1649063275433)
@@ -194,8 +194,18 @@
 fun AttrsScope<HTMLInputElement>.readOnly() =
     attr("readonly", "")
 
-fun AttrsScope<HTMLInputElement>.required(value: Boolean = true) =
-    attr("required", value.toString())
+@Deprecated(
+    message = "Please use `required()` without parameters. Use if..else.. if conditional behaviour required.",
+    replaceWith = ReplaceWith("required()", "org.jetbrains.compose.web.attributes.required"),
+    level = DeprecationLevel.WARNING
+)
+fun AttrsScope<HTMLInputElement>.required(value: Boolean = true) : AttrsScope<HTMLInputElement> {
+    if (value) attr("required", "")
+    return this
+}
+
+fun AttrsScope<HTMLInputElement>.required() =
+    attr("required", "")
 
 fun AttrsScope<HTMLInputElement>.size(value: Int) =
     attr("size", value.toString())


 
```
The patch does:
- Fixes the behaviour of existing `requried(value: Boolean = true)` method
- Creates new method `required()` w/o parameters
- Deprecates existing method (even if it's fixed, it's better to remove it later to have consistent methods). It would help users to understand what needs to be done in their codebases.

veyndan:
I like the idea of deprecating the existing function and adding a new no-arg version. However, I don't think there is necessarily anything to "fix" with the existing function, as it does what it's supposed to do (which is just setting `required=$value`), even if that's surprising (as it is to me in regular html).

hfhbd:
@eymar web too :)

eymar:
thank you!

PR #1990, opened 2022-04-04
Intellij plugin to choose color

dima-avdeev-jb:
View color in editor gutter Color(0xffaabbcc)
image

Click to color icon and change in color palette
image

### DIFF
--- a/examples/intellij-plugin/src/main/kotlin/com/jetbrains/compose/color/ColorLineMarkerProvider.kt
+++ b/examples/intellij-plugin/src/main/kotlin/com/jetbrains/compose/color/ColorLineMarkerProvider.kt
@@ -44,7 +44,7 @@ class ColorLineMarkerProvider : LineMarkerProvider {
                     { mouseEvent, psiElement: PsiElement ->
 
                         class ChooseColorDialog() : DialogWrapper(project) {
-                            val colorState = mutableStateOf(0u)
+                            val colorState = mutableStateOf(0uL)
                             init {
                                 title = "Choose color"
                                 init()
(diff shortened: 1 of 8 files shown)


examples/intellij-plugin/src/main/kotlin/com/jetbrains/compose/color/ColorLineMarkerProvider.kt
```diff
@@ -0,0 +1,120 @@
+/*
```
igordmn, line 1: Could you move this example into 'intellij-plugin-with-experimental-shared-base' from 'intelij-plugin'? `intellij-plugin-with-experimental-shared-base` will be the recommended way to develop plugins for IDEA. Shared base bundles Compose, and other plugins use it. Without this shared base, every Compose plugin would contain its own version of Compose.

igordmn, line 1: Could you move this example into 'intellij-plugin-with-experimental-shared-base' from 'intelij-plugin'? `intellij-plugin-with-experimental-shared-base` will be the recommended way to develop plugins for IDEA. Shared base bundles Compose, and other plugins use it. Without this shared base, every Compose plugin would contain its own version of Compose.

dima-avdeev-jb: Yeah, sure!

dima-avdeev-jb:
@igordmn 
code moved to "intellij-plugin-with-experimental-shared-base" project

igordmn (approved):

PR #1994, opened 2022-04-08
Intellij compose tool window panel

dima-avdeev-jb:
Made panel in Intellij tool window
image


examples/intellij-plugin-with-experimental-shared-base/src/main/kotlin/com/jetbrains/compose/panel/CounterPanel.kt
```diff
@@ -0,0 +1,37 @@
...
+import androidx.compose.foundation.layout.Column
+import androidx.compose.material.Button
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.*
+
+@Composable
+fun CounterPanel(stateWithIdeLifecycle: MutableState<CounterState>) {
+    var stateIDE by remember { stateWithIdeLifecycle }
```
igordmn, line 17: If we inject state outside of Composable, it is better to not remember it (i.e., just use `stateWithIdeLifecycle`). The user of this Composable could want to inject a new state instance, for example.

igordmn (approved):

examples/intellij-plugin-with-experimental-shared-base/src/main/kotlin/com/jetbrains/compose/panel/CounterPanel.kt
dima-avdeev-jb: Done

PR #1997, opened 2022-04-11
Compose multiplatform library template

igordmn:


templates/multiplatform-library-template/library/build.gradle.kts
```diff
@@ -0,0 +1,84 @@
...
+    jvm("desktop")
+
+    sourceSets {
+        named("commonMain") {
+            dependencies {
+                api(compose.runtime)
+                api(compose.foundation)
+                api(compose.material)
+                // Needed only for preview
+                implementation(compose.preview)
```
DevSrSouza, line 25: I think these dependencies should `compileOnly` to make the library also compatible with Android only projects.

PR #2015, opened 2022-04-19
Added animated image component

JetpackDuba:
Animated images support (at least GIF & WebP) for issue https://github.com/JetBrains/compose-jb/issues/153


JetpackDuba:
By the way, @igordmn  @olonho not sure if the PR should already the methods & demo documented in detail or wait until it gets reviewed.

components/AnimatedImage/demo/src/jvmMain/kotlin/org/jetbrains/compose/splitpane/demo/Main.kt
```diff
@@ -0,0 +1,43 @@
+package org.jetbrains.compose.splitpane.demo
```
igordmn, line 1: Package and directory should be `org.jetbrains.compose.animatedimage.demo`

components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/DesktopAnimatedImage.kt
```diff
@@ -0,0 +1,110 @@
...
+    return remember {
+        if (url.startsWith("http://") || url.startsWith("https://")) {
+            NetworkLoader(url)
+        } else
+            LocalLoader(url)
+    }
+}
+
+@Composable
+fun rememberResourceAnimatedImage(resourcePath: String): AnimatedImageLoader {
```
igordmn, line 94: We can make it common too. Each platform has internal resources

components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/NetworkLoader.kt
```diff
@@ -0,0 +1,27 @@
...
+import java.net.URL
+
+internal class NetworkLoader(private val imageUrl: String) : AnimatedImageLoader {
+    var cachedBytes: ByteArray? = null
+
+    override suspend fun loadBytes(): ByteArray = withContext(Dispatchers.IO) {
+        var bytesArray: ByteArray? = cachedBytes
+
+        if (bytesArray == null) {
+            bytesArray = loadNetworkResource(imageUrl)
```
igordmn, line 15: Maybe we simplify it into `URL(imageUrl).readBytes()`? We don't need another withContext here.

components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/LocalLoader.kt
```diff
@@ -0,0 +1,21 @@
...
+import java.io.FileInputStream
+
+class LocalLoader(private val imageUrl: String) : AnimatedImageLoader {
+    var cachedBytes: ByteArray? = null
+
+    override suspend fun loadBytes(): ByteArray = withContext(Dispatchers.IO) {
+        var bytesArray: ByteArray? = cachedBytes
+
+        if (bytesArray == null) {
+            bytesArray = FileInputStream(imageUrl).readAllBytes() ?: byteArrayOf()
```
igordmn, line 14: nit: better to use readBytes extension, it returns `String` instead of `String!`

components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/LocalLoader.kt
```diff
@@ -0,0 +1,21 @@
+package org.jetbrains.compose.animatedimage
+
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.FileInputStream
+
+class LocalLoader(private val imageUrl: String) : AnimatedImageLoader {
```
igordmn, line 7: If we make `suspend load*` function as described earlier, we can make LocalLoader and ResourceLoader internal. It slightly less verbose (`val image = loadExternalAnimatedImage("./image.png")` instead of `val image = FileAnimatedImageLoader("./image.png").load()`).

Otherwise, we need to rename it to `FileAnimatedImageLoader` (because it should contain info about AnimatedImage and info about loading from File), and do the same for other loaders.

igordmn (changes requested):
Thanks!

components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/DesktopAnimatedImage.kt
JetpackDuba: Oh oke, as I have never used Compose for Web I wasn't sure.

JetpackDuba:
I have changed the whole component to make a combination of both both options mentioned in this [comment](https://github.com/JetBrains/compose-jb/pull/2015#discussion_r855372185).

I have encapsulated the `Codec` within the `AnimatedImage` class and exposed an API that allows the user to handle the UI depending on the load state, this way we can use the existing `Image`. 

I have also a couple of additional methods:
- Load synchronously the `AnimatedImage` (local, network & resources)
- Use async load with a custom implementation of AnimatedImageLoader (`asyncAnimatedImageLoaderState`)

The changes can be summed up in the following code snippets:

```
expect class AnimatedImage

expect suspend fun loadAnimatedImage(path: String): AnimatedImage

expect suspend fun loadResourceAnimatedImage(path: String): AnimatedImage

@Composable
expect fun asyncAnimatedImageLoaderState(path: String): AnimatedImageLoaderState

@Composable
expect fun asyncResourceAnimatedImageLoaderState(path: String): AnimatedImageLoaderState

@Composable
expect fun asyncAnimatedImageLoaderState(animatedImageLoader: AnimatedImageLoader): AnimatedImageLoaderState
```

Use for async load
```
val asyncAnimatedImage = asyncAnimatedImageLoaderState(url) 

when (asyncAnimatedImage) {
    is AnimatedImageLoaderState.Success -> {
        Image(
            bitmap = animatedImage(asyncAnimatedImage.animatedImage),
            contentDescription = null,
        )
    }
    AnimatedImageLoaderState.Loading -> {
        CircularProgressIndicator()
    }
    is AnimatedImageLoaderState.Error -> {
        Text("Error!")
    }
}
```

Not really sure about the methods naming tho, not really convincing.

components/AnimatedImage/demo/src/jvmMain/kotlin/org/jetbrains/compose/animatedimage/demo/Main.kt
```diff
@@ -0,0 +1,60 @@
...
+    val preloadedImageValue = preloadedImage.value
+
+    LaunchedEffect(Unit) {
+        // Load an image sync
+        preloadedImage.value = loadAnimatedImage(url) // use "loadResourceAnimatedImage" for resources
+    }
+
+    // Load an image async
+    val asyncAnimatedImage =
+        asyncAnimatedImageLoaderState(url)  // use "asyncResourceAnimatedImageLoaderState" for resources
```
igordmn, line 29: I am a little worried about this name being too long. What if we create a common function for loading something?

```
    // Load an image async
    val image = load { loadAnimatedImage(url) }  // use "load { loadResourceAnimatedImage(url) }" for resources

    Column {
        Image(
            loadOrNull { loadAnimatedImage(url) }?.animate() ?: ImageBitmap.Blank,
            contentDescription = null,
            Modifier.size(100.dp)
        )

        when (image) {
            is LoadState.Success -> Image(
                bitmap = image.value.animate(),
                contentDescription = null,
            )
            is LoadState.Loading -> CircularProgressIndicator()
            is LoadState.Error -> Text("Error!")
        }

        ...
    }
```

Also, we can rename this library to `resources`, and put common loading logic for Android/desktop/web/etc in the future (and when it will be ready for core Compose, we can move it there, and deprecate here):
```
package org.jetbrains.compose.resources

sealed class LoadState<T> {
    class Loading<T> : LoadState<T>()
    data class Success<T>(val value: T) : LoadState<T>()
    data class Error<T>(val exception: Exception) : LoadState<T>()
}

@Composable
fun <T> load(load: suspend () -> T): LoadState<T> {
    var state: LoadState<T> by remember(load) { mutableStateOf(LoadState.Loading()) }
    LaunchedEffect(load) {
        state = try {
            LoadState.Success(load())
        } catch (e: Exception) {
            LoadState.Error(e)
        }
    }
    return state
}

@Composable
fun <T: Any> loadOrNull(load: suspend () -> T): T? {
    val state = load(load)
    return (state as? LoadState.Success<T>)?.value
}

private val BlankBitmap = ImageBitmap(1, 1)
val ImageBitmap.Companion.Blank get() = BlankBitmap

@Composable
fun AnimatedImage.animate(): ImageBitmap {
    ...
}

// AnimatedImageLoader can be private in this case
```

We plan to do it anyway, so if you help now, we appreciate it :)

components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/LocalAnimatedImageLoader.kt
```diff
@@ -0,0 +1,21 @@
+package org.jetbrains.compose.animatedimage
+
+import java.io.FileInputStream
+
+internal class LocalAnimatedImageLoader(private val imageUrl: String) : AnimatedImageLoader() {
+    var cachedBytes: ByteArray? = null
+
+    override suspend fun generateByteArray(): ByteArray {
```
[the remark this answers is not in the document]
JetpackDuba: I omitted it because the method that calls this one already implement `withContent`, but yeah, can be added in case it's called from somewhere else in the future & to get rid of the warning in IDEA :+1:

components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/DesktopAnimatedImage.kt
```diff
@@ -0,0 +1,120 @@
...
+        } catch (ex: Exception) {
+            asyncAnimatedImageState.value = AnimatedImageLoaderState.Error(ex)
+        }
+    }
+
+    return asyncAnimatedImageState.value
+}
+
+@Composable
+fun animatedImage(animatedImage: AnimatedImage): ImageBitmap {
```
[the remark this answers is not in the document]
JetpackDuba: Oopsie, missed that one. Thanks! First time using kotlin multiplatform.

components/AnimatedImage/demo/src/jvmMain/kotlin/org/jetbrains/compose/animatedimage/demo/Main.kt
JetpackDuba: Agreed with the naming being too long, that's why I wasn't sure about it.

Making it as a generic method would be great to make it available to everyone + fix the naming issue, lgtm. I'll update the PR with the suggestions.

However, I am not sure about renaming the lib to `resources`.  It's great for the loading methods but the animated image methods seem a bit off here, as it's something used in a very specific feature that most people don't expect to find in a lib with such name.

What do you think about having 2 different libs, one for the resources and the other for the animated image stuff?

igordmn: > However, I am not sure about renaming the lib to resources

Okay, let's create two separate libraries for now. If we change decision, we can deprecate `animatedimage` library and move it to `resources` library. By the way, could you add 'animatedImage` to [this](https://github.com/JetBrains/compose-jb/blob/f281b75f7731269065cdaa861bed73ef2c24373e/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt#L198) place (`resources` let's keep "private" for now, only as "api" dependency of "animatedImage")?

The plan is to put in `resources` things, which help to load some abstract resources from external/internal sources in commonMain code. Maybe to put some tooling into this library (for example, pack internal resources as part of native distribution). It is also possible, that we will add ability to load concrete resources (images/animations/fonts/strings). But it is very early concept, so I agree, let's don't confuse people, putting only support for animations.

JetpackDuba: I've pushed the changes, let me know what do you think.

igordmn (approved):
Great! Thanks!

igordmn:
> By the way, @igordmn @olonho not sure if the PR should already have the methods & demo documented in detail or wait until it gets reviewed.

Also, we forgot to write docs for public methods :). Could you make a separate PR with docs?

JetpackDuba:
> > By the way, @igordmn @olonho not sure if the PR should already have the methods & demo documented in detail or wait until it gets reviewed.
> 
> Also, we forgot to write docs for public methods :). Could you make a separate PR with docs?

Sure, I'll do it this weekend probably.

components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/DesktopAnimatedImage.kt
```diff
@@ -0,0 +1,86 @@
...
+import androidx.compose.runtime.*
+import androidx.compose.ui.graphics.ImageBitmap
+import androidx.compose.ui.graphics.asComposeImageBitmap
+import org.jetbrains.skia.AnimationFrameInfo
+import org.jetbrains.skia.Bitmap
+import org.jetbrains.skia.Codec
+import java.net.MalformedURLException
+import java.net.URL
+
+private const val DEFAULT_FRAME_DURATION = 50
```
Nohus, line 13: I believe browsers usually default to 10 FPS for GIFs with no delay specified. Would it be worth changing this to 100 for consistency?

PR #2050, opened 2022-05-03
Docs to AnimatedImage public methods

JetpackDuba:
Sorry for the delay. 

I've added some docs to the public methods of the AnimatedImage & Resources components.

Let me know if you would like to change or add something.

### DIFF
--- a/components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
+++ b/components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
@@ -2,13 +2,30 @@ package org.jetbrains.compose.animatedimage
 
 import androidx.compose.ui.graphics.ImageBitmap
 
+/**
+ * Represents an image formed by a sequence of frames with a specific duration each one.
+ */
 expect class AnimatedImage
 
+/**
+ * Loads an [AnimatedImage] from a given path.
+ */
 expect suspend fun loadAnimatedImage(path: String): AnimatedImage
 
+/**
+ * Loads an [AnimatedImage] from the resources.
+ */
 expect suspend fun loadResourceAnimatedImage(path: String): AnimatedImage
 
+/**
+ * Animates an [AnimatedImage] by returning an [ImageBitmap] for each frame of the image.
+ * The parent will be with the recomposed with each new frame that has been rendered.
+ */
 expect fun AnimatedImage.animate(): ImageBitmap
 
 private val BlankBitmap = ImageBitmap(1, 1)
+
+/**
+ * Object used to represent a blank ImageBitmap with the minimum possible size.
+ */
 val ImageBitmap.Companion.Blank get() = BlankBitmap
--- a/components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
+++ b/components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
@@ -19,7 +19,7 @@ expect suspend fun loadResourceAnimatedImage(path: String): AnimatedImage
 
 /**
  * Animates an [AnimatedImage] by returning an [ImageBitmap] for each frame of the image.
- * The parent will be with the recomposed with each new frame that has been rendered.
+ * The caller will be with the recomposed with each new frame that has been rendered.
  */
 expect fun AnimatedImage.animate(): ImageBitmap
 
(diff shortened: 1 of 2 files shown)


components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
```diff
@@ -2,13 +2,30 @@ package org.jetbrains.compose.animatedimage
...
 expect suspend fun loadAnimatedImage(path: String): AnimatedImage
 
+/**
+ * Loads an [AnimatedImage] from the resources.
+ */
 expect suspend fun loadResourceAnimatedImage(path: String): AnimatedImage
 
+/**
+ * Animates an [AnimatedImage] by returning an [ImageBitmap] for each frame of the image.
+ * The caller will be with the recomposed with each new frame that has been rendered.
```
igordmn, line 22: > The caller will be with the recomposed with each new frame that has been rendered.

Maybe "The caller will be recomposed with each new frame that has been rendered." is more correct.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/LoadState.kt
```diff
@@ -7,42 +7,87 @@ import androidx.compose.runtime.remember
...
+ */
 sealed class LoadState<T> {
     class Loading<T> : LoadState<T>()
     data class Success<T>(val value: T) : LoadState<T>()
     data class Error<T>(val exception: Exception) : LoadState<T>()
 }
 
+
+/**
+ * Load a block of code with the operation progress and result of [T] wrapped around a [LoadState].
```
igordmn, line 21: Maybe "Load an item with type [T] asynchronously, and notify the caller about the load state."

components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
JetpackDuba: Oopsie, my sentence is completly incorrect. Not sure what I was thinking about! Will fix it in a moment

components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
```diff
@@ -2,13 +2,30 @@ package org.jetbrains.compose.animatedimage
 
 import androidx.compose.ui.graphics.ImageBitmap
 
+/**
+ * Represents an image formed by a sequence of frames with a specific duration each one.
+ */
 expect class AnimatedImage
 
+/**
+ * Loads an [AnimatedImage] from a given path.
```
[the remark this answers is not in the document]
JetpackDuba: I thought about it but I wasn't completly sure, as the docs it's in the commonMain, so the sentence wouldn't be true if the method was used in compose for mobile/web. Should I still update it? Or can the docs be changed just for the desktop module?

components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
igordmn: Oh, you are right. But let's write that it is platform specific, and add info about each supported platform (we support only desktop at the moment)

components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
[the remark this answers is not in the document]
JetpackDuba: > Oh, you are right. But let's write that it is platform specific, and add info about each supported platform (we support only desktop at the moment)

Shouldn't we add it in the readme file rather than in the methods docs? Talking about the README, should I add one with a summary about the component?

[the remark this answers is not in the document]
igordmn: It is rare when people read Readme :). If we can provide information closer to code, we should do that.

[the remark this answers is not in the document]
JetpackDuba: Got it! I'll push it in a moment then

components/AnimatedImage/library/src/commonMain/kotlin/org/jetbrains/compose/animatedimage/AnimatedImage.kt
```diff
@@ -9,6 +9,8 @@ expect class AnimatedImage
 
 /**
  * Loads an [AnimatedImage] from a given path.
+ * The path can be loaded from different sources depending on the platform (only desktop currently):
```
igordmn, line 12: > The path

We load an image, not a path :). So we should write "The image can be..."

JetpackDuba: Fixed :P

igordmn (approved):

igordmn:
Thanks!

PR #2079, opened 2022-05-28
Add visibility property

InsanusMokrassar:
By this PR I am adding opportunity to use visibility property according to the [visibility references](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility)


Thomas-Vos:
I guess this should be a `.kt` file and not a `.css` file?

InsanusMokrassar:
Oh sorry, my mistake. Thank you :)

eymar:
could you please add some tests for this property?
`CSSDisplayTests` is an example of similar tests

InsanusMokrassar:
@eymar Hello and sorry with this long delay. I have actualized this PR: replaced visibility extension, replaced visibility enums and add tests. Could you please review it?

eymar (approved):

PR #2080, opened 2022-05-28
CMP-5906 Add option to enable Class data sharing (CDS)

Thomas-Vos:
[CMP-5906](https://youtrack.jetbrains.com/issue/CMP-5906) Add option to enable Class data sharing (CDS)

Alternatively this could be enabled by default, what do you think?

This PR adds an option for CDS, not for AppCDS.

AppCDS is probably not worth it. Generating it at first startup slows down the startup. So it should probably be generated at install time somehow (not sure if even possible, and would take too much time). Generating it at build time in this gradle plugin would be another option (which I think is best), but that would increase app size. AppCDS would require more investigation before adding it. I think what would be ideal, is that the gradle plugin would launch the app in the background when building, and maybe run some sort of compose test to warm up code. Then exit and include the generated file in the archive. Does this make sense?

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
@@ -19,6 +19,8 @@ abstract class JvmApplicationDistributions : AbstractDistributions() {
     }
     var includeAllModules: Boolean = false
 
+    var cds: Boolean = false
+
     val linux: LinuxPlatformSettings = objects.newInstance(LinuxPlatformSettings::class.java)
     open fun linux(fn: Action<LinuxPlatformSettings>) {
         fn.execute(linux)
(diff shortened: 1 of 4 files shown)


igordmn:
It will be great to have this feature.

@AlexeyTsvetkov, could you look at the PR?

eskatos:
AppCDS makes starting up Compose Desktop applications muuuuuch faster.
I tried to fiddle around to enable this without a patch but could not figure out a way.
Could this be considered again? Pretty please :)

eskatos:
For readers who want to enable CDS Today, here's a dirty workaround:
```kotlin
afterEvaluate {
    tasks.named<AbstractJLinkTask>("createRuntimeImage") {
        freeArgs.addAll("--generate-cds-archive")
        @Suppress("UNCHECKED_CAST")
        val stripNativeCommands = this::class.java.methods
            .single { it.name.startsWith("getStripNativeCommands") }
            .invoke(this) as Property<Boolean>
        stripNativeCommands.set(false)
        doLast {
            destinationDir.get().dir("bin").asFile.deleteRecursively()
        }
    }
}
```

Thomas-Vos:
Nice, I am curious what AppCDS could do for a compose app in addition to CDS.

eskatos:
I got AppCDS working on top of CDS. I did not measure precisely but it feels snappier.
```kotlin
compose.desktop {
  application {
    nativeDistributions {
      jvmArgs += "-Xlog:cds=error" // change to 'debug' to get more info
      jvmArgs += "-Xshare:auto"
      jvmArgs += "-XX:+AutoCreateSharedArchive"
      jvmArgs += "-XX:SharedArchiveFile=path/to/app-cds.jsa"
    }
  }
}
```
One painful thing is that you need to provide a path relative the the JVM working directory for the jsa file.

Thomas-Vos:
> I got AppCDS working on top of CDS.

I just gave it a try, and it seems to generate the app-cds.jsa as expected. However, the first startup now takes about 30 seconds, so that is pretty bad. I guess it would be better to somehow generate this file in the gradle plugin automatically and include it at build time.

Thomas-Vos:
Hi @igordmn, could you please review this or assign it to the correct person?

kropp:
AppCDS and AOT has been supported in #5644

PR #2084, opened 2022-05-30
Update README.md

marunrun:
useResource to load the images


marunrun:
> The purpose of this example is to show how to load images from the device storage, not from resources folder.



OK

PR #2099, opened 2022-06-06
Add a link to the BrowserView project

Danil-Didkovskiy:
BrowserVIew is a [JxBrowser](https://www.teamdev.com/jxbrowser) component for Compose Desktop that displays content loaded in a Chromium-based web browser. This PR adds a link to the [BrowserView repo](https://github.com/vlad-lubenskyi/compose-browser-view) to the list of experimental components and integrations in README.md. 

Currently, the component has minimal functionality and supports mouse and keyboard events.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ Preview functionality (check your application UI without building/running it) fo
    * [components](components) - custom components of Compose Multiplatform
        * [Split Pane](components/SplitPane)
    * [experimental](experimental) - experimental components and examples
+       * [BrowserView](https://github.com/vlad-lubenskyi/compose-browser-view) - An experimental [JxBrowser](https://www.teamdev.com/jxbrowser) component for Compose Desktop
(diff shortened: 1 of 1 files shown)


pacher:
In my opinion commercial components should be mentioned somewhere else like in a completely separate section or even file with all the appropriate disclaimers.

igordmn:
Link to commercial project only makes sense if there are another free alternatives, and there is a list of third party solutions. We don't plan to make such list in the near future, because it requires some effort to keep it in the actual state.

vladimir-ikryanov:
@igordmn there's a free alternative mentioned in [README.md](https://github.com/JetBrains/compose-jb/blob/master/README.md) — it's [CEF](https://github.com/JetBrains/compose-jb/blob/master/experimental/cef).

We saw it and decided to help the project and increase the number of integrations with Compose. We provided a commercial alternative for the Compose projects where CEF is not an option due to some of its disadvantages. It's probably the first commercial third-party solution that is integrated with Compose. But it looks like Compose is only for the Open-Source projects.

PR #2126, opened 2022-06-17
Web: Support attrs extension functions in sub elements too

hfhbd:
Eg SvgRectElement extends SvgElement, but `width` is defined in `AttrsScope<SvgElement>` which is not usable in SvgRectElement attrs builder because the element types don't match.


eymar:
Thank you!

PR #2130, opened 2022-06-18
Workaround for empty jvmTarget

hfhbd:
With Kotlin 1.7 `jvmTarget` is null (or blank) by default if not set explicit, so `toDouble` throws an NPE during project configuration with a jvm target.
https://youtrack.jetbrains.com/issue/KT-52857


hfhbd:
Only happens with Kotlin 1.7 https://youtrack.jetbrains.com/issue/KT-52857/KGP-with-AGP-721-jvmTarget-is-null

eymar:
Thank you!

Issue #2114, opened 2022-06-14
"falling-balls-mpp" and "minesweeper" fail on iosSimulator


mipastgt:
The experimental examples "falling-balls-mpp" and "minesweeper" fail on iosSimulator.

    ./gradlew iosDeployIPhone8Debug

results in:

```
> Task :iosSimulatorCreateIPhone8 FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Could not evaluate onlyIf predicate for task ':iosSimulatorCreateIPhone8'.
> Field 'dataPathSize' is required for type with serial name 'org.jetbrains.compose.experimental.uikit.internal.DeviceData', but it was missing
```

This was on macOS 12.4 Monterey.

mipastgt:
A discussion and potential solution of the problem can be found here:
 https://kotlinlang.slack.com/archives/C0346LWVBJ4/p1655283852255309

TaehoonLeee:
What's the xcode version?

I faced same issue, but  I solved it by updating the xcode

mipastgt:
That's Xcode "Version 13.4.1 (13F100)" which is the latest official release, according to https://xcodereleases.com/,  on a macMini with macOS 12.4 Monterey.

dima-avdeev-jb:
Thanks for Issue! We will try to fix it!

dima-avdeev-jb:
I prepared PR with fix. We need to wait code review and after need to deploy new version.
I'll let you know when a new version is ready for testing.

mipastgt:
Thanks a lot. Do you have a rough estimate when a new version might be available? (Just for my own planning.)

dima-avdeev-jb:
> Thanks a lot. Do you have a rough estimate when a new version might be available? (Just for my own planning.)

I think 3-5 days

dima-avdeev-jb:
Fixed in build **1.2.0-alpha01-dev724**

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-2114">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #2131, opened 2022-06-19
Fix issue 2114. Gradle plugin with Xcode 13.4

dima-avdeev-jb:
Fix for bug https://github.com/JetBrains/compose-jb/issues/2114
Add default null values for unused json params, because in Xcode 13.4 some params are missing.
And before create new simulator - delete all unavailable simulators.


igordmn:
> And before create new simulator - delete all unavailable simulators.

What if the user don't want unavailable simulators to be deleted? For example, they created some simulator manually, not with our plugin, then it become unavailable, then user wants to create a new simulator, based on unavailable one, reading parameters from it.

igordmn:
> Yeah, it have sence. I can change logic to remove only simulator with current deploy name

This would be better

dima-avdeev-jb:
> > Yeah, it have sence. I can change logic to remove only simulator with current deploy name
> 
> This would be better

done

igordmn (approved):

PR #2152, opened 2022-06-30
Fix CI: Add yarn.lock to ensure non failing builds

hfhbd:
https://teamcity.jetbrains.com/buildConfiguration/JetBrainsPublicProjects_Compose_Publish_2_Subtasks_1ComposePublishNativeOptional/3887770?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true failed, because it tries to fetch @types-eslint 8.4.4 which is indeed not available: https://registry.yarnpkg.com/@types/eslint/-/eslint-8.4.4.tgz

Added a working lock file to prevent build failures due to dynamic dependencies.


eymar:
Thanks for your help! :) 

Did you amend the generated yarn.lock manually? The generated one has some newer version vs the yarn.lock in this PR.
I pushed the generated one for now, the build passed locally, now CI runs.

hfhbd:
Strange, no, I generated the lock file yesterday. Maybe there was a release in the mean time 🤔

PR #2163, opened 2022-07-04
Update README.md

erikhuizinga:
The previous source code link led to an HTTP 404 page.

### DIFF
--- a/tutorials/Web/Events_Handling/README.md
+++ b/tutorials/Web/Events_Handling/README.md
@@ -47,7 +47,7 @@ Form(attrs = {
 ```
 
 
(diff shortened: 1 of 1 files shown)


igordmn:
Thanks!

But it is better to keep the link to the master branch. Docs in master branch should be in sync with code. Because otherwise it will be harder to keep docs in sync - at this point of time it is more likely that we add methods to this interface than rename this file.

I made changes in your PR and added a comment [here](https://github.com/JetBrains/compose-jb/blob/b60de289869e6cc12faa65fb553bc453add998ab/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/EventsListenerScope.kt#L3). Also, because now we point to the master branch, better to not point to specific line.

Issue #2112, opened 2022-06-13
Unsupported Java 8?


soywiz:
https://github.com/korlibs/korge-next/runs/6829604895?check_suite_focus=true#step:7:7449

I'm getting this stacktrace when running it with Java8:

```kotlin
e: java.lang.NoSuchMethodError: java.lang.Integer.parseUnsignedInt(Ljava/lang/CharSequence;III)I
	at androidx.compose.compiler.plugins.kotlin.inference.SchemeStringSerializationReader.number(Scheme.kt:342)
file or directory '/Users/runner/work/korge-next/korge-next/korge-compose/src/posixMain/kotlin', not found
	at androidx.compose.compiler.plugins.kotlin.inference.SchemeKt.deserializeScheme$item(Scheme.kt:209)
file or directory '/Users/runner/work/korge-next/korge-next/korge-compose/src/nativeMain/kotlin', not found
```

Is this intended?

kirill-grouchnikov:
Compose needs 11 at least. This is documented.

kirill-grouchnikov:
https://github.com/JetBrains/compose-jb/tree/master/tutorials/Getting_Started#new-project-wizard

soywiz:
Thanks for the info! Maybe it could check the runtime before and providing a meaningful error when configuring the plugin? 
Also, is it final the JVM 11 version requirement? Any chance to allow Java8? Since this is Kotlin targeting 1.8 should be easy right?

igordmn:
Supporting Java 8 will add some burden. For example we won't be able to used the new API, which was added in JDK 9-11, and we have to live with the bugs, which were fixed in the newer versions of JDK. JDK 11 is a balance between the need for a lower version and a burden which this support adds. 

> Maybe it could check the runtime before and providing a meaningful error when configuring the plugin

This check was in the past, probably something is broken, or accidentally removed. We will fix this.

soywiz:
Oh, that's a pity. The Kotlin compiler still supports Java8, as well as proguard and all the other plugins I was using. So including compose would force users of my libraries to migrate to Java11.

I used to believe that by using Kotlin you could target Java8 and only get APIs available there so avoiding using APIs not available by mistake, and that Kotlin already provide a lot of goodies to make development enjoyable even in Java8. I also thought that Java8 had many patch releases fixing bugs, while still allowing people to using Java 11 or greater if they want.

kirill-grouchnikov:
[...]

PR #2164, opened 2022-07-05
Gradle Plugin: Fix NPE and use 1.8 by default

hfhbd:
At the moment, the [740 build failed](https://teamcity.jetbrains.com/buildConfiguration/JetBrainsPublicProjects_Compose_Publish_2_AllGitHubRelease/3890160?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&expandBuildProblemsSection=true&expandBuildTestsSection=true) because with Kotlin 1.7 jvmTarget is null (or blank) by default if not set explicit, so toDouble throws an NPE during project configuration with a jvm target.
https://youtrack.jetbrains.com/issue/KT-52857

Based on https://github.com/JetBrains/compose-jb/issues/2112#issuecomment-1159521628 I set the jvmTarget to 1.8


igordmn:
Thanks!

I will also cherry-pick the rest of the commits from [release1.2-integration](https://github.com/JetBrains/compose-jb/compare/release1.2-integration)

PR #2174, opened 2022-07-07
Fix AnimatedImage component

orangy:
Fixes and optimizes cases with 0 and 1 frames, changes default fps to 10 to match modern browsers.

### DIFF
--- a/components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/DesktopAnimatedImage.kt
+++ b/components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/DesktopAnimatedImage.kt
@@ -27,7 +27,7 @@ actual suspend fun loadResourceAnimatedImage(path: String): AnimatedImage {
 @Composable
 actual fun AnimatedImage.animate(): ImageBitmap {
     when (codec.frameCount) {
-        0 -> return ImageBitmap(0, 0) // No frames at all
+        0 -> return ImageBitmap.Blank // No frames at all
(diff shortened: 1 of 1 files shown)


components/AnimatedImage/library/src/desktopMain/kotlin/org/jetbrains/compose/animatedimage/DesktopAnimatedImage.kt
```diff
@@ -26,42 +26,50 @@ actual suspend fun loadResourceAnimatedImage(path: String): AnimatedImage {
...
-            animation = keyframes {
-                durationMillis = 0
-                for ((index, frame) in codec.framesInfo.withIndex()) {
-                    index at durationMillis
-                    val frameDuration = calcFrameDuration(frame)
-
-                    durationMillis += frameDuration
-                }
+    when (codec.frameCount) {
+        0 -> return ImageBitmap(0, 0) // No frames at all
```
igordmn, line 30: Compose have a bug (or a strange feature), when we use bitmaps with zero size:
```
Exception in thread "main" java.lang.IllegalStateException: Size is unspecified
	at androidx.compose.ui.geometry.Size.getWidth-impl(Size.kt:48)
	at androidx.compose.ui.draw.PainterModifier.draw(PainterModifier.kt:267)
	at androidx.compose.ui.node.DrawEntity.draw(DrawEntity.kt:98)
	at androidx.compose.ui.node.LayoutNodeWrapper.drawContainedDrawModifiers(Layout
```

Use `ImageBitmap.Blank` here, which has size = 1, until it will be fixed

PR #2183, opened 2022-07-11
Correct information about releases in main README.md

igordmn:
The current link leads to the list of dev releases, and can confuse people if they want to find the latest stable, not the latest dev release.

The alternative is to remove this section and keep only "default" places to show the current releases:
image

But probably it is better to write about dev releases explicitly.


jimgoog:
looks good 👍

PR #2228, opened 2022-08-04
Add support for transition in CSS api.

Ayfri:
This PR adds a builder to create transitions with the CSS API.


eymar:
Hi @Ayfri, Thank you for your contribution!

Could you please add unit tests for all new APIs in `.../web/core/src/jsTest`? Look at `TransformTests` as an example. We simply verify that the built css string is correct. 

Also, since this a new API, we usually mark it with `@ExperimentalComposeWebStyleApi` annotation. It needs to be added to all new classes and extension functions.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,81 @@
...
+		transitions += transition
+	}
+	
+	fun duration(value: CSSSizeValue<out CSSUnitTime>) = apply { defaultDuration = value }
+	fun timingFunction(value: AnimationTimingFunction) = apply { defaultTimingFunction = value }
+	fun delay(value: CSSSizeValue<out CSSUnitTime>) = apply { defaultDelay = value }
+	fun ease(ease: AnimationTimingFunction) = timingFunction(ease)
+}
+
+inline fun StyleScope.transitions(transitions: Transitions.() -> Unit) {
```
eymar, line 78: Would be nice to have some short doc here with an example:

```
@ExperimentalComposeWebStyleApi
/**
 * Example:
 * ```
 *  transitions {
 *    "width" { duration(1.s) }
 *   }
 * ```
 * @see https://developer.mozilla.org/en-US/docs/Web/CSS/transition
 */
 ```
 
 Feel free to add more interesting example.

**Comment by Ayfri** [comment]
I'm creating the unit tests and facing a problem where we have to decide how the API behave :
```kotlin
transitions {
	duration(1.s)
	delay(.5.s)
	properties("height", "width")
	
	"width" {
		duration(2.s)
	}
}
```
Should it returns 
 1. `height 1s 0.5s, width 2s` by replacing the `width` property.
 2. `height 1s 0.5s, width 2s 0.5s` by modifying the `width` property. 
 2. `height 1s 0.5s, width 1s 0.5s, width 2s` by just adding the property, behaves same as 1 but has a repetition.
 
Currently, it has the third behavior.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,100 @@
...
+	
+	inline fun all(block: Transition.() -> Unit) = Transition().apply(block).also { transition ->
+		transition.property = "all"
+		transitions += transition
+	}
+	
+	fun duration(value: CSSSizeValue<out CSSUnitTime>) = apply { defaultDuration = value }
+	fun timingFunction(value: AnimationTimingFunction) = apply { defaultTimingFunction = value }
+	fun delay(value: CSSSizeValue<out CSSUnitTime>) = apply { defaultDelay = value }
+	fun ease(ease: AnimationTimingFunction) = timingFunction(ease)
```
eymar, line 77: And here as well. 
Is `ease` needed? It just calls timingFunction. Those 2 functions have the same parameter. 
So it seems like `timingFunction` should be enough and it named according to https://developer.mozilla.org/en-US/docs/Web/CSS/transition-timing-function. Do I miss something?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,100 @@
+/*
+ * Copyright 2020-2022 JetBrains s.r.o. and respective authors and developers.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package org.jetbrains.compose.web.css
+
+import org.jetbrains.compose.web.ExperimentalComposeWebApi
+
+data class Transition(
```
eymar, line 10:
```suggestion
@ExperimentalComposeWebApi
data class Transition(
```

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,100 @@
...
+		
+		duration?.let { result += " $it" }
+		timingFunction?.let { result += " $it" }
+		delay?.let { result += " $it" }
+		
+		return result
+	}
+}
+
+fun Transition.duration(value: CSSSizeValue<out CSSUnitTime>) = apply { duration = value }
```
eymar, line 28:
```suggestion
@ExperimentalComposeWebApi
fun Transition.duration(value: CSSSizeValue<out CSSUnitTime>) = apply { duration = value }
```

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,100 @@
...
+fun Transition.timingFunction(value: AnimationTimingFunction) = apply { timingFunction = value }
```
eymar, line 29:
```suggestion
@ExperimentalComposeWebApi
fun Transition.timingFunction(value: AnimationTimingFunction) = apply { timingFunction = value }
```

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,100 @@
...
+fun Transition.delay(value: CSSSizeValue<out CSSUnitTime>) = apply { delay = value }
```
eymar, line 30:
```suggestion
@ExperimentalComposeWebApi
fun Transition.delay(value: CSSSizeValue<out CSSUnitTime>) = apply { delay = value }
```

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,100 @@
...
+fun Transition.ease(ease: AnimationTimingFunction) = timingFunction(ease)
```
eymar, line 31: Do we need `ease` if we have `timingFunction`?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,100 @@
...
+
+data class Transitions(
```
eymar, line 33:
```suggestion
@ExperimentalComposeWebApi
data class Transitions(
```

eymar:
@Ayfri 
> 3\. `height 1s 0.5s, width 1s 0.5s, width 2s` by just adding the property, behaves same as 1 but has a repetition.

I think 3rd variant is okay, since it's possible to write such a css manually as well.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,100 @@
...
+	}
+	
+	inline fun properties(vararg properties: String, block: Transition.() -> Unit = {}) = properties.invoke(block)
+	
+	inline fun all(block: Transition.() -> Unit) = Transition().apply(block).also { transition ->
+		transition.property = "all"
+		transitions += transition
+	}
+	
+	fun duration(value: CSSSizeValue<out CSSUnitTime>) = apply { defaultDuration = value }
```
eymar, line 74: What do you think if we name these functions: `defaultDuration`, `defaultTimingFunction`, `defaultDelay`?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,100 @@
...
+			it.property
+		}.joinToString(", ") {
+			it.apply {
+				if (defaultDelay != null && delay == null) delay = defaultDelay!!
+				if (defaultDuration != null && duration == null) duration = defaultDuration!!
+				if (defaultTimingFunction != null && timingFunction == null) timingFunction = defaultTimingFunction!!
+			}.toString()
+		}
+	
+	inline operator fun String.invoke(block: Transition.() -> Unit) = Transition().apply(block).also {
```
eymar, line 50: Is it intentional  that this function returns the Transition as a result? What if it returns just Unit? (same question for the functions below)

Could you please specify the return type explicitly for every function, so we don't break it accidentally later?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
eymar: Also this and the functions below return not Unit as a result. Is it intentional?

Ayfri: It's a good idea, it could avoid doing `duration` for a property and having it resolved to default duration for example, I faced this problem while writing the unit tests.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,105 @@
...
+@ExperimentalComposeWebApi
+/**
+ * Examples:
+ * ```
+ *  transitions {
+ *    "width" { duration(1.s) }
+ *  }
+ *
+ *  transitions {
+ *    duration(.5.s)
```
eymar, line 94: Since the functions were renamed to `defaultDuration`, `defaultTimingFunction`, the example needs to be updated as well.

Thank you :)

eymar, line 94:
```suggestion
 *    defaultDuration(.5.s)
```

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/transitions.kt
```diff
@@ -0,0 +1,105 @@
...
+ *    timingFunction(AnimationTimingFunction.EaseInOut)
```
eymar, line 95:
```suggestion
 *    defaultTimingFunction(AnimationTimingFunction.EaseInOut)
```

eymar (approved):

PR #2256, opened 2022-08-22
Use another version of Compose Compiler inside Compose Gradle plugin

igordmn:
Google recently switched to a separate release cycle for Compose Compiler. It is more frequent (we already have Compose Compiler 1.3.0 stable) and supports recent Kotlin's.

We should do the same. We can't use compiler published by Google yet, because it is in Google maven repository, and don't have `-hosted` version.

I created [a task on CI](https://teamcity.jetbrains.com/buildConfiguration/JetBrainsPublicProjects_Compose_Publish_2_ComposePluginWithoutTests?mode=builds) for building our compiler manually from aosp repository, and removed it from auto-building.


igordmn:
[Successful build](https://teamcity.jetbrains.com/buildConfiguration/JetBrainsPublicProjects_Compose_Publish_2_All_2/3926992?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true) with this feature

eymar (approved):

jakobkmar:
Would it be possible to make the compose-jb Gradle plugin always apply the correct Compose compiler version for the Kotlin version the user actually has applied to Gradle via the Kotlin Gradle plugin?
If I am not mistaken the version is currently hardcoded.

igordmn:
> to make the compose-jb Gradle plugin always apply the correct Compose compiler version

It is a good idea. We will experiment with this, when Kotlin 1.7.20 and compiler for it will be released.

PR #2278, opened 2022-08-29
Simplify js condition

dima-avdeev-jb:
I found, that previous check fails on stable JS projects. Because skiko dependency exist's also in stable JS build. 

I found, that it's simpler to check only experimental extension for JS target:
```
    compose.experimental {
    web.application {}

```
Fixed.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/experimental/internal/checkExperimentalTargets.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/experimental/internal/checkExperimentalTargets.kt
@@ -6,6 +6,7 @@
 package org.jetbrains.compose.experimental.internal
 
 import org.gradle.api.Project
+import org.jetbrains.compose.experimental.dsl.ExperimentalExtension
 import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
 import org.jetbrains.kotlin.gradle.plugin.KotlinTarget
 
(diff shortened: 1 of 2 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/experimental/internal/checkExperimentalTargets.kt
```diff
@@ -6,31 +6,41 @@
...
 private val TargetType.gradlePropertyName get() = "org.jetbrains.compose.experimental.$id.enabled"
 
 private val EXPERIMENTAL_TARGETS: Set<TargetType> = setOf(
-    TargetType("uikit", presets = listOf("iosSimulatorArm64", "iosArm64", "iosX64")),
-    TargetType("macos", presets = listOf("macosX64", "macosArm64")),
-    TargetType("jscanvas", presets = listOf("jsIr", "js")),
+    TargetType("uikit", presets = listOf("iosSimulatorArm64", "iosArm64", "iosX64")) { true },
+    TargetType("macos", presets = listOf("macosX64", "macosArm64")) { true },
+    TargetType("jscanvas", presets = listOf("jsIr", "js")) { experimentalExtension ->
+        experimentalExtension.web._isApplicationInitialized
```
igordmn, line 30: Can JS canvas work if we omit DSL in `build.gradle`?
```
compose.experimental {
    web.application {}
```

dima-avdeev-jb: Theoretically yes. If user add's dependency "org.jetbrains.skiko:skiko-js-wasm-runtime:$version" by hands

dima-avdeev-jb: And unpack skiko wasm binary in correct directory with js resources

igordmn: Also, it is strange that we have skiko in compose runtime JS dependencies. If we have it, we should remove this dependency.

cc @eymar 

Does the project fail in web-template?

eymar: We never had skiko dependency in runtime. I'll double check tomorrow.

dima-avdeev-jb: Sorry, I have not correct say about stable JS build.
@eymar Yeah, you correct! org.jetbrains.skiko:skiko not exist's in stable js-html projects.

Also if newer projects have dependencies to skiko, it means that js is in experimental mode.
So, previously created PR work's good.
Current PR will be closed.

PR #2293, opened 2022-09-06
Fix demo run, add gradle.properties

dima-avdeev-jb:
Fix demo project's run. Right from README.md:

Run jvm desktop sample:
```bash
./scripts/runGradle run
```
another jvm desktop samples: 
```bash
./scripts/runGradle run1 run2 run3 run4 runWindowApi runVsync runLayout
```

Run wasm sample:
```bash
./scripts/runGradle runMppJs
```

Run native macos sample:
```bash
./scripts/runGradle runMppMacos
```

add more memory to gradle.properties

### DIFF
--- a/compose/gradle.properties
+++ b/compose/gradle.properties
@@ -1,0 +1,1 @@
+org.gradle.jvmargs=-Xmx8g -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC -Dkotlin.daemon.jvm.options=-XX:MaxMetaspaceSize=1g -Dlint.nullness.ignore-deprecated=true


igordmn (approved):

igordmn:
Thanks!

By the way, we should probably move to the androidx-only configuration in the future, because it simpler to use in compare to a composite project + bash scripts

dima-avdeev-jb:
> Thanks!
> 
> By the way, we should probably move to the androidx-only configuration in the future, because it simpler to use in compare to a composite project + bash scripts

Yeah, I think so

PR #2297, opened 2022-09-07
Temporary deprecate 3 overloads of StyleScope.borderWidth

eymar:
The reason is wrong parameter names. The behaviour is still correct. The idea is to fix the parameter names in 2.0 (which is more appropriate for breaking changes)

This PR is an alternative for https://github.com/JetBrains/compose-jb/pull/1982

### DIFF
--- a/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/border.kt
+++ b/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/border.kt
@@ -83,26 +83,115 @@ fun StyleScope.borderWidth(width: CSSNumeric) {
     property("border-width", width)
 }
 
+
+/**
+ * https://developer.mozilla.org/en-US/docs/Web/CSS/border-width
+ *
+ * First argument sets the vertical borders width.
+ * Second argument sets the horizontal borders width.
+ *
+ * NOTE: Temporary deprecation until 2.0 because of wrong parameters names.
+ * In 2.0 the parameter names should be fixed. Despite the wrong parameter names, the function behaves correctly.
+ */
+@Deprecated(
+    message = "This function has misleading parameter names. " +
+            "Despite that, it behaves correctly (see https://developer.mozilla.org/en-US/docs/Web/CSS/border-width)." +
+            "Therefore this function still can be used without specifying parameter names. Replacement is not necessary." +
+            " It'll remain Deprecated until 2.0, and in 2.0 parameter names should be changed to: " +
+            "`vertical: CSSNumeric` and `horizontal: CSSNumeric`",
+    replaceWith = ReplaceWith(
+        expression = "borderWidth2(vertical = topLeft, horizontal = bottomRight)"
+    )
+)
 fun StyleScope.borderWidth(topLeft: CSSNumeric, bottomRight: CSSNumeric) {
-    property("border-width", "$topLeft $bottomRight")
+    borderWidth2(topLeft, bottomRight)
 }
 
+/**
+ * This function is a temporary replacement for `borderWidth(topLeft: CSSNumeric, bottomRight: CSSNumeric)`.
+ * After a fix for parameters names in `borderWidth`,
+ * this function will be Deprecated with a replaceWith `borderWidth`. Later `borderWidth2` will be Removed.
+ * @see borderWidth
+ */
+fun StyleScope.borderWidth2(vertical: CSSNumeric, horizontal: CSSNumeric) {
+    property("border-width", "$vertical $horizontal")
+}
+
+
+/**
+ * https://developer.mozilla.org/en-US/docs/Web/CSS/border-width
+ *
+ * First argument sets the top border width.
+ * Second argument sets the horizontal borders width.
+ * Third argument sets the bottom border width.
+ *
+ * NOTE: Temporary deprecation until 2.0 because of wrong parameters names.
+ * In 2.0 the parameter names should be fixed. Despite the wrong parameter names, the function behaves correctly.
+ */
+@Deprecated(
+    message = "This function has misleading parameter names. " +
+            "Despite that, it behaves correctly (see https://developer.mozilla.org/en-US/docs/Web/CSS/border-width)." +
+            "Therefore this function still can be used without specifying parameter names. Replacement is not necessary." +
+            " It'll remain Deprecated until 2.0, and in 2.0 parameter names should be changed to: " +
+            "`top: CSSNumeric`, `horizontal: CSSNumeric`, `bottom: CSSNumeric`",
+    replaceWith = ReplaceWith(
+        expression = "borderWidth3(top = topLeft, horizontal = topRightAndBottomLeft, bottom = bottomRight)"
+    )
+)
 fun StyleScope.borderWidth(
     topLeft: CSSNumeric,
     topRightAndBottomLeft: CSSNumeric,
     bottomRight: CSSNumeric
 ) {
-    property("border-width", "$topLeft $topRightAndBottomLeft $bottomRight")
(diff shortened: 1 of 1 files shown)


web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/border.kt
```diff
@@ -83,26 +83,115 @@ fun StyleScope.borderWidth(width: CSSNumeric) {
...
+ * Second argument sets the horizontal borders width.
+ * Third argument sets the bottom border width.
+ *
+ * NOTE: Temporary deprecation until 2.0 because of wrong parameters names.
+ * In 2.0 the parameter names should be fixed. Despite the wrong parameter names, the function behaves correctly.
+ */
+@Deprecated(
+    message = "This function has misleading parameter names. " +
+            "Despite that, it behaves correctly (see https://developer.mozilla.org/en-US/docs/Web/CSS/border-width)." +
+            "Therefore this function still can be used without specifying parameter names. Replacement is not necessary." +
```
igordmn, line 134: Is [JsName](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.js/-js-name/) annotation applicable to arguments? Docs says that it is not, but not sure how it is in reality.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/border.kt
```diff
@@ -83,26 +83,115 @@ fun StyleScope.borderWidth(width: CSSNumeric) {
...
+    borderWidth3(topLeft, topRightAndBottomLeft, bottomRight)
+}
+
+/**
+ * This function is a temporary replacement for `borderWidth(topLeft: CSSNumeric, topRightAndBottomLeft: CSSNumeric, bottomRight: CSSNumeric)`.
+ * After a fix for parameters names in `borderWidth`,
+ * this function will be Deprecated with a replaceWith `borderWidth`. Later `borderWidth3` will be Removed.
+ * @see borderWidth
+ */
+fun StyleScope.borderWidth3(top: CSSNumeric, horizontal: CSSNumeric, bottom: CSSNumeric) {
```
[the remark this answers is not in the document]
eymar: I decided to add 3 new functions: 
borderWidth2 borderWidth3 and borderWidth 4 to note the number of parameters.
I didn't want to add one new function and overload it for different parameters count, because I didn't see a good function name for it. So the number just notes the number of parameters

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/border.kt
eymar: Even if it was applicable, the wrong parameters name is still in our source code written in kotlin. 
I can't see how it could help

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/border.kt
```diff
@@ -83,26 +83,115 @@ fun StyleScope.borderWidth(width: CSSNumeric) {
...
 }
 
+
+/**
+ * https://developer.mozilla.org/en-US/docs/Web/CSS/border-width
+ *
+ * First argument sets the vertical borders width.
+ * Second argument sets the horizontal borders width.
+ *
+ * NOTE: Temporary deprecation until 2.0 because of wrong parameters names.
```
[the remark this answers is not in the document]
eymar: I see. I think it's also okay. 
So in 1.4 the parameters names will be changed.

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/border.kt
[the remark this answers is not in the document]
igordmn: But when we undeprecate the original function, it will be the single function, without suffix. So maybe better to make API closer to the new API?

2/3/4 as suffix for number of arguments isn't obvious :)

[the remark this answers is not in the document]
eymar: Okay. Then I think the number suffix won't make any sense at all.
Maybe borderWidthTemp? Meaning temporary?

web/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/properties/border.kt
igordmn: We could write the right names in Kotlin, and generate JS with the old names. If there is a possibility.

igordmn: I have an idea. We can:

- keep the old function without deprecation, rename arguments, and add `@JsName("borderWidth2")` to it
- add new function `borderWidth2` with the old arguments, add `@JsName("borderWidth"), and deprecate it.

igordmn: That would work in JVM, but not sure how it will be in JS

eymar: > keep the old function without deprecation, rename arguments, and add @JsName("borderWidth2") to it

But the kotlin code won't compile if they have call with explicit old names :

borderWidth(topLeft = 10.px, bottomRight = 20.px)

This function won't have parameters with such names (the new names are horizontal and vertical)

igordmn: Seems it won't work, because klib doesn't contain names defined in JsName.

igordmn: Will this work? I added `unused` parameter to the old function
```
@Deprecated("")
fun borderWidth(left: Int, right: Int, unused: Int = 0) {
    
}

fun borderWidth(left: Int, rightTop: Int) {
    
}

fun main() {
    borderWidth(23, 53)
    borderWidth(left = 23, right = 54)
}
```

eymar: Great and creative idea! It seems to work. Hopefully no one would use `unused` parameter.

So in deprecation message we suggest to add the parameters names explicitly? Because with old parameter names or without names at all, the deprecated method would be used.

Please look at the last commit with your idea applied.

Thank you!

igordmn: > Because with old parameter names or without names at all, the deprecated method would be used.

Hmm, on JVM it uses the new one if we omit all parameters. Is that not true in your case?

In my head it was that no one will see the deprecation message. Only the old compiled libraries will use the deprecated method. And the new ones, will use the new method.

If that is not true, then in the deprecation message we can tell users to define parameters explicitly.

eymar: > Is that not true in your case?
Yes, it uses the old deprecated method if no parameters names specified.

+ Updated the deprecation message

igordmn: In this code, it uses the new one
image

eymar: I got it wrong yesterday. I added `unused` parameter to a new (not deprecated) method. I'll fix now.

eymar: Thank you! really nice idea :) 

done

igordmn (approved):

PR #2303, opened 2022-09-09
Documentation for how to package apps using Conveyor.

mikehearn:
[Hydraulic Conveyor](https://hydraulic.software/) is a packaging tool.

- Reorganize the pages on packaging, signing and notarization. 
    - The original packaging page is now a comparison to help people pick between Conveyor and `jpackage`. 
    - The `jpackage` specific docs moved to a new page.
    - The signing docs moved to be next to the `jpackage` docs.
- Add a brief tutorial on adapting a project to use Conveyor.


tutorials/Native_distributions_and_local_execution/conveyor.md
```diff
@@ -0,0 +1,187 @@
...
+4. The `app.dir` system property is set to the absolute install path, so you can use it to load packaged files you added using the `app.inputs` key.
+5. The `app.version` system property is wired to your `app.version` config key, which is set to the value of your Gradle `version` property.
+6. Make a mistake in your packaging? No big deal - use `app.revision` to make a new version of the app that doesn't affect the version numbers displayed to end users.
+7. Explicit modules (those with a valid `module-info.class` file) are automatically linked into the JVM `modules` file, which is a file format optimized for smaller size and faster startup. They won't be available as JARs on disk as a consequence.
+8. Conveyor will add the Java Accessibility Bridge to your application automatically, to ensure blind users can access the UI.
+9. You can include [a bundled config snippet to improve your JVM configuration in various ways](https://conveyor.hydraulic.dev/latest/stdlib/jvm-clients/).
+10. You can use ProGuard or R8 to obfuscate your app without problems, however Conveyor doesn't automate this. You'll need to integrate it into your build.
```
igordmn, line 185: We [recently](https://github.com/JetBrains/compose-jb/pull/2313/files) added this support to our Gradle plugin. We will check if it works with Conveyor, when there will be new compatible versions of Conveyor and/or of Compose 1.2.x.

Also, we will put a link to the ProGuard tutorial, when we write it.

mikehearn:
Thanks for the review! 

There are several ideas for improvement here. I think the way to go here is to keep this PR open for a while whilst we work through the issues, and try to merge whatever we have before the Compose 1.2 release. 

Here's a task list that I'll update as we complete tasks.

* [ ] Look at moving the tutorial into the Conveyor tutorial. (see below for thoughts)
* [ ] Fix the Conveyor plugin to understand the newer version of the Compose JVM Gradle plugin. It is probably not public API we're using so it's on us to fix things.
* [ ] Figure out the best way to integrate with the KMP plugin. (see below for thoughts)

## KMP plugin

I filed [KT-54094](https://youtrack.jetbrains.com/issue/KT-54094/KMP-Gradle-plugin-should-have-machine-specific-JVM-targets) to request support for machine-specific JVM dependency configurations, because that's where the difficulty lies. Both Conveyor and KMP have their own notion of how to express machine-specific dependency graphs, and Conveyor has to "patch up" the current runtime configuration so the app remains runnable after the Compose dependencies were separated out into machine specific configs. 

I noticed that some sample apps linked from the KMP docs actually use Compose Desktop with the JVM plugin. I previously thought a project had to be all KMP or no KMP but it seems you can mix and match them. This should work with the current support. Still, we should still support projects/modules that are using the KMP plugin only.

## Tutorial

As you mentioned on Slack, it may be surprising to have so much material in the Compose docs. The reason for this was just trying to 'beat the clock' for the Compose 1.2 release, because the regular Conveyor tutorial is starting to reach the limits of its current organization. There are quite a few "dimensions" e.g. JVM/Electron/native, plus JVM-JavaFX/JVM-Compose, and there should be even more (template project, adapt existing, compose-jvm-plugin, compose-kmm-plugin). To add more dimensions without it becoming confusing we'll need to do a significant re-org of the tutorial. The embedded tutorial here benefits from context: it's only Compose, it's only adapting an existing project, and - maybe - it'll only be for KMP plugin once that works.

Nonetheless, maybe it's time to bite the bullet and find a better organization so the main tutorial can grow. I was thinking about this over the weekend and have some ideas. We will experiment with this and update the PR to remove the tutorial if we are happy with the results.

tutorials/Native_distributions_and_local_execution/README.md
```diff
@@ -5,11 +5,18 @@
 In this tutorial, we'll show you how to create native distributions (installers/packages) for all the supported systems.
 We will also demonstrate how to run an application locally with the same settings as for distributions.
 
-## Gradle plugin
+## Available tools
+
+There are two tools available for packaging desktop applications:
+
+1. [jpackage](https://openjdk.java.net/jeps/343), which comes with the JDK.
```
AlexeyTsvetkov, line 12: We consider `jpackage` an implementation detail. Also we already have some code, that does not use `jpackage` at all. I suggest to rephrase to something like: "Compose Multiplatform Gradle Plugin provides basic packaging, obfuscation (via ProGuard), and signing (currently, macOS only) out-of-the-box."

tutorials/Native_distributions_and_local_execution/jpackage-conveyor-comparison.md
```diff
@@ -0,0 +1,25 @@
+# Packaging tools comparison 
+
+Use [Conveyor](https://www.hydraulic.software) if you want any of the following features:
+
+* Online updates. jpackage creates packages that users must update manually by reinstalling.
+* Creation, signing and notarization of packages for all supported targets from any OS (e.g. your developer laptop, a CI machine). jpackage must be run on each platform you wish to support.
+* Self-signed packages with `curl | bash` style installs, which don't require you to purchase signing certificates.
+* Better support for JNI native libraries: 
+  * jpackage doesn't know how to extract or sign native libraries bundled into library JARs, which can cause macOS notarization failures.
```
AlexeyTsvetkov, line 9: Our plugin signs native libraries just fine. Again, I would prefer "Compose Multiplatform [Gradle] plugin" or something like that instead of `jpackage`.

tutorials/Native_distributions_and_local_execution/jpackage-conveyor-comparison.md
```diff
@@ -0,0 +1,25 @@
+# Packaging tools comparison 
+
+Use [Conveyor](https://www.hydraulic.software) if you want any of the following features:
```
AlexeyTsvetkov, line 3: Maybe it's just me, but the comparison seems to favour Conveyor over the default plugin.
While I do think:
* that listing alternative options is useful for current and future users;
* Conveyor does provide a lot more features, and likely will be for the foreseeable future; 
it seems reasonable to describe what is provided out-of-the-box first,
and then describing, what Conveyor or other third-party tools improve. 

What do you think about restructuring this section to something like:
```
Compose Multiplatform Gradle plugin covers basic packaging options:
...

Consider using [Conveyor](https://www.hydraulic.software) (a third party tool developed by Hydraulic Software), if you want any of the following features:
...  
```

AlexeyTsvetkov (changes requested):
I would prefer to have a few changes, but overall the PR looks good to me.

mikehearn:
@AlexeyTsvetkov OK, comments all addressed. Please take another look when convenient.

We've also been busy with addressing the other issues. The [Conveyor 3 update we released today](https://www.hydraulic.software/blog/4-september-updates.html) improves Compose support in the following ways:

* New Conveyor Gradle plugin that syncs with the Compose 1.2 Gradle plugin and supports Kotlin/Multiplatform plugin.
* Built-in template/sample apps are now using KMP.
* Imports JVM toolchain settings from the Gradle Java plugin if you set it and the requested JVM is recognized.

We re-organized the tutorial so it's easier to follow a Compose-specific path and adapt existing apps, not just package the generated sample apps. Thanks to @igordmn for that suggestion.

It also added support for a few other things like being able to start automatically on Windows login, which is good for people using Compose Desktop for embedded/kiosk apps. That was a request spotted in the #compose-desktop slack.

I did a quick check and of the last 25 questions in the Compose Desktop slack, 8 of them are related to packaging/distribution problems in some way or around 1/3rd of all problems. So getting better packaging tools to users is going to hopefully be of great benefit to the Compose Desktop community! :)

tutorials/Native_distributions_and_local_execution/jpackage-conveyor-comparison.md
mikehearn: OK, thanks. I've removed this line from the PR already. I wonder what was going wrong for this other user (they posted about it in slack).

AlexeyTsvetkov (approved):

PR #2306, opened 2022-09-11
Upgrade the vlcj and Add a new Compose way to integrate with vlcj

succlz123:
## Changes
1. Upgrade the vlcj to fix the Apple Silicon JNA Error https://github.com/caprica/vlcj/issues/1145 https://github.com/caprica/vlcj/issues/1146
2. Add a new Compose way to integrate with vlcj

## Screenshot
image

Integrating the vlcj in the new way allows us to prevent the Compose Content from being overlaid by SwingPanel.


kropp:

PR #2307, opened 2022-09-12
Add NO_SKIKO platforms set in ComposePlatforms

eymar:
They can be used to build and publish compose-runtime. Other compose modules can't be built for NO_SKIKO since those platforms are not supported by skiko yet.

Related: https://github.com/JetBrains/androidx/pull/292 (adding more targets for compose-runtime in androidx fork)

### DIFF
--- a/compose/buildSrc/src/main/kotlin/ComposePlatforms.kt
+++ b/compose/buildSrc/src/main/kotlin/ComposePlatforms.kt
@@ -15,7 +15,18 @@ enum class ComposePlatforms(vararg val alternativeNames: String) {
     MacosArm64("Macos"),
     UikitX64("UiKit"),
     UikitArm64("UiKit"),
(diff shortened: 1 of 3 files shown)


igordmn (approved):

compose/buildSrc/src/main/kotlin/ComposePlatforms.kt
```diff
@@ -34,6 +45,20 @@ enum class ComposePlatforms(vararg val alternativeNames: String) {
             ComposePlatforms.AndroidRelease
         )
 
+        // Thses platforms are not supported by skiko yet
```
dima-avdeev-jb, line 48:
```suggestion
        // These platforms are not supported by skiko yet
```

dima-avdeev-jb (approved):

Issue #2304, opened 2022-09-10
Web: Make AttrBuilderContext Composable


hfhbd:
https://github.com/JetBrains/compose-jb/blob/471d5c652db20286d478c25c2b3d80ab4200215a/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Base.kt#L21 It is already Composable, but this feature/annotation is no forward to other calls, namely: https://github.com/JetBrains/compose-jb/blob/471d5c652db20286d478c25c2b3d80ab4200215a/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Base.kt#L109

Use case:
Use `remember` in the styling api to create ids automatically.

eymar:
Could you please add a snippet how you would use `remember` in style or attrs blocks?

hfhbd:
@eymar What the test of the PR? :D https://github.com/JetBrains/compose-jb/pull/2308/files#diff-96fe1b0a96b139b5b9b4fd22973cc6e48a52b8eb3fd477296bc29cb40b7b85b8

hfhbd:
I did try to make `style` composable too, but the non composable test `attrsBuilderCopyFromPreservesExistingAttrs` failed (or my setup is somehow broken...)

eymar:
```kotlin
Div({
                val color by remember {
                    derivedStateOf {
                        if (colorFlip) {
                            Color.green
                        } else Color.red
                    }
                }
                style {
                    backgroundColor(color)
                }
            })
        }
```

Perhaps, I'm missing something in this snippet. As I understand it now, we can move `val color by remember ...` above the Div, and use `color` in `style {}` and it should work. 
Maybe this test is just to validate that composable attrs block works. 
So maybe, your use-case `Use remember in the styling api to create ids automatically.` could help me understand your goal better. How would remember help create ids automatically?

hfhbd:
@eymar Thanks for the feedback! The goal is to support `display-grid` (and some other css features): https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout
In the sample of MDN they use classes and bootstrap-compose generates these class names automatically with a UUID. So we use this:
```kotlin
Style {
    val classname = remember { "_${UUID()}" }

    className(classname) style {
        // apply the styling
    }
}
return classname
```
and the same class name in the actually html element.

[...]

PR #2308, opened 2022-09-12
Web: Make AttrBuilderContext @Composable

hfhbd:
Fixes https://github.com/JetBrains/compose-jb/issues/2304

Draft because I get an error running the tests locally:
```
Execution failed for task ':web-core:allTests'.
> Failed to execute all tests:
  :web-core:jsBrowserTest: java.lang.IllegalStateException: command '/Users/philipwedemann/.gradle/nodejs/node-v16.13.1-darwin-arm64/bin/node' exited with errors (exit code: 1)
```

### DIFF
--- a/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Base.kt
+++ b/web/core/src/jsMain/kotlin/org/jetbrains/compose/web/elements/Base.kt
@@ -106,7 +106,7 @@ private class DomElementWrapper(override val node: Element): DomNodeWrapper(node
 @Composable
 fun <TElement : Element> TagElement(
     elementBuilder: ElementBuilder<TElement>,
-    applyAttrs: (AttrsScope<TElement>.() -> Unit)?,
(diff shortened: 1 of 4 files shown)


eymar:
Some tests failed because of 10 seconds timeout.
So I tried to play with it in Sample.kt (web-integration-core module, fun main). 

It seems that, for some reason attrs block doesn't recompose:

```
renderComposableInBody {
        var flip by remember { mutableStateOf(false) }

        Div(attrs = {
            id(if (flip) "red" else "green")
            style {
                backgroundColor(if (flip) Color.red else Color.green)
            }
        }) {
            Text("TEST, flip = $flip")
        }

        Button(
            attrs = {
                onClick { flip = !flip }
            }
        ) { Text("Flip") }
    }
    ```
    The text in Div changes when button is clicked, but the id and backgroundColor doesn't change. 
    
    No clues for now, why it behaves like this.

hfhbd:
Thanks for looking into it. 
Maybe this is the cause for the unsupported `@Composable AttrsScope`? :D

I am okay with keeping the current behavior.

PR #2315, opened 2022-09-13
Tutorial for context menu interop with Swing

igordmn:


eymar:
Maybe it's worth to describe how to create a JPopupTextMenu with custom list of items?

igordmn:
> Maybe it's worth to describe how to create a JPopupTextMenu with custom list of items?

Done

tutorials/Context_Menu/README.md
```diff
@@ -124,4 +122,261 @@ fun main() = singleWindowApplication(title = "Context menu") {
...
+                        swingItem(localization.paste, Color.BLUE, KeyEvent.VK_V, it)
+                    )
+                }
+                textManager.selectAll?.also {
+                    add(JPopupMenu.Separator())
+                    add(
+                       swingItem(localization.selectAll, Color.BLACK, KeyEvent.VK_A, it)
+                    )
+                }
+                for (item in items) {
```
eymar, line 343: The `items` is received here from somewhere else.  As I understand, `items` is defined in `JPopupTextMenu`:
```kotlin
ContextMenuArea(items = { emptyList() }, state, content = content)
```
So it's hardcoded to be empty and  when using JPopupTextMenu directly, we can't change it. Maybe using `items` in this example could confuse a bit, since we can't change it? 

Maybe it's a question to the PR in androidx, but anyway:
Does `JPopupTextMenu` need List<ContextMenuItem>  in `private val createMenu: (TextManager, List<ContextMenuItem>) -> JPopupMenu` if the list is always empty?

Do I miss something?

igordmn: This tutorial is mostly about how to change appearance for all text context menus in the application, not only in some part of it. And when we change it on the global level, implementation of TextContextMenu should cover the case, when the user pass custom items down to all context menus:

```
         ContextMenuDataProvider(
            items = {
                listOf(
                    ContextMenuItem("User-defined Action") {/*do something here*/ },
                    ContextMenuItem("Another user-defined action") {/*do something else*/ }
                )
            }
        ) {
            TextField(
                value = text.value,
                onValueChange = { text.value = it },
                label = { Text(text = "Input") }
            )

            Spacer(Modifier.height(16.dp))

            SelectionContainer {
                Text("Hello World!")
            }
        }
```
Even if we pass emptyList in `JPopupTextMenu`, it will be merged with the other items, passed via ContextMenuDataProvider

eymar: I see. Thanks

eymar (approved):

PR #2318, opened 2022-09-15
Fix idea plugin compatability

igordmn:
IDEA doesn't provide backward compatibility policy, and breaking changes can happen in any version, without prior deprecation:

https://plugins.jetbrains.com/docs/intellij/api-changes-list.html

~~Because of that, we can't remove `plugin.until.build`, and have to build Compose plugin for each new version of IDEA separately~~ (see Update)

~~Update 1:
I removed plugin.until.build anyway. In most cases there won't be a broken plugin. And if it will be broken, we will know it during IDEA EAP. Considering this, better when users don't wait the new version, and use already available one.~~

Update 2:
If we omit plugin.until.build, then it has default value +8. Returned, but made the default value big enough.

What we should do after that commit, is make sure, that we don't forget to check this compatibility. And ship a new stable version for every new stable IDEA.

The ideal way is to run `runPluginVerifier` on CI for the latest version, which will fail the build, if our plugin isn't compatible.

But for some reason, there is no EAP for IDEA 2022.3 here:
https://jb.gg/intellij-platform-builds-list

### DIFF
--- a/idea-plugin/build.gradle.kts
+++ b/idea-plugin/build.gradle.kts
@@ -57,7 +57,6 @@ tasks {
 
     patchPluginXml {
         sinceBuild.set(projectProperties.pluginSinceBuild)
-        untilBuild.set(projectProperties.pluginUntilBuild)
     }
 
     runPluginVerifier {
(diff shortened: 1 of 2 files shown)


igordmn:
We also need to release 1.2.2, because 1.2.1 is only compatible with IDEA 2022.1

igordmn:
It seems if we omit "until", it has a default value +8: https://plugins.jetbrains.com/plugin/16541-compose-multiplatform-ide-support/versions/Snapshots/223890

YazaWainakh (approved):

Issue #1972, opened 2022-03-22
Support for custom WXS entries during MSI packaging


dmvanderuit:
Hi,

I've got a question regarding the packaging of the application. In the [section on native distributions](https://github.com/JetBrains/compose-jb/tree/master/tutorials/Native_distributions_and_local_execution), the docs mention that the packaging uses JPackage under the hood. In JPackage, it's possible to [further customise the installer using WXS files](https://stackoverflow.com/a/71050276). Is this something that is supported in the packaging of MSI packages run by the Gradle task in Compose too?

Thanks in advance.

sebkur:
I think this might already be possible. In [AbstractJPackageTask.kt](https://github.com/JetBrains/compose-jb/blob/master/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt) we already have

```kotlin
if (targetFormat != TargetFormat.AppImage) {
    …
    cliArg("--resource-dir", jpackageResources)
    …
}
```

where `jpackageResources` is defined here:

```kotlin
@get:LocalState
protected val jpackageResources: Provider<Directory> = project.layout.buildDirectory.dir("compose/tmp/resources")
```

so apparently `compose/tmp/resources` is passed to `jpackage` via `--resource-dir` and files to modify the input for Wix creating the `msi` such as a modified `main.wxs` could be copied there.

sebkur:
ah, well, OK, but then there's also `cleanDirs(jpackageResources)` executed during `prepareWorkingDir`, so essentially anything we copy to `compose/tmp/resources` to modify the input data for Wix will be deleted before running `jpackage`.

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-1972">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #2331, opened 2022-09-20
Add support for adding custom resources for jpackage

sebkur:
I need to customize the WXS entries for packaging MSI files on Windows as described in #1972. I have added a solution here that works similar to the `appResourceRootDir` option: `jpackageResourceRootDir`. I'm not sure which resources can be added for Linux or macOS and am currently only aware of the customizations for Windows using Wix, however I presume there are things that can be customized for those operating systems as well.


sebkur:
Just noticed there is #2266 already, but it's not as far developed as this I think.

krizzu:
I think this needs more attention, who could we ping to have a look at this?

PR #2366, opened 2022-10-06
Allow to use Compose on multiple Kotlin versions

igordmn:
JS target supports a lower version (1.7.10), because we have a bug in Koltin 1.7.20

Compose 1.2.0 will support:
- 1.7.20 and 1.7.10 for Android and Desktop
- 1.7.10 for JS

We will release the new patchset (1.2.1) with 1.7.2X support for JS later


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatability.kt
```diff
@@ -0,0 +1,14 @@
+package org.jetbrains.compose
+
+internal object ComposeCompilerCompatability {
+    fun compilerVersionFor(kotlinVersion: String): ComposeCompilerVersion? = when (kotlinVersion) {
+        "1.7.10" -> ComposeCompilerVersion("1.3.0-alpha01")
+        "1.7.20" -> ComposeCompilerVersion("1.3.2-alpha01", isJsSupported = false)
```
dima-avdeev-jb, line 6: 1.3.0-alpha01 and 1.3.2-alpha01
Maybe we can you stable versions?
https://mvnrepository.com/artifact/androidx.compose.compiler/compiler

gradle-plugins/gradle.properties
```diff
@@ -1,13 +1,16 @@
...
 
-# __LATEST_COMPOSE_RELEASE_VERSION__
 # Default version of Compose Libraries used by Gradle plugin
 compose.version=1.2.0-beta02
-# Default version of Compose Compiler used by Gradle plugin
-compose.compiler.version=1.3.2-alpha01
-# Default version of Kotlin compatible with compose.compiler.version
-compose.compiler.compatible.kotlin.version=1.7.20
+# The latest version of Compose Compiler used by Gradle plugin. Used only in tests.
+compose.tests.compiler.version=1.3.2-alpha01
```
dima-avdeev-jb, line 7: maybe we can you stable version? 1.3.2

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatability.kt
igordmn: Yes, we will release stable versions and use them here, when we will be sure that everything works.

dima-avdeev-jb (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/ComposeCompilerArtifactProvider.kt
```diff
@@ -5,33 +5,69 @@
...
+
+    init {
+        val customPlugin = customPluginString()
+        val customCoordinates = customPlugin?.split(":")
+        when (customCoordinates?.size) {
+            null -> {
+                val version = requireNotNull(
+                    ComposeCompilerCompatability.compilerVersionFor(kotlinVersion)
+                ) {
+                    "This version of Compose Multiplatform doesn't support Kotlin " +
```
AlexeyTsvetkov, line 43: The message is a bit confusing. As I understand it, here we simply don't know, what version of Compose Compiler to choose. Instead the message says "this version of Compose doesn't support Kotlin $kotlinVersion", which may or may not be true. For example, if there is a version of a Compose Compiler, that works with current plugin and the desired version of Kotlin, then in my understanding the Gradle plugin & libraries are compatible, we simply don't know the right version of the compiler.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -52,6 +57,7 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
 
     override fun applyToCompilation(kotlinCompilation: KotlinCompilation<*>): Provider<List<SubpluginOption>> {
         val target = kotlinCompilation.target
+        composeCompilerArtifactProvider.checkTargetSupported(target)
```
AlexeyTsvetkov, line 60: The overall code design became confusing.
Previously, `ComposeCompilerArtifactProvider` had a single responsibility of providing a version of `ComposeCompilerArtifact`.  Now `ComposeCompilerArtifactProvider` also knows how to check if a target is compatible with a Compose Compiler, but only if the default compiler is used. 
So the following configuration 
```
plugins {
  id("org.jetbrains.kotlin.multiplatform") version "1.7.20"
  id("org.jetbrains.compose") version "1.2.0"
}

kotlin {
   js(IR) {
        browser()
        binaries.executable()
    }
}

compose {
     kotlinCompilerPlugin.set("1.3.2-alpha01")
     
     web {}
}
```
will not trigger the check at all, despite that we know they are incompatible.

I suggest to separate *compiler inference* and *compiler compatibility*:
1. The inference should only try to choose a version. If inference fails, it should tell user to try updating the Compose Compiler (if a compatible version exists).
2. Then we run compiler compatibility checks, regardless of whether the version was inferred or set manually.

AlexeyTsvetkov (approved):
All issues are not critical, but ideally we should look at them

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/ComposeCompilerArtifactProvider.kt
```diff
@@ -5,33 +5,69 @@
...
 
-import org.jetbrains.compose.ComposeBuildConfig
+import org.jetbrains.compose.ComposeCompilerCompatability
+import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
+import org.jetbrains.kotlin.gradle.plugin.KotlinTarget
 import org.jetbrains.kotlin.gradle.plugin.SubpluginArtifact
 
-internal class ComposeCompilerArtifactProvider(private val customPluginString: () -> String?) {
+private const val KOTLIN_COMPATABILITY_LINK =
+    "https://github.com/JetBrains/compose-jb/blob/master/VERSIONING.md#kotlin-compatibility"
```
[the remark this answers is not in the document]
igordmn: How can I create such link?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/ComposeCompilerArtifactProvider.kt
igordmn: Also, there is an option to completely prohibit using Kotlin 1.7.20, even if the user sets compatible Compose Compiler. But I don't think it is a good idea.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
igordmn: > I suggest to separate compiler inference and compiler compatibility:

See my comment in the other discussion. We can't make compatibility check for an arbitrary compiler in the Gradle plugin. We only can make such check, if the compiler was chosen from our list.

As for code quality, we expand responsibility of `ComposeCompilerArtifactProvider` from `providing a version of ComposeCompilerArtifact` to `providing info about compatible compiler`. It still can be considered as a single responsibility.

PR #2367, opened 2022-10-06
Relationship between the Jetpack Compose and Compose Multiplatform release cycles

igordmn:
There were a lot of question to our release cycle and a lot of confusion with Jetpack Compose. This section answers on most popular questions.

Also add documentation how to use Jetpack Compose Compiler.


VERSIONING.md
```diff
@@ -33,3 +33,35 @@ Compose version | Kotlin version
...
+When you build your application for Android, the artifacts published by Google are used.
+
+In the next table you can see versions of the corresponding Jetpack Compose libraries.
+These are the versions on which we based Compose Multiplatform, and the versions used when you build your application for Android.
+
+Compose Multiplatform | Jetpack Compose Compiler | Jetpack Compose Runtime, UI, Foundation, Material | Jetpack Compose Material 3
+--- |--------------------|---------------------------------------------------| ---
+1.0.0 | 1.1.0-rc03 | 1.1.0-rc03                                        | 1.0.0-alpha03
+1.1.1 | 1.1.0 | 1.1.0                                             | 1.0.0-alpha05
+1.2.0 | 1.3.2 | 1.2.1                                             | 1.0.0-alpha14
```
eymar, line 54: Do we plan to add new rows to this table for every release? 

I guess the answer is yes. Would be good to not forget about it. Maybe some checklist for a release could help.

eymar (approved):

VERSIONING.md
```diff
@@ -33,3 +33,35 @@ Compose version | Kotlin version
 1.0.0 | 1.5.31
 1.1.1 | 1.6.20
 1.2.0 | 1.7.10 - 1.7.20 (Desktop, Android), 1.7.10 (JS)
+
+### Relation to Jetpack Compose and Compose Multiplatform release cycle
+
+Compose Multiplatform shares a lot of code with [Jetpack Compose](https://developer.android.com/jetpack/compose) for Android developed by Google.
+We keep our release cycles aligned, so the common part is properly tested and stabilized.
+
+When the new minor version of Jetpack Compose is released, we pick the release commit, use it as a base for [Compose Multiplatform fork](https://github.com/JetBrains/androidx), finish new platform features, stabilize all platforms, and release Compose Multiplatform after some time.
```
jimgoog, line 42: Let's avoid the term "fork", as it can sometimes be interpreted as implying that it is creating a competing/hostile product.  Instead of "Compose Multiplatform fork", perhaps consider "Compose Multiplatform build" or even just "Compose Multiplatform"

jimgoog: Or... "use it as a base for the next Compose Multiplatform release" or "use it as a base for the next Compose Multiplatform version"

VERSIONING.md
jimgoog: Unclear why we are maintaining this table.  What is a user supposed to do with this table?

VERSIONING.md
```diff
@@ -33,3 +33,35 @@ Compose version | Kotlin version
...
+In the next table you can see versions of the corresponding Jetpack Compose libraries.
+These are the versions on which we based Compose Multiplatform, and the versions used when you build your application for Android.
+
+Compose Multiplatform | Jetpack Compose Compiler | Jetpack Compose Runtime, UI, Foundation, Material | Jetpack Compose Material 3
+--- |--------------------|---------------------------------------------------| ---
+1.0.0 | 1.1.0-rc03 | 1.1.0-rc03                                        | 1.0.0-alpha03
+1.1.1 | 1.1.0 | 1.1.0                                             | 1.0.0-alpha05
+1.2.0 | 1.3.2 | 1.2.1                                             | 1.0.0-alpha14
+
+You may want to use [a cutting edge Compose Compiler](https://developer.android.com/jetpack/androidx/releases/compose-kotlin) published by Google for your Compose Multiplatform project.
```
jimgoog, line 56: Consider adding: "The Compose Compiler version can be changed independently of other Compose libraries.  In order to support newer versions of Kotlin, " to the beginning of the sentence.

So this becomes:
The Compose Compiler version can be changed independently of other Compose libraries.  In order to support newer versions of Kotlin, you may want to use [a cutting edge Compose Compiler](https://developer.android.com/jetpack/androidx/releases/compose-kotlin) published by Google for your Compose Multiplatform project.

VERSIONING.md
```diff
@@ -33,3 +33,35 @@ Compose version | Kotlin version
...
+For example, when the new Kotlin is released,  and there is the Jetpack Compose Compiler that supports it,
+and Compose Multiplatform still doesn't support it.
```
jimgoog, line 58: The current wording here may be confusing to users.  First you tell them about Compose Compilers, and then you tell them that Compose Multiplatform doesn't support it.  It makes sense when the reader already knows what you're trying to say, but isn't necessarily clear to users who are seeing this for the first time.

Instead consider something like:

> For example, when the new Kotlin is released, a corresponding Compose Multiplatform release may not yet have been published, but manually specifying a newer Compose Compiler version can allow you to build your Compose Multiplatform app using the latest Kotlin release.

VERSIONING.md
igordmn: > Do we plan to add new rows to this table for every release?

Yes, we should update this table for every stable release.

> Maybe some checklist for a release could help.

Good idea, I will make such checklist.

> What is a user supposed to do with this table?

Knowing version of Jetpack Compose is useful in some cases:
- to know the current issues, and when they are/will be fixed
- to watch Jetpack Compose changelog
- to investigate compatibility issues (for example, if users use some Android library that depend on the other version of Jetpack Compose)

There were a few questions in Slack/GitHub about on which version we were based.

VERSIONING.md
igordmn: "use it as a base for the next Compose Multiplatform version" looks better, thanks!

VERSIONING.md
igordmn: Changed.

VERSIONING.md
igordmn: Your variant is better, changed.

VERSIONING.md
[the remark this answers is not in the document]
igordmn: > It's probably "a new minor"

Fixed

> release after each minor release of Jetpack Compose

https://semver.org describes Major as the first number of the version, and Minor as the second. But to avoid confusion, I removed "major".

> Overall could be rephrased to something like:

My points are mostly subjective, but:
- "Compose Multiplatform shares a lot of code with Jetpack Compose" is better than "Compose Multiplatform is built on top of Jetpack Compose"
- I would like to not mention upstreaming process. Upstreaming has two parts - upstreaming common features and upstreaming desktop features. First may be relevant for users, but we need to think more how to properly describe it, without adding complexity, and the second probably isn't much relevant, it just our "implementation details"

VERSIONING.md
```diff
@@ -33,3 +33,35 @@ Compose version | Kotlin version
...
 1.1.1 | 1.6.20
 1.2.0 | 1.7.10 - 1.7.20 (Desktop, Android), 1.7.10 (JS)
+
+### Relation to Jetpack Compose and Compose Multiplatform release cycle
+
+Compose Multiplatform shares a lot of code with [Jetpack Compose](https://developer.android.com/jetpack/compose) for Android developed by Google.
+We keep our release cycles aligned, so the common part is properly tested and stabilized.
+
+When the new minor version of Jetpack Compose is released, we pick the release commit, use it as a base for [Compose Multiplatform fork](https://github.com/JetBrains/androidx), finish new platform features, stabilize all platforms, and release Compose Multiplatform after some time.
+The gap between the Compose Multiplatform release and the Jetpack Compose release is usually 1-3 months.
```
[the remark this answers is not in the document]
igordmn: Fixed

VERSIONING.md
```diff
@@ -33,3 +33,35 @@ Compose version | Kotlin version
...
+
+When you build your application for Android, the artifacts published by Google are used.
+
+In the next table you can see versions of the corresponding Jetpack Compose libraries.
```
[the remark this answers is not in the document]
igordmn: Changed, thanks!

VERSIONING.md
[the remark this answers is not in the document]
igordmn: Fixed

VERSIONING.md
```diff
@@ -33,3 +33,35 @@ Compose version | Kotlin version
...
+
+You may want to use [a cutting edge Compose Compiler](https://developer.android.com/jetpack/androidx/releases/compose-kotlin) published by Google for your Compose Multiplatform project.
+For example, when the new Kotlin is released,  and there is the Jetpack Compose Compiler that supports it,
+and Compose Multiplatform still doesn't support it.
+
+In that case you can set `kotlinCompilerPlugin` in the `compose` section of your `build.gradle.kts` file:
+
+```kotlin
+compose {
+    kotlinCompilerPlugin.set("androidx.compose.compiler:compiler:1.3.1")
```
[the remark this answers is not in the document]
igordmn: If we release the new version of plugin every time when we release a new version of Compose Compiler, then (considering that we also support multiple version of Kotlin now) this information won't be relevant to majority of users. There won't a case when they needed to overwrite it separately. Upgrading to a newer patch of Compose Gradle Plugin (1.2.0 -> 1.2.1) is similar to upgrading Compose Compiler.

Also, if we describe it, we should also describe differences in our Compose Compiler and Jetpack Compose Compiler, and make a separate list of all our versions of Compiler. The Jetpack Compose users will think "by default" that they should pick  a version from [here](https://developer.android.com/jetpack/androidx/releases/compose-kotlin), which is not true.

This feature I believe is useful in these cases:
- asking users to use a different compiler to investigate an issue
- we released a new compiler, but still haven't released a new plugin (we have an error on CI, for example)
- users don't want to upgrade to major version of Compose (we released 1.3.0 with Kotlin 1.7.30 support, but didn't release 1.2.3)
- we realize that our current approach has flaws, or users actively use it instead of our current approach

So, maybe we will see how it works in battle, suggesting this "hidden" feature in issues, discussions and Slack before we describe it as another way to support new Kotlin? Having two ways (one - upgrading Compose Plugin, two - overriding a separate field) complicates things, and we have to be sure that another way adds enough value. 

On the other hand, using third-party (Jetpack) compiler is more useful feature, because Jetpack Compose Compiler is released more frequently and earlier.

VERSIONING.md
jimgoog: All of those sound like things that are almost release-notes, and most relevant when the user is upgrading.  Eg. "know the current issues and when they are/will be fixed", "watch jetpack compose changelog", "investigate compatibility issues".  I wonder if instead of a hard-to-discover table in the middle of a doc on release cycles, maybe the release notes for each release could contain a mention of which version of Compose for Android is being referenced.

igordmn: > maybe the release notes for each release could contain a mention of which version of Compose for Android is being referenced.

Good idea! I will then prepare a change in our `CHANGELOG.md` adding something like this for each version:

### Updated dependencies
This version of Compose Multiplatform is based on:
* Jetpack Compose 1.2.1
* Jetpack Compose Compiler 1.3.2
* Jetpack Compose Material3 1.0.0-alpha14

igordmn: Moved to [CHANGELOG](https://github.com/JetBrains/compose-jb/pull/2374/files)

PR #2406, opened 2022-10-14
Pack Uikit task per device configuration

blazsolar:
Remove `packComposeUikitApplicationForXCode` task as a prebuilt script of the XCode project and instead creates per device and configuration task and run it inside the same grade dependency graph. 

This makes development easier as kotlin compiling happens as part of the same grade task graph excetution, and with that, logs and progress are printed in place instead of in XCode logs. It also helps with plugin debugging because before, you had to change the plugin to pass parameters into `packComposeUikitApplicationForXCode` task.


pjBooms (approved):
Good job, Blaž! Ironically we have discussed internally to remove Gradle invocation from xcodebuild exactly at the time you provided us the PR!

blazsolar:
Great timing then 😄 I'm glad it was useful.

PR #2407, opened 2022-10-14
Fix compose gradle plugin for iOS device deployment

pjBooms:


dima-avdeev-jb (approved):
Cool, it works!
I tested it on Xcode 13 and Xcode 14.

PR #2449, opened 2022-11-05
Gradle tests additional fixes

AlexeyTsvetkov:


gradle-plugins/compose/src/test/test-projects/application/proguard/rules.pro
```diff
@@ -1,3 +1,4 @@
 -keep public class Main {
     public void keptByKeepRule(...);
-}
\ No newline at end of file
+}
+-dontnote
```
igordmn, line 4: Shouldn't we ignore just some notes, related to our libraries and Kotlin?

Some notes can be helpful

gradle-plugins/gradle.properties
```diff
@@ -4,7 +4,7 @@ kotlin.code.style=official
 # Default version of Compose Libraries used by Gradle plugin
 compose.version=1.2.0
 # The latest version of Compose Compiler used by Gradle plugin. Used only in tests.
-compose.tests.compiler.version=1.3.2
+compose.tests.compiler.version=1.3.2.1-rc02
```
igordmn, line 7:
```suggestion
compose.tests.compiler.version=1.3.2.1
```

(it is released)

gradle-plugins/compose/src/test/test-projects/application/proguard/rules.pro
igordmn: I realized that it is a test project :)

PR #2458, opened 2022-11-08
Allow the default window to go fullscreen

zeromod:
This is an odd behaviour, but I couldn't find any other solution. If `UILaunchStoryboardName` key is  present, `UIScreen.mainScreen.bounds` will have proper size instead of a smaller window. 
[Reference](https://stackoverflow.com/a/56950026)


dima-avdeev-jb:
I will check in next week!

Issue #2376, opened 2022-10-10
AbstractConfigureDesktopPreviewTask is not compatible with Gradle configuration cache


ZacSweers:
See: https://docs.gradle.org/7.5.1/userguide/configuration_cache.html

<img width="1709" alt="image" src="https://user-images.githubusercontent.com/1361086/194786887-9a1778d4-a4c4-403e-bca2-7b2739b58ea7.png">

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-2376">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #2470, opened 2022-11-14
Fix Preview compatibility with Gradle configuration cache

igordmn:
Fixes https://github.com/JetBrains/compose-jb/issues/2376

See also https://jetbrains.slack.com/archives/C015PT6ALK1/p1668109768592559


dima-avdeev-jb (approved):

AlexeyTsvetkov:
I think there should be a test with the Configuration cache enabled

igordmn:
> I think there should be a test with the Configuration cache enabled

Thanks! Added a test

PR #2515, opened 2022-11-29
Remove JVM target version override

### LINKED ISSUE #2511: jvmTarget value specified using jvmToolchain is ignored and overwritten with "1.8"
compose version=1.2.1
kotlin version=1.7.20

Specify jvmTarget value with kotlinOptions.jvmTarget parameter and compile 

```
kotlin {
    jvm {
        compilations.all {
            kotlinOptions.jvmTarget = "11"
        }
        withJava()
    }
}
```

jvmTarget value is equal to "11".

Replace jvmTarget with jvmToolchain and recompile, jvmTarget value is equal to "1.8" but not to "11"
```
kotlin {
    jvm {
        jvmToolchain(11)
        withJava()
    }
}
```

AlexeyTsvetkov:
Previously, we were setting kotlin.jvmTarget version to 1.8 if it was null or < 1.8.
As an unintended consequence we were also overriding a version set by the jvmToolchain property.
So while users expected the jvmToolchain property
to set both jdk home & jdk target, we were quietly overriding jdk target.

At the same time, Kotlin 1.7 sets the minimum
target version to 1.8 anyway, so our override
does not make sense with Kotlin 1.7+.

This commit removes overriding altogether.

Fixes #2511

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
@@ -68,17 +68,6 @@ class ComposePlugin : Plugin<Project> {
                     it.replacedBy(replacement, "org.jetbrains.compose isn't compatible with androidx.compose, because it is the same library published with different maven coordinates")
                 }
             }
-
-            val overrideDefaultJvmTarget = ComposeProperties.overrideKotlinJvmTarget(project.providers).get()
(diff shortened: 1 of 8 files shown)


dima-avdeev-jb (approved):

dima-avdeev-jb:
I checked it on sample in Issue https://github.com/JetBrains/compose-jb/issues/2511
Your PR work's fine! 
For jvmToolchain(11) result: **major version: 55**

Issue #2459, opened 2022-11-08
Kotlin 1.7.21 compatibility and adding an opt-out for compatibility checks


ZacSweers:
Something changed recently (possible 1.3.1?) where using a newer kotlin version (in this case 1.7.21) fails the build now.

```
This version of Compose Multiplatform doesn't support Kotlin 1.7.21. Please see https://github.com/JetBrains/compose-jb/blob/master/VERSIONING.md#kotlin-compatibility to know the latest supported version of Kotlin.
```

This is surprising as compose/kotlin support should be left to the compose compiler no? I'm not sure why this project is adding its own layer of checks, and it prevents us from both updating Kotlin versions and testing upcoming versions when they're available. The compose compiler plugin allows for opting out of this failure, could one be added for compose-jb as well?

igordmn:
> possible 1.3.1

This layer of checks was added in [Compose Multiplatform 1.2.0](https://github.com/JetBrains/compose-jb/releases/tag/v1.2.0), when we added [the feature of supporting multiple Kotlin's](https://github.com/JetBrains/compose-jb/pull/2366), in which our Gradle plugin chooses appropriate Compose Compiler according to the applied version of Kotlin.

Compose Compiler was always added behind the scenes in Compose Multiplatform (in compare to Jetpack Compose, where it is added explicitly). For users who knows about that there is Compose Compiler, there is a way to explicitly configure Compose Compiler. And because it is configured explicitly, we bypass this new layer of checks:

```
compose {
    kotlinCompilerPlugin.set("org.jetbrains.compose.compiler:compiler:1.3.2.1") // see versions here https://mvnrepository.com/artifact/org.jetbrains.compose.compiler/compiler
}

kotlinOptions.freeCompilerArgs += listOf(
    "-P",
    "plugin:androidx.compose.compiler.plugins.kotlin:suppressKotlinVersionCompatibilityCheck=true"
)
```

What we lack right now is proper documentation, and maybe proper DSL. We will fix that. Maybe it will be like that:
```
compose {
    kotlinCheck = false
}
```
or
```
compose {
    kotlinCompilerPlugin.set(ComposeCompilers.latest("suppressKotlinVersionCompatibilityCheck=true")
}
```
still not sure

> updating Kotlin versions and testing upcoming versions when they're available

[...]

PR #2527, opened 2022-12-02
Improve DSL for setting a custom Compose Plugin

igordmn:
Fixes https://github.com/JetBrains/compose-jb/issues/2459

Readme: https://github.com/JetBrains/compose-jb/pull/2526

1. Add `dependencies: Dependencies` extension that is accessible in `compose { }` block
2. Add `Dependencies.compiler` property that can return versions of Compose compiler used by the plugin:
```
compose {
    kotlinCompilerPlugin.set(dependencies.compiler.forKotlin("1.7.20"))
    //kotlinCompilerPlugin.set(dependencies.compiler.auto) // determined by applied version of Kotlin. It is a default.
}
```

3. Add ability to set arguments for Compose Compiler. Now we can write:
```
compose {
    kotlinCompilerPlugin.set(dependencies.compiler.forKotlin("1.7.20"))
    kotlinCompilerPluginArgs.add("suppressKotlinVersionCompatibilityCheck=1.7.21")
}
```


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/utils/GradlePluginTestBase.kt
```diff
@@ -15,12 +15,6 @@ abstract class GradlePluginTestBase {
     val defaultTestEnvironment: TestEnvironment
         get() = TestEnvironment(workingDir = testWorkDir)
 
-    val defaultAndroidxCompilerEnvironment: TestEnvironment
```
igordmn, line 18: I removed `defaultAndroidxCompilerEnvironment` and `compose.tests.androidx.compiler.version` because they are used only in one test, and it seems there is no reason to use it in another place in the future

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -58,7 +57,6 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
 
     override fun applyToCompilation(kotlinCompilation: KotlinCompilation<*>): Provider<List<SubpluginOption>> {
         val target = kotlinCompilation.target
-        composeCompilerArtifactProvider.checkTargetSupported(target)
```
igordmn, line 61: We had a separate check for JS, when we released 1.2.0. It didn't support Kotlin 1.7.20 at that moment.

It is hard to refactor this feature in the new code, so I removed it. It is not needed now and it had an ugly code. When we will need it again, we'll write it again.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -68,6 +68,15 @@ class ComposePlugin : Plugin<Project> {
                     it.replacedBy(replacement, "org.jetbrains.compose isn't compatible with androidx.compose, because it is the same library published with different maven coordinates")
                 }
             }
+
+            project.tasks.withType(KotlinCompile::class.java) {
```
AlexeyTsvetkov, line 72: It's better to use `withType(...).configureEach` to be compatible with lazy configuration https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:old_vs_new_configuration_api_overview

AlexeyTsvetkov (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
igordmn: Fixed

ZacSweers:
Worked like a charm in 1.2.2, thanks!

ZacSweers:
actually seems not, will comment on the issue

nayea222387 (commented):
No es justo lo que hacen porque nadie me ayudó para que yo aprendiera solo mi esposo y yo

PR #2544, opened 2022-12-07
Move CodeViewer to KMM structure + add iOS target

pjBooms:


dima-avdeev-jb (approved):

experimental/examples/codeviewer/README.md
```diff
@@ -1,20 +1,27 @@
...
+Choose a run configuration for an appropriate target in IDE and run it.
+
+![run-configurations.png](run-configurations.png)
+
+To run on iOS device, please correct `iosApp/Configuration/TeamId.xcconfig` with your Apple Team ID.
+Alternatively, you may setup signing within XCode opening `iosApp/Codeviewer.xcworkspace` and then
+using "Signing & Capabilities" tab of `Codeviewer` target.
+
+Then choose **iosApp** configuration in IDE and run it
+(may also be referred as `Codeviewer` in the Run Configurations or `iosApp_` for Android studio).
```
igordmn, line 15: Is `iosApp_` only for Android Studio? Let's name it explicitly then as `iosApp (Android Studio)`.

On Windows in IDEA it looks this way:
image

pjBooms: >On Windows in IDEA it looks this way:

igordmn: > Should we mention that iOS target is available on Mac's only explicitly

It wouldn't hurt if we mention that it is required to have macOS to run the application on iOS. But we can do that in the future, if it is not simple right now. I suppose, besides macOS, we need xCode installed? If yes, then one of the solutions will be to create a separate page with the requirements. This page can also be used for Android (it requires Android SDK, and some manipulations to enable installing/debugging)

igordmn (approved):

experimental/examples/codeviewer/README.md
dima-avdeev-jb: Yeah, we need Xcode to be installed

PR #2545, opened 2022-12-07
Compose Web: add tests to check Stability inference (by counting recompositions)

eymar:
Currently these tests fail. But they will pass with compose compiler 1.3.2.2 (tested locally with a build in mavenLocal)

Or tests pass with kotlin 1.7.10


dima-avdeev-jb (approved):

dima-avdeev-jb:
Can we skip these tests for now in our dev builds?

eymar:
> Can we skip these tests for now in our dev builds? Or is it better to wait for new compiler 1.3.2.2?

I think we better wait. The PR is not urgent.

igordmn (approved):

Issue #2587, opened 2023-01-01
ExperimentalUnpackSkikoWasmRuntimeTask Gradle Configuration cache compatibility


aSemy:
When I run `./gradlew jsBrowserDevelopmentWebpack` I get errors relating to [Gradle Configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html) in my console output

```none
7 problems were found storing the configuration cache, 2 of which seem unique.
- Task `:site2:unpackSkikoWasmRuntimeJs` of type `org.jetbrains.compose.experimental.web.tasks.ExperimentalUnpackSkikoWasmRuntimeTask`: invocation of 'Task.project' at execution time is unsupported.
  See https://docs.gradle.org/7.6/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
- Task `:unpackSkikoWasmRuntimeJs` of type `org.jetbrains.compose.experimental.web.tasks.ExperimentalUnpackSkikoWasmRuntimeTask`: cannot serialize object of type 'org.gradle.api.internal.artifacts.configurations.DefaultConfiguration', a subtype of 'org.gradle.api.artifacts.Configuration', as these are not supported with the configuration cache.
  See https://docs.gradle.org/7.6/userguide/configuration_cache.html#config_cache:requirements:disallowed_types
```

https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution

> A task must not use any Project objects at execution time. This includes calling `Task.getProject()` while the task is running.

However, `ExperimentalUnpackSkikoWasmRuntimeTask` does use `Task.getProject()` during the execution phase:

https://github.com/JetBrains/compose-jb/blob/d6ff80c062e43395923d6da8c59b395c17c8478b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/experimental/web/tasks/ExperimentalUnpackSkikoWasmRuntimeTask.kt#L26-L27

https://github.com/JetBrains/compose-jb/blob/d6ff80c062e43395923d6da8c59b395c17c8478b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/experimental/web/tasks/ExperimentalUnpackSkikoWasmRuntimeTask.kt#L40-L41

[...]

PR #2589, opened 2023-01-02
Add task dependency on unpacking wasm runtime.

christiandeange:
The `jsProcessResources` task needs to depend on the `unpackSkikoWasmRuntimeJs` task since it explicitly consumes one of its outputs. Otherwise, a Gradle warning occurs during compilation that task execution optimizations have been disabled:

```
> Task :jsApp:unpackSkikoWasmRuntimeJs
Execution optimizations have been disabled for task ':jsApp:unpackSkikoWasmRuntimeJs' to ensure correctness due to the following reasons:
  - Gradle detected a problem with the following location: '/Users/chr/Documents/compose-jb/experimental/examples/falling-balls-mpp/jsApp/build/compose/skiko-wasm/js'. Reason: Task ':jsApp:jsProcessResources' uses this output of task ':jsApp:unpackSkikoWasmRuntimeJs' without declaring an explicit or implicit dependency. This can lead to incorrect results being produced, depending on what order the tasks are executed. Please refer to https://docs.gradle.org/7.5.1/userguide/validation_problems.html#implicit_dependency for more details about this problem.
```

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/experimental/web/internal/configureExperimentalWebApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/experimental/web/internal/configureExperimentalWebApplication.kt
@@ -22,4 +22,7 @@ internal fun KotlinJsIrTarget.configureExperimentalWebApplication(app: Experimen
     mainCompilation.compileKotlinTaskProvider.configure { compileTask ->
         compileTask.dependsOn(unpackRuntime)
     }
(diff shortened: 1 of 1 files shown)


dima-avdeev-jb:
Thanks, I think it may be related to the issue https://github.com/JetBrains/compose-jb/issues/2587

aSemy:
> Thanks, I think it may be related to the issue #2587

I don't think it is :) 

#2587 is specific to [Gradle Config Cache](https://docs.gradle.org/current/userguide/configuration_cache.html) compatibility. This PR is specific to [incremental build](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks).

PR #2607, opened 2023-01-09
Fix configuration cache issues

AlexeyTsvetkov:


dima-avdeev-jb (commented):
I checked on local builds, it works well. But GitHub Actions tests are failed.

PR #2614, opened 2023-01-09
resources demo js workaround

dima-avdeev-jb:


eymar:
I asked about the issue with resources in different modules in k/js. It's a known thing, but no official solution available yet.
This workaround is a valid thing to do. I took it for Compose k/wasm experiments. 
So it's okay to merge in my opinion.

dima-avdeev-jb:
@eymar 
For now, JS resources work fine if JS app "executable" is specified in shared module. So, maybe this PR is redundant.

eymar:
That's right. This workaround will be needed if we have many modules with different resources and all of them need to be included in the final distribution.
So if we keep the "executable" JS in the same module where we keep the resources, then it's fine and workaround is not needed indeed.

dima-avdeev-jb:
@eymar 
I think we can postpone current PR after Kotlin Conf.

PR #2617, opened 2023-01-10
Fix running unsigned apps locally on macOS Ventura

AlexeyTsvetkov:
Normally macOS Gatekeeper does not allow unsigned apps to run. However, apps created on the same machine were allowed to run. This allows developers to test package apps on their machines without configuring Apple Developer ID.

Previously, the Compose Multiplatform Gradle plugin simply did not do anything, when the signing was not configured. However, in macOS Ventura the Gatekeeper checks
became stricker, so "unsigned" packaged apps started to be shown as "damaged".

This seems to happen, because parts of a final app image were signed. Also they were signed by different certificates (a runtime image could be signed by a runtime vendor, while Skiko binary is signed by JetBrains).

This change removes all signatures if signing is not configured.

See also https://bugs.openjdk.org/browse/JDK-8276150


dima-avdeev-jb (approved):
Do we need to change doc for signing after the current change?
Signing_and_notarization_on_macOS/README.md

PR #2621, opened 2023-01-11
Test with Gradle 8-RC-1

AlexeyTsvetkov:


hfhbd:
What about using https://docs.gradle.org/current/userguide/toolchain_plugins.html to let Gradle download the missing JVM instead of downloading it by yourself?

dima-avdeev-jb (approved):

PR #2710, opened 2023-02-06
Update bug_report.md

igordmn:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -13,18 +13,17 @@ A clear and concise description of what the bug is.
...
-- Compose multiplatform version*:
-- OS version(s)* (required for desktop and k/native issues):
+- Compose Multiplatform version*:
+- OS version(s)* (required for Desktop and iOS issues):
 - OS Architecture (x86 or ARM64): 
 
-
 **To Reproduce**
 Steps and/or the code snippet to reproduce the behavior:
 1. Go to '...'
```
igordmn, line 29: Should we add an article in `4. See error`? (I don't know :) )

eymar: I always have doubts about articles)
In this case I took those steps from the default github template, so... I still don't know)

eymar (approved):

.github/ISSUE_TEMPLATE/bug_report.md
igordmn: If it is a default, it should be good

PR #2764, opened 2023-02-19
Added viewport units (lvw,dvw,svw,lvh,dvh,svh)

andylamax:


eymar:
Thank you!
Could you please cover the new units in the tests https://github.com/JetBrains/compose-jb/blob/master/web/core/src/jsTest/kotlin/CSSUnitApiTests.kt similar to other units?

PR #2793, opened 2023-02-23
Add resources library to  dependencies declared by compose gradle plugin

pjBooms:


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -121,6 +121,7 @@ class ComposePlugin : Plugin<Project> {
         val uiTooling get() = composeDependency("org.jetbrains.compose.ui:ui-tooling")
         val preview get() = composeDependency("org.jetbrains.compose.ui:ui-tooling-preview")
         val materialIconsExtended get() = composeDependency("org.jetbrains.compose.material:material-icons-extended")
+        val resources get() = composeDependency("org.jetbrains.compose.components:components-resources")
```
igordmn, line 124: We should move it to `DesktopComponentsDependencies` and add @ExperimentalComposeLibrary. We can't recommed it as a stable solution yet. We can move this functionality to compose-ui in the future (or just implicitly depend on it as a maven artifact)

igordmn: >add @ExperimentalComposeLibrar

Not sure though, but the rule was to add it for non-stable libraries so far :)

pjBooms: >We should move it to DesktopComponentsDependencies

It is common dependency, why to move to desktop?

pjBooms: Moved to separate CommonComponentsDependencies so it can be referenced as comose.components.resources with @OptIn(org.jetbrains.compose.ExperimentalComposeLibrary::class)

igordmn (approved):

PR #2797, opened 2023-02-25
Update README.md for accuracy

sdfgsdfgd:
Other compose README guides accurately describe which compose platform it is for, but not the desktop version. Refactor the README to reflect this.

### DIFF
--- a/tutorials/Getting_Started/README.md
+++ b/tutorials/Getting_Started/README.md
@@ -1,4 +1,4 @@
-# Getting Started with Compose Multiplatform
+# Getting Started with Compose for Desktop
 
 ## What is covered
 


sdfgsdfgd:
@igordmn @dima-avdeev-jb  

guys why can't I request a review from you, it won't let me  ??

dima-avdeev-jb:
@sdfgsdfgd We are moving to Multiplatform. In this README file we describe project creation, and users already can create projects with multiplatform support between Android and Desktop.
So I think, better to stay it as is.

sdfgsdfgd:
@dima-avdeev-jb  thank you for the explanation !  also taking the time - one more question around this if that's alright 

`users already can create projects with multiplatform support between Android and Desktop.`

do you mean we can simultaneously provide support for Desktop and Android at the same time ? Is it the same for web ? 

The only guide/tutorial/example walkthrough in the READMEs so far doesn't even involve use of Android Studio, but IntelliJ instead. Could you clarify some of these questions, can we use android studio with desktop/web support, do we necessarily need multiple modules for this, can Desktop and android reside in same project, can android-web-desktop reside in same project, and can they be supported by android studio ? Thank you :)

dima-avdeev-jb:
@sdfgsdfgd 
> do you mean we can simultaneously provide support for Desktop and Android at the same time ?
Yes, we can write one code for Android and Desktop. Because most of the JetPack Compose functionality is ported to jvm Desktop. The code should be located in a commonMain sourceSet.

>  Is it the same for web ?
No, for web we can't use the same functions like in Android and Desktop. But, maybe in the future...
Web for now is having separate functionality to provide HTML like API.

> can we use android studio with desktop/web support
Yes, it is possible. But for web, you can't attach the debugger in Android Studio and IntelliJ IDEA Community. Debugger now works in IntelliJ IDEA Ultimate or in a browser.

> do we necessarily need multiple modules
You can use one module or multiple. With one module, you can configure sorceSets in build.gradle.

> can Desktop and android reside in same project
Yes, it's possible. You can look here: https://github.com/JetBrains/compose-jb/tree/master/examples/widgets-gallery

 can android-web-desktop reside in same project
> Yes, it's possible. But you can't use the same basic Composable functions. You can look here: https://github.com/JetBrains/compose-jb/tree/master/examples/todoapp

> can they be supported by android studio ?
Yes

PR #2805, opened 2023-02-27
Update instructions to use JDK-17 or later

dima-avdeev-jb:
It needs to be compatible with the new Kotlin Wizard, which will apply jvmToolchain(17) for all jvm projects.

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -8,11 +8,9 @@
 
 ### Limitations
 
-Following limitations apply to 1.0 release.
-
   * Only 64-bit x86 Windows is supported
-  * Only JDK 11 or later is supported due to the memory management scheme used in Skia bindings
-  * Only JDK 15 or later is supported for packaging native distributions due to jpackage limitations
+  * Only JDK 17 or later is supported due to the memory management scheme used in Skia bindings
+  * Only JDK 17 or later is supported for packaging native distributions due to jpackage limitations
 
 Knowing issues on older versions:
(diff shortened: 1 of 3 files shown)


igordmn:
These versions in readme are minimal versions. Wizard shouldn't affect them and our support for them.

We can write a recommended version though. Not sure what exactly it should be - "JDK 17", "JDK 19", or "the latest available".

dima-avdeev-jb:
> These versions in readme are the minimal versions. The wizard shouldn't affect them and our support for them.
> 
> We can write a recommended version though. Not sure what exactly it should be - "JDK 17", "JDK 19", or "the latest available".

New wizard projects will work only with JDK 17 out of the box. But, users may change jvmToolchain(11) version in build.gradle.kts file for desktop module.

igordmn:
We should keep JDK 11 in VERSIONING anyway, it doesn't depend on the wizard.

In `Getting started` we can change to JDK 17, it partially depends on wizard.

igordmn (approved):

dima-avdeev-jb:
Also, we need JDK17 in compose/README (Issue here: https://github.com/JetBrains/compose-jb/issues/2826)

frankkienl:
Does using JDK17 keep compatibility with Android API 33 and lower?

According to these docs, 
https://developer.android.com/build/jdks#compileSdk
API33 uses JDK11, and API34 uses JDK17

dima-avdeev-jb:
@frankkienl 
> Does using JDK17 keep compatibility with Android API 33 and lower?

Yes

dima-avdeev-jb:
This PR is outdated

PR #2817, opened 2023-03-02
ImageViewer: Implementing redesigned UI

SebastianAigner:


dima-avdeev-jb (approved):
Checked, it works fine.
Code is not final, we need this PR to sync.
We will polish sources later.

PR #2838, opened 2023-03-07
Unzip wix to build directory

### LINKED ISSUE #2804: Use `buildDir` instead when unzipping WixTools
Recently, I noticed that a directory `wix311` is being created whenever I run the task `runDistributable`, and the culprit appears to be, https://github.com/JetBrains/compose-jb/blob/f6c36c392696a2d4229f1cd5db02248ba9b8ff62/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/wixToolset.kt#L40

I believe it should've been `root.buildDir`, rather than `root.projectDir`

Looking into the pull request related to it,
https://github.com/JetBrains/compose-jb/pull/2639/files#diff-c17db91fb640e81e12c9c330652e7388a45e41d222dc27014b8e0d7c6fbd76f3

…you'll see that `buildDir` was indeed used before that pull request.

This is currently a hassle in version control and I see no reason why it shouldn't reside in the `build` directory instead.

AlexeyTsvetkov:
Resolves #2804


eymar:
let's also take it in release/1.3.1 branch?

jasonsparc:
Hi @eymar! I just wanted to say that, while <https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.3.1> says that this fix is already included in that release, that is actually not the case.

PR #2846, opened 2023-03-08
Add test cases for expect actual functions with default values

eymar:
Currently these tests fail.
They pass if the compose compiler is built with this fix in PR https://github.com/JetBrains/compose-multiplatform-core/pull/428
We'll merge this PR when we have a compiler plugin published with a fix.
___
UPD (7.04.23):
We have a compiler plugin (1.4.5-rc03) with a fix for these tests, so it can be merged.


eymar:
We have a compiler plugin (1.4.5-rc03) with a fix for these tests, so it can be merged.

igordmn (approved):

PR #2884, opened 2023-03-15
ImageViewer: Remove mavenLocal

SebastianAigner:
...so that you don't accidentally depend on some unreleased artifacts :)


dima-avdeev-jb (approved):
Yeah, good suggestion!

PR #2886, opened 2023-03-15
SplitPane improvements and fixes.

m-sasha:
### Allow SplitPane to receive exact incoming size constraints
Fixes [Setting a modifier on the splitter makes it crash](https://github.com/JetBrains/compose-multiplatform/issues/2214).

### Correctly abide by constraint on minimum size for "second part"
The behavior before would incorrectly calculate the minimum size of the second part. Below is how it would behave when both first and second part minimum widths were set to 100.dp (200px).

https://user-images.githubusercontent.com/5230206/225445578-0be88e7c-0152-4e7d-a89f-4fa50e656645.mp4

### DIFF
--- a/components/SplitPane/library/src/desktopMain/kotlin/org/jetbrains/compose/splitpane/DesktopSplitPane.kt
+++ b/components/SplitPane/library/src/desktopMain/kotlin/org/jetbrains/compose/splitpane/DesktopSplitPane.kt
@@ -10,6 +10,9 @@ import kotlin.math.roundToInt
 
 private fun Constraints.maxByDirection(isHorizontal: Boolean): Int = if (isHorizontal) maxWidth else maxHeight
 private fun Placeable.valueByDirection(isHorizontal: Boolean): Int = if (isHorizontal) width else height
+private fun Constraints.withUnconstrainedWidth() = copy(minWidth = 0, maxWidth = Constraints.Infinity)
+private fun Constraints.withUnconstrainedHeight() = copy(minHeight = 0, maxHeight = Constraints.Infinity)
+
 
 @OptIn(ExperimentalSplitPaneApi::class)
 @Composable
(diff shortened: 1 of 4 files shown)


components/SplitPane/library/src/desktopMain/kotlin/org/jetbrains/compose/splitpane/DesktopSplitPane.kt
```diff
@@ -51,101 +50,91 @@ internal actual fun SplitPane(
...
-                val constrainedMin = constraints.minByDirection(isHorizontal) + firstMinSizePx
-                val constrainedMax =
-                    (constraints.maxByDirection(isHorizontal).toFloat() - secondMinSizePx).let {
-                        if (it <= 0 || it <= constrainedMin) {
-                            constraints.maxByDirection(isHorizontal).toFloat()
-                        } else {
-                            it
-                        }
-                    }
+                val firstMeasurable = measurables[0]
```
MatkovIvan, line 53: I guess [destructuring](https://kotlinlang.org/docs/reference/multi-declarations.html) will [work](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/component2.html) in this case

components/SplitPane/library/src/desktopMain/kotlin/org/jetbrains/compose/splitpane/DesktopSplitPane.kt
```diff
@@ -51,101 +50,91 @@ internal actual fun SplitPane(
...
-                    maxPosition = constrainedMin
-                }
+                // Need the size of the splitter to determine the min/max position
+                // Constrain the splitter only on the "other" axis
+                val splitterConstraints =
+                    if (isHorizontal) {
+                        Constraints(
+                            minHeight = constraints.minHeight,
+                            maxHeight = constraints.maxHeight
+                        )
```
MatkovIvan, line 65: I guess it's a good candidate to extract this to extension function for readability

components/SplitPane/library/src/desktopMain/kotlin/org/jetbrains/compose/splitpane/DesktopSplitPane.kt
```diff
@@ -51,101 +53,84 @@ internal actual fun SplitPane(
...
-                        if (it > 0f) {
-                            (it * positionPercentage).coerceIn(constrainedMin, constrainedMax).roundToInt()
-                        } else {
-                            constrainedMin.roundToInt()
-                        }
-                    }
+                minPosition = constrainedMin
+                maxPosition = constrainedMax
 
+                val position = (constrainedMin * (1-positionPercentage) + constrainedMax * positionPercentage)
```
Walingar, line 80: For me personally `position = (min + percentage * (max - min)).roundToInt()` is more clear, so we handle percentage only on the handleArea

Walingar (approved):
Cool! Good solution

PR #2898, opened 2023-03-20
Fix selection in codeviewer example

### LINKED ISSUE #2890: Text selection doesn't work correctly in Codeviewer
**Describe the bug**

**Affected platforms**
Select one of the platforms below:
- Desktop
- probably iOS/Web as well

**Versions**
- Compose Multiplatform version*: 1.3.0
- OS version(s)* (required for Desktop and iOS issues): Windows 11, macOS

**To Reproduce**
Steps and/or the code snippet to reproduce the behavior:
1. run examaples/codeviewer (./gradlew run)
2. try to select text, dragging to the left. Start from the top side of the line
3. Text is selecting from beggning to the starting point

**Expected behavior**
Text is selecting from starting point to the cursor

**Screenshots**
https://user-images.githubusercontent.com/5963351/225635189-64e62061-40ca-4abd-8522-58c7dbe1f57c.mp4

MatkovIvan:
Fixes #2890

### DIFF
--- a/examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
+++ b/examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
@@ -89,15 +89,15 @@ private fun java.io.File.readLinePositions(
             val ib = channel.map(
                 FileChannel.MapMode.READ_ONLY, 0, channel.size()
             )
-            var isBeginOfLine = true
             var position = 0L
+            starts.add(0)
             while (ib.hasRemaining()) {
                 val byte = ib.get()
+                position++
+                val isBeginOfLine = byte.toInt().toChar() == '\n'
                 if (isBeginOfLine) {
                     starts.add(position.toInt())
                 }
-                isBeginOfLine = byte.toInt().toChar() == '\n'
-                position++
             }
             channel.close()
         }
--- a/examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
+++ b/examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
@@ -72,14 +72,15 @@ fun java.io.File.toProjectFile(): File = object : File {
             private fun lineRange(index: Int): IntRange {
                 val startPosition = lineStartPositions[index]
                 val nextLineIndex = index + 1
-                var endPosition = if (nextLineIndex < size) {
-                    // Remove \n itself from the range
-                    lineStartPositions[nextLineIndex] - 1
-                } else byteBufferSize
+                var endPosition = if (nextLineIndex < size) lineStartPositions[nextLineIndex] else byteBufferSize
 
-                // Clip windows style line ending from the range too
-                if (endPosition > startPosition && byteBuffer[endPosition - 1].isChar('\r')) {
-                    endPosition--
+                // Remove line endings from the range
+                while (endPosition > startPosition) {
+                    val lastSymbol = byteBuffer[endPosition - 1]
+                    when (lastSymbol.toInt().toChar()) {
+                        '\n', '\r' -> endPosition--
+                        else -> break
+                    }
                 }
                 return startPosition..endPosition
             }
--- a/examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
+++ b/examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
@@ -78,7 +78,7 @@ fun java.io.File.toProjectFile(): File = object : File {
                 } else byteBufferSize
 
                 // Clip windows style line ending from the range too
-                if (byteBuffer[endPosition - 1].isChar('\r')) {
+                if (endPosition > startPosition && byteBuffer[endPosition - 1].isChar('\r')) {
                     endPosition--
                 }
                 return startPosition..endPosition
--- a/examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
+++ b/examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
@@ -59,10 +59,9 @@ fun java.io.File.toProjectFile(): File = object : File {
 
             override fun get(index: Int): String {
                 val startPosition = lineStartPositions[index]
-                var endPosition = if (index + 1 < size) lineStartPositions[index + 1] else byteBufferSize
-                endPosition-- // Skip \n
-                if (byteBuffer[endPosition].toInt().toChar() == '\r') {
-                    endPosition-- // Skip \r
+                var endPosition = if (index + 1 < size) lineStartPositions[index + 1] - 1 else byteBufferSize
+                if (endPosition < byteBufferSize && byteBuffer[endPosition].toInt().toChar() == '\r') {
+                    endPosition--
                 }
                 // Only JDK since 13 has slice() method we need, so do ugly for now.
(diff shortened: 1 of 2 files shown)


examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
```diff
@@ -59,12 +59,15 @@ fun java.io.File.toProjectFile(): File = object : File {
 
             override fun get(index: Int): String {
                 val startPosition = lineStartPositions[index]
-                val length = if (index + 1 < size) lineStartPositions[index + 1] - startPosition else
-                    byteBufferSize - startPosition
+                var endPosition = if (index + 1 < size) lineStartPositions[index + 1] else byteBufferSize
+                endPosition-- // Skip \n
```
igordmn, line 63: It fixes the example, but we can have \n and \r in the user code as well, and Compose should behave well if text have them.

igordmn (approved):

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
igordmn: Discussed, that we have ovelapped lines because of that

igordmn, line 63: Does it work with the last line? Shouldn't we check if it is really `\n`?

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
```diff
@@ -87,15 +89,15 @@ private fun java.io.File.readLinePositions(
             val ib = channel.map(
                 FileChannel.MapMode.READ_ONLY, 0, channel.size()
             )
-            var isBeginOfLine = true
             var position = 0L
+            starts.add(0)
             while (ib.hasRemaining()) {
                 val byte = ib.get()
+                position++
+                val isBeginOfLine = byte.toInt().toChar() == '\n'
```
igordmn, line 97: If I am not mistaken, in the previous version `\n` was in the end of a line. Now it is in the begining of a line.

By design, we should create a new line after `\n`, not before. Not sure why it works in codeviewer. Maybe because of `maxLines = 1`.

If it works, okay, but it is better to make a fix differently, so people won't be confused.

For example, we can remove these characters here, not in `get(index: Int)`. Or check for 2 symbols in `get(index: Int)`

MatkovIvan: No, this is not changed. This code finds indexes of `\n`, and previously the problem was that it didn't add the last one.

Your confusion is probably because of the fact that previous implementation adds the position **only on next loop iteration**.

Let me try to refactor it/write comments to make it more readable

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
```diff
@@ -59,12 +59,14 @@ fun java.io.File.toProjectFile(): File = object : File {
 
             override fun get(index: Int): String {
                 val startPosition = lineStartPositions[index]
-                val length = if (index + 1 < size) lineStartPositions[index + 1] - startPosition else
-                    byteBufferSize - startPosition
+                var endPosition = if (index + 1 < size) lineStartPositions[index + 1] - 1 else byteBufferSize
+                if (endPosition < byteBufferSize && byteBuffer[endPosition].toInt().toChar() == '\r') {
```
igordmn, line 63: What about files that don't contain `\r` symbol, just `\n`?

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
igordmn: Indeed. The results are not changed:

Before:
image

After:
image

The confusion because of `byte` corresponding to `position-1` now (not `position`). We can rename it to `previousByte`

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
igordmn: I see that we substract: `- 1`. We should probably do that for `byteBufferSize` as well. Or better check for `\n` to avoid assumptions that the last symbol is `\n`

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
```diff
@@ -58,22 +64,36 @@ fun java.io.File.toProjectFile(): File = object : File {
...
                 return StandardCharsets.UTF_8.decode(slice).toString()
             }
+
+            private fun lineRange(index: Int): IntRange {
+                val startPosition = lineStartPositions[index]
+                val nextLineIndex = index + 1
+                var endPosition = if (nextLineIndex < size) {
+                    // Remove \n itself from the range
+                    lineStartPositions[nextLineIndex] - 1
+                } else byteBufferSize
```
igordmn, line 78:
```suggestion
                } else byteBufferSize - 1
```

Should we also substract in this case?

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
```diff
@@ -58,22 +64,36 @@ fun java.io.File.toProjectFile(): File = object : File {
...
+
+                // Clip windows style line ending from the range too
+                if (byteBuffer[endPosition - 1].isChar('\r')) {
```
igordmn, line 81:
```suggestion
                if (endPosition > 0 && byteBuffer[endPosition - 1].isChar('\r')) {
```

endPosition can be zero after the previous substraction (for a file with unix-like line endings)

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
MatkovIvan: No, it's only for line separator, the character _between lines_

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
MatkovIvan: No, it can't.
Look at `lineStartPositions[nextLineIndex] - 1`:
- `nextLineIndex` always > 0
- Zero value can be only first element of `lineStartPositions`

MatkovIvan: Yep. But I guess `endPosition > startPosition` would work even better

examples/codeviewer/common/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
igordmn: Ok. What do you think if we replace it by this?

```
if (endPosition > startPosition && byteBuffer[endPosition - 1].isChar('\n')) {
    endPosition--
}
```

It is more readable

igordmn: This is also good

PR #2911, opened 2023-03-23
Resources library minSdk 21

dima-avdeev-jb:
@terrakok suggest using a lover version on minimal available Android minSdk.
JetPack Compose for Android is available since minSdk 21. Checked with a demo resource project.

### DIFF
--- a/components/resources/demo/androidApp/build.gradle.kts
+++ b/components/resources/demo/androidApp/build.gradle.kts
@@ -16,7 +16,7 @@ android {
     compileSdk = 33
     defaultConfig {
         applicationId = "me.user.androidApp"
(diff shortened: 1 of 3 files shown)


MatkovIvan (commented):
Can we downgrade `minSdk` in our examples too? I see we're using API `26` there

dima-avdeev-jb:
> Can we downgrade `minSdk` in our examples too? I see we're using API `26` there

Yes, we can do so. But at first we should merge this PR and build a new version of resources library.

terrakok (approved):

MatkovIvan (approved):

pjBooms (approved):

PR #2949, opened 2023-03-31
Clean compose/frameworks/support IDE files and temp files

dima-avdeev-jb:
This script helps to clean the workspace before opening in another IDE


igordmn (approved):
We still need to move towards removing all these scripts from this repo, and make -core repo independent, where we have all the scripts (in Bash or Gradle).

But that should be done separately, so, it is okay to continue to add them here until we do the migration.

PR #2975, opened 2023-04-04
Update experimental Codeviewer with changes from non-experimental one

pjBooms:


experimental/examples/codeviewer/shared/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
```diff
@@ -58,22 +64,37 @@ fun java.io.File.toProjectFile(): File = object : File {
...
     }
 }
 
-private fun java.io.File.readLinePositions(
-    starts: IntList
-) {
+// Backport slice from JDK 13
+private fun ByteBuffer.slice(index: Int, length: Int): ByteBuffer {
+    position(index)
+    return slice().limit(length) as ByteBuffer
```
pjBooms, line 94: limit(length) returns Buffer and not ByteBuffer on Android

igordmn (approved):

experimental/examples/codeviewer/shared/src/jvmMain/kotlin/org/jetbrains/codeviewer/platform/JvmFile.kt
MatkovIvan: Good catch. Thanks for the fix!

PR #2993, opened 2023-04-05
ImageViewer - limit zoom by the window/screen size

igordmn:
- add a `zoom` field, which is the same across different screen/window sizes

- The `scale` field is not a base state now, it is derived from the current zoom and the current screen/window size.

- drag amount is still independent of scale/zoom (if we drag by 5 pixels, the image moves by 5 pixels)

- offset is still limited by the area and the current scale

https://user-images.githubusercontent.com/5963351/230199995-f7d93a80-01b3-4cab-aeaa-f84bb2dbaa4f.mp4

### DIFF
--- a/examples/imageviewer/shared/src/commonMain/kotlin/example/imageviewer/view/ScalableImage.common.kt
+++ b/examples/imageviewer/shared/src/commonMain/kotlin/example/imageviewer/view/ScalableImage.common.kt
@@ -18,7 +18,6 @@ import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.platform.LocalDensity
 import example.imageviewer.model.ScalableState
 import example.imageviewer.utils.onPointerEvent
-import kotlin.math.min
 import kotlin.math.pow
 
 @Composable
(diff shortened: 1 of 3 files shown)


igordmn:
cc @dima-avdeev-jb

dima-avdeev-jb (approved):
Nice!
Tested, it works on all platforms.

examples/imageviewer/shared/src/commonMain/kotlin/example/imageviewer/view/ScalableImage.common.kt
```diff
@@ -55,22 +44,22 @@ internal fun ScalableImage(scalableState: ScalableState, image: ImageBitmap, mod
...
-                    scalableState.addScale(zoom, centroid - areaCenter)
+                    scalableState.addZoom(zoom, centroid - areaCenter)
                 }
                 .pointerInput(Unit) {
                     detectTapGestures(onDoubleTap = { position ->
-                        scalableState.setScale(
-                            if (scalableState.scale > 2.0) {
-                                scalableState.scaleLimits.start
+                        scalableState.setZoom(
+                            if (scalableState.zoom > 1.5f) {
```
MatkovIvan, line 59: It's not super clear what exactly this check do and why `1.5`?

igordmn: Added a comment and extracted a constant

MatkovIvan (approved):

PR #2997, opened 2023-04-06
Rename Web to HTML

igordmn:
1. Rename "compose.web" to "compose.html" in Gradle DSL. Users will see this:
image

2. Rename maven artifacts (with backward compatible "relocation" artifact). In case users don't use Gradle DSL, or when they use a library that depend on the old Compose version.

3. Rename "web" folder to "html"

An example build: 0.0.0-igor.demin-rename-web-to-html-dev1014

Will do in support/1.4.0 branch:
1. Move examples/web-* to examples/html/*
2. Rename Tutorials/Web to Tutorials/HTML
3. Rename "Compose for Web" to "Compose HTML Library" in the tutorials


html/gradlew.bat
```diff
@@ -1,89 +1,89 @@
...
-rem Set variable GRADLE_EXIT_CONSOLE if you need the _script_ return code instead of
-rem the _cmd.exe /c_ return code!
-if  not "" == "%GRADLE_EXIT_CONSOLE%" exit 1
-exit /b 1
-
-:mainEnd
-if "%OS%"=="Windows_NT" endlocal
-
-:omega
+@rem
```
igordmn, line 1: It seems it was converted to Unix-Style line endings on committing. Bat files works normally on Windows with these line endings.

html/build.gradle.kts
```diff
@@ -91,9 +92,40 @@ subprojects {
...
+                    "internal-html-core-runtime" -> "internal-web-core-runtime"
+                    "html-integration-core" -> "web-integration-core"
+                    "compose-compiler-integration" -> "compose-compiler-integration"
+                    "compose-compiler-integration-lib" -> "compose-compiler-integration-lib"
+                    else -> null
+                }
+
+                // TODO Remove this publishing in Compose 1.7. The package was migrated in 1.4.
+                if (oldArtifactId != null) {
+                    create<MavenPublication>("relocation") {
```
eymar, line 116: Do we need to run it separately on CI? or does MavenPublication plugin takes care about it?

igordmn: No need, CI calls all publishings

eymar (approved):

PR #3000, opened 2023-04-06
Support Kotlin 1.8.20

igordmn:


eymar:
Not sure if that's enough. Just a sec, I'll double check

eymar:
Perhpas, it's good change here  too https://github.com/JetBrains/compose-multiplatform/blob/01d8540500ffa69a760d0c1245ac11dadac78a3c/gradle-plugins/gradle.properties#L7

https://github.com/JetBrains/compose-multiplatform/blob/01d8540500ffa69a760d0c1245ac11dadac78a3c/gradle-plugins/gradle.properties#L9
https://github.com/JetBrains/compose-multiplatform/blob/01d8540500ffa69a760d0c1245ac11dadac78a3c/gradle-plugins/gradle.properties#L11

igordmn:
> Not sure if that's enough. Just a sec, I'll double check

Good catch, thanks. Fixed

PR #3012, opened 2023-04-07
ImageViewer, share image with description.

dima-avdeev-jb:


examples/imageviewer/shared/src/iosMain/kotlin/example/imageviewer/ImageViewer.ios.kt
```diff
@@ -57,7 +57,10 @@ fun getDependencies(ioScope: CoroutineScope, toastState: MutableState<ToastState
                         val window = UIApplication.sharedApplication.windows.last() as? UIWindow
                         val currentViewController = window?.rootViewController
                         val activityViewController = UIActivityViewController(
-                            activityItems = listOf(UIImage(data = data)),
+                            activityItems = listOf(
+                                UIImage(data = data),
+                                picture.name + "\n"  + picture.description
```
pjBooms, line 62: I think we don't need picture.name for sharing. What about Android?

dima-avdeev-jb: Done

pjBooms (approved):

PR #3015, opened 2023-04-08
Add opportunity to use custom prefixes in StyleSheet

InsanusMokrassar:


InsanusMokrassar:
Is there any chance to see this patch in the master?

InsanusMokrassar:
@eymar hello, could you please review this PR? It would be very useful to see it in master and in next update of compose (for example, `1.5.13`)

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheet.kt
```diff
@@ -38,12 +41,21 @@ class CSSRulesHolderState : CSSRulesHolder {
  * ```
  */
 open class StyleSheet(
+    prefix: String?,
     private val rulesHolder: CSSRulesHolder = CSSRulesHolderState(),
-    val usePrefix: Boolean = true,
 ) : StyleSheetBuilder, CSSRulesHolder by rulesHolder {
     private val boundClasses = mutableMapOf<String, CSSRuleDeclarationList>()
+    val prefix = prefix ?: "${this::class.simpleName}-"
```
eymar, line 48: does `prefix` need to be public? can it be private?

Also to preserve the compatibility, maybe it's worth to add a public `val usePrefix` with only getter that returns true when prefix is not null. 

I gues `val usePrefix` didn't need to be public at all, but since it's already here, it's better to avoid breaking changes.

eymar: > to see it in master and in next update of compose
it will likely go into 1.6

InsanusMokrassar: Looks like I resolved this point

InsanusMokrassar:
@eymar Thank you for your review! :heart: I have added tests and fixed the points you mentioned above

html/core/src/jsTest/kotlin/css/StyleSheetTests.kt
```diff
@@ -17,13 +17,13 @@ class StyleSheetTests {
 
     @Test
     fun extendExistingStyleSheet() {
-        val styleSheet = object : StyleSheet(usePrefix = false) {
```
eymar, line 20: let's keep the old constructor test here too. 
Feel free to add a copy of this test with new constructor

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheet.kt
```diff
@@ -38,12 +41,22 @@ class CSSRulesHolderState : CSSRulesHolder {
  * ```
  */
 open class StyleSheet(
+    prefix: String?,
```
eymar, line 44: Does it make sense in your opinion to name this constructor parameter something like `customNamePrefix` or just `customPrefix`?
So it's not confused with `protected val prefix: String`... 

Then it will be easier with this line:
`val usePrefix: Boolean = prefix == null` // by looking at this line it's not clear right away what `prefix` is meant here (from constrcutor of the property)

InsanusMokrassar: How about `selectorsPrefix`?

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheet.kt
InsanusMokrassar: Done :)

InsanusMokrassar: @eymar :)

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleSheet.kt
InsanusMokrassar: renamed to `customPrefix` :)

html/core/src/jsTest/kotlin/css/StyleSheetTests.kt
InsanusMokrassar: Fixed :)

eymar (approved):
Loos good to me!

PR #3032, opened 2023-04-11
Improve a message about incompatible kotlin version

eymar:
Add Compose Multiplatform version to the message


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
```diff
@@ -14,7 +14,7 @@ internal object ComposeCompilerCompatibility {
 
     fun compilerVersionFor(kotlinVersion: String): String {
         return kotlinToCompiler[kotlinVersion] ?: throw RuntimeException(
-            "This version of Compose Multiplatform doesn't support Kotlin " +
+            "This version of Compose Multiplatform ${ComposeBuildConfig.composeGradlePluginVersion} doesn't support Kotlin " +
```
igordmn, line 17:
```suggestion
            "Compose Multiplatform ${ComposeBuildConfig.composeGradlePluginVersion} doesn't support Kotlin " +
```

igordmn (approved):

PR #3038, opened 2023-04-11
Rewrite Compose examples

igordmn:
Also will add new screenshots for each example and target in a separate PR


examples/html/landing/README.md
```diff
@@ -1,4 +1,4 @@
-Landing page of Compose for Web
+Landing page of Compose for Web (a former name of Compose HTML Library)
```
katia-energizer, line 1: I would use the new name as primary - `Landing page of Compose HTML (previously called Compose for Web)`
or use a note instead of brackets:
> Note that Compose HTML was previously called Compose for Web. When we refer to Compose for Web now, we mean Compose based on WebAssembly.

examples/falling-balls/README.md
```diff
@@ -16,7 +16,9 @@ Choose a run configuration for an appropriate target in Android Studio and run i
 
 `./gradlew desktopApp:run`
 
-## Run JS in browser with WebAssembly Skia via Gradle
+## Run JS in browser
```
katia-energizer, line 19: Discussed with @etolstoy. Decided that it is easier to remove the whole part about `Run JS in browser` in order not to explain what we don't support. If someone explores this target themselves, they can use it.

examples/minesweeper/README.md
```diff
@@ -18,7 +18,9 @@ Choose a run configuration for an appropriate target in Android Studio and run i
 
 `./gradlew desktopApp:run`
 
-## Run JS in browser with WebAssembly Skia via Gradle
+## Run JS in browser
```
katia-energizer, line 21: Let's remove this section `Run JS in browser`. See the comment to the `chat` example.

katia-energizer (changes requested):
Left several comments.

examples/minesweeper/README.md
igordmn: Done

examples/falling-balls/README.md
igordmn: Done. Remove run configuration in IDEA as well. Only code remains

examples/html/landing/README.md
igordmn: Rewrited this way. is it okay?

A landing page example

> Note that this page describes Compose HTML which was previously called Compose for Web. When we refer to Compose for Web now, we mean Compose based on WebAssembly.

examples/README.md
```diff
@@ -0,0 +1,13 @@
+* [Imageviewer](imageviewer) - Image Viewer application (Android, iOS, Desktop)
```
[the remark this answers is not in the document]
igordmn: Changed to a table. Does it [looks](https://github.com/JetBrains/compose-multiplatform/blob/e39682f81e6f92d44b940b4234ca0792b6c62893/examples/README.md) good?

(The HTML page still without a table)

PR #3105, opened 2023-04-25
Add Xml parsing support to js platform

mdigman:
Adds support for Xml parsing in javascript with minimal changes to parser pipeline.

Side effect: moves useful error to common module for general use.

Tests: ran code against sample project that includes an appt node with a gradient via:
`./gradlew :resources:demo:shared:jsBrowserDevelopmentRun`
Looks as expected.

Screenshot 2023-04-26 at 11 01 35 AM


components/resources/library/src/jsMain/kotlin/org/jetbrains/compose/resources/xmldom/NodeImpl.kt
```diff
@@ -0,0 +1,29 @@
...
+internal open class NodeImpl(val n: DomNode): Node {
+    override val nodeName: String
+        get() = n.nodeName
+
+    override val localName = "" /* localName is not a Node property, only applies to Elements and Attrs */
+
+    override val namespaceURI = "" /* namespaceURI is not a Node property, only applies to Elements and Attrs */
+
+    override val childNodes: NodeList
+        get() =
```
eymar, line 15: minor: this getter creates a new `NodeList` instance every time it's invoked. Can it reuse the same instance?

eymar (approved):

components/resources/library/src/jsMain/kotlin/org/jetbrains/compose/resources/xmldom/NodeImpl.kt
mdigman: You got it. Updated to use `by lazy` so the NodeList is cached when used.

I (mostly) copied the original code from NodeImpl in jvmAndAndroidMain which has the same getter issue. Seems out of scope for this commit? Worth changing there as well.

eymar:
Lgtm

PR #3180, opened 2023-05-17
Add macOS native target to visual-effects-ny benchmark

pjBooms:


benchmarks/ios/visual-effects-ny/shared/src/commonMain/kotlin/HappyNY.kt
```diff
@@ -17,8 +17,8 @@ import org.jetbrains.compose.demo.visuals.platform.nanoTime
 import kotlin.math.*
 import kotlin.random.Random
 
-const val width = 400
-const val height = 800
+const val width = 640
+const val height = 480
```
dima-avdeev-jb, line 21: Will it display properly on iPhone?
Maybe we need some expect / actual getters

dima-avdeev-jb (approved):

benchmarks/ios/visual-effects-ny/shared/src/commonMain/kotlin/HappyNY.kt
pjBooms: Good catch. Added expect/actuals

PR #3193, opened 2023-05-21
Added example using Ballast state management in KMP Compose project.

charlee-dev:
Hi. I added an example of a Kotlin Multiplatform project for Android, iOS, Desktop and Web. The project showcases state management in the KMP project with UI shared between Android, iOS and Desktop.


kropp:
We'd like to avoid third party dependencies in samples.

Issue #3133, opened 2023-05-08
Could not infer Java runtime version for Java home directory


wiiznokes:
**Describe the bug**

When I run `./gradlew packageReleaseAppImage`, I get this error message

```
bash-5.2$ ./gradlew packageReleaseAppImage
Starting a Gradle Daemon, 1 stopped Daemon could not be reused, use --status for details
> Task :app:checkRuntime FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:checkRuntime'.
> Could not infer Java runtime version for Java home directory: /usr/lib/jvm/java-17-openjdk-17.0.6.0.10-1.fc38.x86_64
```

My project targets the JVM.

I don't understand because my `$JAVA_HOME` env variable is correct.

**Versions**
- Kotlin version*: 1.8.20
- Compose Multiplatform version*: 1.4.0
- OS version(s)* (required for Desktop and iOS issues): Linux (Fedora)
- OS architecture (x86 or arm64): x86-64
- JDK (for desktop issues): openjdk-17

**Additional context**

in my gradle file:

```
jvm {

        compilations.configureEach {

            kotlinOptions {
                jvmTarget = "17"
                freeCompilerArgs += listOf(
                    "-opt-in=kotlin.RequiresOptIn"
                )
            }
        }
        withJava()


    }
```

pjBooms:
What happens when you execute `/usr/lib/jvm/java-17-openjdk-17.0.6.0.10-1.fc38.x86_64/bin/java`?

pjBooms:
Can you execute `./gradlew packageReleaseAppImage --debug` and attach the whole output?

wiiznokes:
sure. [log.txt](https://github.com/JetBrains/compose-multiplatform/files/11432818/log.txt)

pjBooms:
That is strange that I do not see "Could not infer Java runtime version" in the attached log

wiiznokes:
Sorry you're right, it's because `./gradlew packageReleaseAppImage --debug > log.txt` doesn't include stderr I think.
Here is the correct log file: [log.txt](https://github.com/JetBrains/compose-multiplatform/files/11445246/log.txt)

AlexeyTsvetkov:
@wiiznokes Hi! Can you also try to run `./gradlew :app:checkRuntime --stacktrace` and copy the full exception stacktrace here?

wiiznokes:
[...]

PR #3194, opened 2023-05-21
Improve error messages when checking tasks

AlexeyTsvetkov:
Previously some errors in checkRuntime task
were reported as a nested exception.
By default, Gradle shows only top-level
error message of an exception, which
made some errors confusing.
For example, when javac was missing from JDK,
Gradle only showed "Could not infer Java runtime version for Java home directory". The part that said javac was missing was only shown, when Gradle was run with --stacktrace argument.
This is suboptimal UX, so this commit refactors
checkRuntime to make error messages more descriptive.

#3133

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractCheckNativeDistributionRuntime.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractCheckNativeDistributionRuntime.kt
@@ -5,7 +5,7 @@
 
 package org.jetbrains.compose.desktop.application.tasks
 
-import org.gradle.api.file.Directory
+import org.gradle.api.file.ConfigurableFileCollection
 import org.gradle.api.file.RegularFile
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.Provider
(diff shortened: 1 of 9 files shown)


dima-avdeev-jb (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractCheckNativeDistributionRuntime.kt
```diff
@@ -36,41 +36,50 @@ abstract class AbstractCheckNativeDistributionRuntime : AbstractComposeDesktopTa
...
-            """|Packaging native distributions requires JDK runtime version >= $MIN_JAVA_RUNTIME_VERSION
-               |Actual version: '${javaRuntimeVersion ?: "<unknown>"}'
-               |Java home: ${javaHome.get()}
-            """.trimMargin()
+        val jvmRuntimeVersionString = getJavaRuntimeVersion(
+            javaExecutable = javaExecutable,
+            javacExecutable = javacExecutable
+        )
+
+        val jvmRuntimeVersion = jvmRuntimeVersionString?.toIntOrNull()
```
MatkovIvan, line 70: Applying it to `System.getProperty("java.version")` doesn't seem right because it might contain something like `11.0.4`

PR #3218, opened 2023-05-30
visual-effects-ny benchmark: remove dependencies on time and random. Introduce number of iterations

pjBooms:


benchmarks/ios/visual-effects-ny/shared/src/commonMain/kotlin/HappyNY.kt
```diff
@@ -240,10 +256,14 @@ fun NYContent() {
 
         LaunchedEffect(Unit) {
             while (true) {
-                withFrameNanos {
-                    prevTime = time
-                    time = it
-                }
+                prevTime = time
+                time += 10000000
+                delay(10)
```
igordmn, line 261: Not sure how this affects the results, but it is better to use `withFrameNanos` instead of this.

`withFrameNanos` is consistenly called one time each frame. `delay` on the other hand is unpredictable. The actual delay can be 7-13 milliseconds.

pjBooms: fixed

igordmn (approved):

PR #3220, opened 2023-05-30
Widgets gallery: Use title, not readableName

tethridge:
MainView should use title so that if a custom title is available it will be used. Hide readableName to prevent this from happening again by mistake.


AlexeyTsvetkov (approved):

AlexeyTsvetkov:
@tethridge good catch, thanks for the contribution!

PR #3287, opened 2023-06-26
Compose 1.4.1 docs, examples

igordmn:
Also:
- Remove compose-integration from replaceVersion.sh

### DIFF
--- a/ci/compose-uber-jar/gradle.properties
+++ b/ci/compose-uber-jar/gradle.properties
@@ -1,2 +1,2 @@
-compose.version=1.3.1
+compose.version=1.4.1
 kotlin.code.style=official
--- a/compose/integrations/composable-test-cases/gradle.properties
+++ b/compose/integrations/composable-test-cases/gradle.properties
@@ -4,7 +4,7 @@ kotlin.native.enableDependencyPropagation=false
 android.useAndroidX=true
 kotlin.version=1.8.20
 agp.version=7.3.0
-compose.version=1.4.1
+compose.version=1.5.0-dev1063
 kotlin.native.cacheKind=none
 
 #empty by default - a default version will be used
--- a/compose/integrations/composable-test-cases/gradle.properties
+++ b/compose/integrations/composable-test-cases/gradle.properties
@@ -4,7 +4,7 @@ kotlin.native.enableDependencyPropagation=false
 android.useAndroidX=true
 kotlin.version=1.8.20
 agp.version=7.3.0
-compose.version=1.5.0-dev1063
+compose.version=1.4.1
 kotlin.native.cacheKind=none
 
 #empty by default - a default version will be used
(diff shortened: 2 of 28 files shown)


compose/integrations/composable-test-cases/gradle.properties
```diff
@@ -4,7 +4,7 @@ kotlin.native.enableDependencyPropagation=false
 android.useAndroidX=true
 kotlin.version=1.8.20
 agp.version=7.3.0
-compose.version=1.5.0-dev1063
```
MatkovIvan, line 7: Why it was dev version? The required change is released in 1.4.1 now?

MatkovIvan (commented):
We need to check why CI fails

compose/integrations/composable-test-cases/gradle.properties
igordmn: This was changed in https://github.com/JetBrains/compose-multiplatform/pull/3246.

@eymar, should we remove `compose/integrations` from `tools/replaceVersion.sh`?

The alternative is to have 1.4.1 here, and if we need to test other versions, pass them via `-Pcompose.version=`

ci/compose-uber-jar/gradle.properties
```diff
@@ -1,2 +1,2 @@
-compose.version=1.3.1
+compose.version=1.4.1
```
igordmn, line 1: > We need to check why CI fails

CI fails, because 1.4.1 isn't published yet. Will rerun it when we publish

compose/integrations/composable-test-cases/gradle.properties
eymar: The tests cases were failing when running with kotlin 1.9.0-Beta. 
The required change is https://github.com/JetBrains/compose-multiplatform/pull/3170 and it should be in 1.4.1

eymar: I checked the release/1.4.1 branch, and #3170 is not there yet.
If we want users to be able to try kotlin 1.9.0-RC with 1.4.1, it's better to cherry pick that change.

igordmn: If we need to test Kotlin 1.9.0, we can:
1. pass `-Pcompose.version=1.5.0-dev1063` every time and have `1.4.1` in the default configuration.
2. or have this folder out of the alignment with the latest stable release and remove it from `tools/replaceVersion.sh`.

What option is more convinient?

I guess, if this folder is only for our internal purposes, we can just remove it from `tools/replaceVersion.sh`

eymar: > if this folder is only for our internal purposes

that's true.
It's rather an exception that it was needed to use a dev version here. 
Let's keep `1.5.0-dev1063` for now, and remove the folder from `tools/replaceVersion.sh`.

igordmn: Done, reverted this change, and removed folders from the script

PR #3291, opened 2023-06-26
Update ComposeCompilerCompatibility mapping

eymar:


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
```diff
@@ -10,6 +10,9 @@ internal object ComposeCompilerCompatibility {
         "1.8.0" to "1.4.0",
         "1.8.10" to "1.4.2",
         "1.8.20" to "1.4.5",
+        "1.8.21" to "1.4.7-rc01",
```
igordmn, line 13: It would be strange that a stable CM uses non-stable compiler for stable Kotlin.

We either need to build 1.4.7 or remove this line (I would remove for simplicity)

eymar: I'll build a stable 1.4.7 then. It shouldn't take long

eymar, line 13:
```suggestion
        "1.8.21" to "1.4.7",
```

eymar: fixed

igordmn (approved):

PR #3297, opened 2023-06-28
Compose benchmarks that run in attended mode for comparison JVM and K/N

pjBooms:


benchmarks/ios/jvm-vs-kotlin-native/build.gradle.kts
```diff
@@ -0,0 +1,83 @@
...
+                implementation(compose.material)
+                implementation(compose.runtime)
+                @OptIn(org.jetbrains.compose.ExperimentalComposeLibrary::class)
+                implementation(compose.components.resources)
+            }
+        }
+
+        val desktopMain by getting {
+            dependencies {
+                implementation(compose.desktop.currentOs)
```
igordmn, line 54: I committed `gradlew.bat` in the last commit (please, check).

And discovered a possible bug in 1.4.1 on `gradlew run`
```
e: file:///D:/Development/Work/compose-multiplatform/benchmarks/ios/jvm-vs-kotlin-native/src/commonMain/kotlin/MeasureComposable.kt:22:37 Unresolved reference: skia
e: file:///D:/Development/Work/compose-multiplatform/benchmarks/ios/jvm-vs-kotlin-native/src/commonMain/kotlin/MeasureComposable.kt:26:23 Cannot access class 'org.jetbrains.skia.Canvas'. Check your module classpath for missing or conflicting dependencies
```

It succesfully runs on macOs, or on 1.4.0

igordmn (approved):

benchmarks/ios/jvm-vs-kotlin-native/build.gradle.kts
igordmn: Will investigate separetely

PR #3313, opened 2023-07-05
Add a warning for a user who sets `compose.kotlinCompilerPlugin` to `androidx.compose.compiler.compiler`

eymar:
To help users figure out the reason why their compose app doesn't compile for ios or web, let's show a warning when they use `androidx.compose.compiler:compiler`

The warning will show up only when a multiplatform project contains at least one of the non-jvm targets: k/js, k/native or k/wasm

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
@@ -108,7 +108,7 @@ private const val COMPOSE_COMPILER_COMPATIBILITY_LINK =
 internal fun createWarningAboutNonCompatibleCompiler(currentCompilerPluginGroupId: String): String {
     return """
 WARNING: Usage of the Custom Compose Compiler plugin ('$currentCompilerPluginGroupId') 
-with non-JVM targets targets (Kotlin/Native, Kotlin/JS, Kotlin/WASM) is not supported.
+with non-JVM targets (Kotlin/Native, Kotlin/JS, Kotlin/WASM) is not supported.
 For more information, please visit: $COMPOSE_COMPILER_COMPATIBILITY_LINK
 """.trimMargin()
 }
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
@@ -8,6 +8,7 @@ package org.jetbrains.compose
 import org.gradle.api.Project
 import org.gradle.api.provider.Provider
 import org.jetbrains.compose.internal.ComposeCompilerArtifactProvider
+import org.jetbrains.compose.internal.mppExt
 import org.jetbrains.compose.internal.webExt
 import org.jetbrains.kotlin.gradle.plugin.*
 import org.jetbrains.kotlin.gradle.targets.js.ir.KotlinJsIrTarget
@@ -23,6 +24,23 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
             composeCompilerArtifactProvider = ComposeCompilerArtifactProvider {
                 composeExt.kotlinCompilerPlugin.orNull ?:
                     ComposeCompilerCompatibility.compilerVersionFor(target.getKotlinPluginVersion())
+            }
+        }
+        warnAboutJetpackComposeCompilerUsageForNonJvm(target)
+    }
+
+    @Suppress("NON_EXHAUSTIVE_WHEN")
+    private fun warnAboutJetpackComposeCompilerUsageForNonJvm(target: Project) {
+        val isUsingJetpackComposeCompilerPlugin =
+            composeCompilerArtifactProvider.compilerArtifact.groupId.startsWith("androidx.compose.compiler")
+
+        if (isUsingJetpackComposeCompilerPlugin) {
+            target.mppExt.targets.forEach {
+                when (it.platformType) {
+                    KotlinPlatformType.native,
+                    KotlinPlatformType.js,
+                    KotlinPlatformType.wasm -> target.logger.warn(WARN_ABOUT_JC_COMPILER)
+                }
             }
         }
     }
@@ -69,3 +87,14 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
     private fun options(vararg options: Pair<String, String>): List<SubpluginOption> =
         options.map { SubpluginOption(it.first, it.second) }
 }
+
+private const val COMPOSE_COMPILER_COMPATIBILITY_LINK =
+    "https://github.com/JetBrains/compose-jb/blob/master/VERSIONING.md"
+
+private val WARN_ABOUT_JC_COMPILER = """
+    | WARNING: You are using the 'androidx.compose.compiler' plugin in your Kotlin multiplatform project.
+    | This plugin is only guaranteed to work with JVM targets (desktop or Android).
+    | The usage with Kotlin/JS or Kotlin/Native targets is not supported and might cause issues.
+    | Make sure you are using compatible versions of the Jetpack Compose Compiler and Kotlin.
+    | You can find the compatibility table here: $COMPOSE_COMPILER_COMPATIBILITY_LINK
+""".trimMargin()
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
@@ -7,13 +7,18 @@ package org.jetbrains.compose
 
 import org.gradle.api.Project
 import org.gradle.api.provider.Provider
+import org.gradle.build.event.BuildEventsListenerRegistry
 import org.jetbrains.compose.internal.ComposeCompilerArtifactProvider
 import org.jetbrains.compose.internal.mppExtOrNull
 import org.jetbrains.compose.internal.webExt
 import org.jetbrains.kotlin.gradle.plugin.*
 import org.jetbrains.kotlin.gradle.targets.js.ir.KotlinJsIrTarget
+import javax.inject.Inject
 
-class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
+@Suppress("UnstableApiUsage")
+class ComposeCompilerKotlinSupportPlugin @Inject constructor(
+    @Suppress("UnstableApiUsage") private val buildEventsListenerRegistry: BuildEventsListenerRegistry
+) : KotlinCompilerPluginSupportPlugin {
     private lateinit var composeCompilerArtifactProvider: ComposeCompilerArtifactProvider
 
     override fun apply(target: Project) {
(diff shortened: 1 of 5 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -25,6 +26,23 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
                     ComposeCompilerCompatibility.compilerVersionFor(target.getKotlinPluginVersion())
             }
         }
+        warnAboutJetpackComposeCompilerUsageForNonJvm(target)
+    }
+
+    @Suppress("NON_EXHAUSTIVE_WHEN")
+    private fun warnAboutJetpackComposeCompilerUsageForNonJvm(target: Project) {
+        val isUsingJetpackComposeCompilerPlugin =
+            composeCompilerArtifactProvider.compilerArtifact.groupId.startsWith("androidx.compose.compiler")
```
igordmn, line 35: Maybe we should check `!composeCompilerArtifactProvider.compilerArtifact.groupId.startsWith("org.jetbrains.compiler")`

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -25,6 +26,23 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
                     ComposeCompilerCompatibility.compilerVersionFor(target.getKotlinPluginVersion())
             }
         }
+        warnAboutJetpackComposeCompilerUsageForNonJvm(target)
```
igordmn, line 29: Not sure yet, but may we should to call it inside `ComposeCompilerArtifactProvider`. We created Provider for a reason that on the `apply` stage isn't everything ready yet.

And if we do that, we need to show it only once, adding a check `isWarningShown`

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -68,4 +86,15 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
...
+}
+
+private const val COMPOSE_COMPILER_COMPATIBILITY_LINK =
+    "https://github.com/JetBrains/compose-jb/blob/master/VERSIONING.md"
+
+private val WARN_ABOUT_JC_COMPILER = """
+    | WARNING: You are using the 'androidx.compose.compiler' plugin in your Kotlin multiplatform project.
+    | This plugin is only guaranteed to work with JVM targets (desktop or Android).
+    | The usage with Kotlin/JS or Kotlin/Native targets is not supported and might cause issues.
+    | Make sure you are using compatible versions of the Jetpack Compose Compiler and Kotlin.
```
igordmn, line 98: We can't make compatibility table between Jetpack Compose Compiler, only between JetBrains Compose Compiler

eymar: Thanks! it's a typo

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
eymar: I moved the call to `project.afterEvaluate {}`, so it's called only once per project (gradle module with compose applied)

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
eymar: good idea. done

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/KotlinCompatibilityTest.kt
```diff
@@ -23,6 +23,22 @@ class KotlinCompatibilityTest : GradlePluginTestBase() {
     @Test
     fun testKotlinJsMpp_1_7_20() = testJsMpp("1.7.20")
 
+    @Suppress("RedundantUnitExpression")
+    @Test
+    fun testKotlinJs_shows_warning_for_androidx_compose_compiler() = testProject(
```
igordmn, line 28: Thanks! Can we add a test for jvm and check that it doesn't contain this warning?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -25,6 +26,25 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
                     ComposeCompilerCompatibility.compilerVersionFor(target.getKotlinPluginVersion())
             }
         }
+        target.afterEvaluate {
+            warnAboutJetpackComposeCompilerUsageForNonJvm(it)
```
igordmn, line 30: This will be shown multiple times, as for the error (https://github.com/JetBrains/compose-multiplatform/issues/2866).

The fix is probably using some global AtomicBoolean, but I am not sure where to place it (Gradle can cache confiugurations).

So, i just mentioned it to not forget to fix it in the future.

igordmn (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
AlexeyTsvetkov: @eymar @igordmn 

If there is a cache entry for a particular set of scripts, plugins, build arguments and environment variables, configuration phase will be skipped (including `afterEvaluate` and all code under `Plugin.apply`)

One way to avoid that and show a warning just once is to use a build service.

```
import org.gradle.api.Plugin
import org.gradle.api.Project
import org.gradle.api.logging.Logging
import org.gradle.api.provider.Provider
import org.gradle.api.provider.SetProperty
import org.gradle.api.services.BuildService
import org.gradle.api.services.BuildServiceParameters
import org.gradle.api.services.BuildServiceRegistry
import org.gradle.build.event.BuildEventsListenerRegistry
import org.gradle.tooling.events.FinishEvent
import org.gradle.tooling.events.OperationCompletionListener
import javax.inject.Inject

// The service implements OperationCompletionListener just so we Gradle would use the service
// even if the service is not used by any task or transformation
abstract class ComposeMultiplatformBuildService : BuildService<ComposeMultiplatformBuildService.Parameters>, OperationCompletionListener, AutoCloseable {
    interface Parameters : BuildServiceParameters {
        val postBuildWarnings: SetProperty<String>
    }

    private val log = Logging.getLogger(this.javaClass)

    override fun close() {
        parameters.postBuildWarnings.get().forEach {
            log.error(it)
        }
    }

    override fun onFinish(event: FinishEvent) {}

    companion object {
        fun configure(project: Project, fn: Parameters.() -> Unit): Provider<ComposeMultiplatformBuildService> =
            project.gradle.sharedServices.registerOrConfigure<Parameters, ComposeMultiplatformBuildService> {
                fn()
            }
    }
}

inline fun <reified P : BuildServiceParameters, reified S : BuildService<P>> BuildServiceRegistry.registerOrConfigure(
    crossinline fn: P.() -> Unit
): Provider<S> {
    val serviceClass = S::class.java
    val serviceFqName = serviceClass.canonicalName
    val existingService = registrations.findByName(serviceFqName)
        ?.apply { (parameters as? P)?.fn() }
        ?.service
    return (existingService as? Provider<S>)
        ?: registerIfAbsent(serviceFqName, serviceClass) {
            parameters.fn()
        }
}
```

The service first be registered like this:
```
abstract class ComposePlugin : Plugin<Project> {
    @Inject
    abstract fun getEventsListenerRegistry(): BuildEventsListenerRegistry

    override fun apply(project: Project) {
        val service = ComposeMultiplatformBuildService.configure(project) {}
        // we add the service to the BuildEventsListenerRegistry, so
        // Gradle uses it even if the service is not used by tasks
        getEventsListenerRegistry().onTaskCompletion(service)
    }
}
```

Then, a warning can be added like this:
```
ComposeMultiplatformBuildService.configure(project) {
    postBuildWarnings.add("A warning that will be shown once per build!")
}   
```

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -68,4 +88,17 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
...
     private fun options(vararg options: Pair<String, String>): List<SubpluginOption> =
         options.map { SubpluginOption(it.first, it.second) }
+}
+
+private const val COMPOSE_COMPILER_COMPATIBILITY_LINK =
+    "https://github.com/JetBrains/compose-jb/blob/master/VERSIONING.md#using-compose-multiplatform-compiler"
+
+private fun createWarningAboutNonCompatibleCompiler(currentCompilerPluginGroupId: String): String {
+    return """
+    | WARNING: You are using the '$currentCompilerPluginGroupId' compiler plugin in your Kotlin multiplatform project.
```
AlexeyTsvetkov, line 98: I suggest to keep the message as concise as possible, so users are less likely to ignore it.
Especially if the warning would be shown for each affected module,
in which case a more verbose warning becomes visual noise.

My attempt:
```
WARNING: Using Custom Compose Compiler plugin ('$currentCompilerPluginGroupId') 
with non-JVM targets targets (Kotlin/Native, Kotlin/JS, Kotlin/WASM) is not supported.
More information: $COMPOSE_COMPILER_COMPATIBILITY_LINK
```

Alternatively, you could ask in `#proofread-me` Slack channel.

At the very least, `multiplatform`, `jvm`, and `dekstop` (target) should start with an upper case character.

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/KotlinCompatibilityTest.kt
```diff
@@ -23,6 +23,22 @@ class KotlinCompatibilityTest : GradlePluginTestBase() {
...
+        TestProjects.customCompilerArgs, defaultTestEnvironment.copy(
+            kotlinVersion = "1.8.22",
+            composeCompilerPlugin = "\"androidx.compose.compiler:compiler:1.4.8\"",
+            composeCompilerArgs = "\"suppressKotlinVersionCompatibilityCheck=1.8.22\""
+        )
+    ).let {
+        it.gradle(":compileKotlinJs").checks {
+            check.taskSuccessful(":compileKotlinJs")
+            check.logContains("WARNING: You are using the 'androidx.compose.compiler' compiler plugin in your Kotlin multiplatform project.")
```
AlexeyTsvetkov, line 37: Maybe using a const from the main source set could be more convenient in case anyone wants to edit the message.

AlexeyTsvetkov (approved):
Generally LGTM, but I suggest considering using a build service to show a warning even if configuration is loaded from the cache, and to show warning only once https://github.com/JetBrains/compose-multiplatform/pull/3313/files#r1253668708

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
eymar: ack

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/KotlinCompatibilityTest.kt
eymar: The const is not possible because we use a template message. But I made a function internal, so we call it in the tests as well.

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/KotlinCompatibilityTest.kt
eymar: done

AlexeyTsvetkov (approved):

PR #3330, opened 2023-07-07
Migrate to notarytool

### LINKED ISSUE #3208: Switch to notarytool
Currently both the [guide](https://github.com/JetBrains/compose-multiplatform/blob/master/tutorials/Signing_and_notarization_on_macOS/README.md) [gradle plugin](https://github.com/JetBrains/compose-multiplatform/blob/dab4531fe4ffe8509ac4ce7a7336f1016960d67c/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractUploadAppForNotarizationTask.kt#L42) use`altool` which has been deprecated in favor of [notarytool](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution/customizing_the_notarization_workflow#3087734)

DRSchlaubi:
This is currently untested and is mainly to discuss the needed API changes.

I didn't forget about `processUploadToolOutput` and will update it, once I got to try it on a MAC and can take a look at the actual output

Fixes #3208

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/MacOSNotarizationSettings.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/MacOSNotarizationSettings.kt
@@ -9,6 +9,7 @@ import org.gradle.api.model.ObjectFactory
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.ProviderFactory
 import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 import org.jetbrains.compose.desktop.application.internal.ComposeProperties
 import org.jetbrains.compose.internal.utils.nullableProperty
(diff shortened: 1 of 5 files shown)


igordmn (commented):
Thanks!

>  discuss the needed API changes.

We should replace all properties for which we find a replacement, and deprecate others.

Let us know when the PR is ready for review.

DRSchlaubi:
> Thanks!
> 
> > discuss the needed API changes.
> 
> We should replace all properties for which we find a replacement, and deprecate others.
> 
> Let us know when the PR is ready for review.

You can review the API changes, those shouldn't change, it's just that I can't test the response parser yet

igordmn:
The current API change ascProvider -> teamId looks good to me.

And if it is possible to keep `ascProvider` working, we should do that (now it throws an exception).

Is it in the same format as the new parameter teamId?

DRSchlaubi:
> The current API change ascProvider -> teamId looks good to me.
> 
> And if it is possible to keep `ascProvider` working, we should do that (now it throws an exception for existing usages).
> 
> Is it in the same format as the new parameter teamId?

Not at all if I understood the docs correctly, that's why I disabled it

igordmn:
If it is not possible to convert `asc-provider` to `team-id`, then an exception is fine.

igordmn:
The alternative is still use `altool` if `ascProvider` is provided to give people seamless transition for a few month until it stops working in 11.2023

DRSchlaubi:
Is that really worth it? Most people probably don't use that option anyways and this will require adding two seperate implementations of the task to delay a change the have to do anyw

igordmn:
Sounds reasonable, let's just throw an exception

DRSchlaubi:
Ig I need to update the tests as well

DRSchlaubi:
Superseded by #3642

PR #3336, opened 2023-07-10
Add custom VideoSurface for the VideoPlayer example

DrewCarlson:
With this approach the Video can be properly displayed within your Compose UI and removes the need for a SwingPanel.

image


mahozad:
I think the following pull request is relavent:
- https://github.com/JetBrains/compose-multiplatform/pull/2306/files

DrewCarlson:
Thanks, did not see that one.  Would be great to have either merged, the solution is not very discoverable as a PR.

This feels like a bad case to demonstrate Swing interop with.  There are better examples and given the limitations there is little value in a video player that cannot have overlapping controls.

avently:
Hey, @DrewCarlson. You did perfect job, works like a charm! I spend so many hours trying to make Swing variant working but you just give me a chance to remove this problem from my desk. Thank you!

mahozad:
[Here is another implementation](https://github.com/mahozad/cutcon/blob/v4/src/main/kotlin/ir/mahozad/cutcon/component/MediaPlayer.kt).

kropp:

Issue #3113, opened 2023-04-30
iOS: Include resources without needing to run `pod install`


alexzaitsev:
**Describe the bug**
Using a [compose-multiplatform-ios-android-template](https://github.com/JetBrains/compose-multiplatform-ios-android-template), try to add a new resource and use it in Composable function in shared module. It successfully works for Android but crashes for iOS.

**Affected platforms**
- iOS

**Versions**
- Kotlin version*: 1.8.20
- Compose Multiplatform version*: 1.4.0
- OS version(s)* (required for Desktop and iOS issues): 16.4
- OS architecture (x86 or arm64): 
- JDK (for desktop issues): 

**To Reproduce**
1. Clone [the repo](https://github.com/alexzaitsev/test-resource-bug).
2. Run the project on iOS simulator.
3. Press the button on the screen to display the image.
4. Observe crash stacktrace.

The resource I'm using:
```xml
<vector xmlns:android="http://schemas.android.com/apk/res/android" android:width="24dp" android:height="24dp"
    android:autoMirrored="true" android:tint="#FFFFFF" android:viewportWidth="24"
    android:viewportHeight="24">
    <path android:fillColor="#ffffffff"
        android:pathData="M20,11H7.83l5.59,-5.59L12,4l-8,8 8,8 1.41,-1.41L7.83,13H20v-2z" />
</vector>
```
I cloned a fresh template and added [this commit](https://github.com/alexzaitsev/test-resource-bug/commit/3c6347f7d66dbac28b7b057bf03ffc9df437b89d) to demonstrate the issue.   
**Stacktrace:**
```
Uncaught Kotlin exception: org.jetbrains.compose.resources.MissingResourceException: Missing resource with path: ic_arrow_back.xml
    at 0   My application                      0x100cec877        kfun:kotlin.Throwable#<init>(kotlin.String?){} + 123 
    at 1   My application                      0x100ce51d3        kfun:kotlin.Exception#<init>(kotlin.String?){} + 119 
    at 2   My application                      0x1014dec97        kfun:org.jetbrains.compose.resources.MissingResourceException#<init>(kotlin.String){} + 195 
    at 3   My application                      0x1014de877        kfun:org.jetbrains.compose.resources.UIKitResourceImpl.readBytes#internal + 1907 
    at 4   My application                      0x1014cbdef        kfun:org.jetbrains.compose.resources.$readBytesSync$lambda$0COROUTINE$0.invokeSuspend#internal + 455 
[...]

Issue #3073, opened 2023-04-18
iOS Resources without CocoaPods


dima-avdeev-jb:
Ability to use org.jetbrains.compose.resources.painterResource without requiring CocoaPods.
Related to https://github.com/JetBrains/compose-multiplatform/issues/3066

raghavpai:
This feature will be a game changer

ronjunevaldoz:
I thought I was the only one encountered this issue

dima-avdeev-jb:
@mpierucci Can you please create a minimal reproducible sample project on GitHub?
Also - you can try to create a project with [new wizard](https://kmp.jetbrains.com/) and compare gradle files with your own project.

mpierucci:
Hey @dima-avdeev-jb  Thanks for the reply. I was able to get a minimal reproducer based on the new wizard but mimicking our set up closely.
The App runs well on Android but fails on iOS simulator with the above-mentioned error 
[You can find the reproducer here](https://github.com/mpierucci/ComposeMPResourcesIssue)

terrakok:
The issue fixed long time ago. Check the [documentation](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-images-resources.html) carefully: multimodule projects are not supported yet. Your reproducer has two modules, am I right?

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-3073">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

Issue #3066, opened 2023-04-17
Regular Framework Compose support for iOS


raghavpai:
Why is cocoapods always necessary for building compose for iOS, will we get the regular framework support for iOS in future?

dima-avdeev-jb:
Yes, we plan to use a regular framework in the future as well.

StefanOltmann:
@dima-avdeev-jb 
Does this issue include that we will be able to use `org.jetbrains.compose.resources.painterResource` (as showcased in https://github.com/JetBrains/compose-multiplatform-ios-android-template/blob/main/shared/src/commonMain/kotlin/App.kt) to load SVG files **without** requiring CocoaPods or should I create a separate issue for that?

dima-avdeev-jb:
@StefanOltmann 
I think it is good to have separate issues for that
I created a new one: https://github.com/JetBrains/compose-multiplatform/issues/3073

raghavpai:
@dima-avdeev-jb Sorry to bother you

Just wanted to know if the above feature is prioritised and a tentative release of it is it planned?


Needed to know for my upcoming projects.

Thanks

dima-avdeev-jb:
> Just wanted to know if the above feature is prioritised and a tentative release of it is it planned?

We are not planning to do it in the next release. But we want to fix it in the future. (Hope in Summer 2023)

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-3066">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #3340, opened 2023-07-11
Simplify resource management for iOS

AlexeyTsvetkov:
Introduces a new task 'sync<FRAMEWORK_CLASSIFIER>ComposeIosResources', which collects resources from all source sets, included in iOS targets.

Now, CocoaPods integration does not require any configuration or calling 'pod install' after changing resources. * Important: existing projects need to remove 'extraSpecAttributes["resources"] = ...' from build scripts, and rerun `./gradlew podInstall` once!
Regular static frameworks should also just work.
Regular dynamic frameworks were not tested.
Currently, the task only copies resources from the module, which exports the framework.

Resolves:
 - #3073
 - #3113
 - #3066

### DIFF
--- a/benchmarks/ios/animation-from-template/shared/build.gradle.kts
+++ b/benchmarks/ios/animation-from-template/shared/build.gradle.kts
@@ -24,7 +24,6 @@ kotlin {
             baseName = "shared"
             isStatic = true
         }
-        extraSpecAttributes["resources"] = "['src/commonMain/resources/**', 'src/iosMain/resources/**']"
     }
 
     sourceSets {
--- a/benchmarks/ios/animation-from-template/shared/build.gradle.kts
+++ b/benchmarks/ios/animation-from-template/shared/build.gradle.kts
@@ -24,6 +24,7 @@ kotlin {
             baseName = "shared"
             isStatic = true
         }
+        extraSpecAttributes["resources"] = "['src/commonMain/resources/**', 'src/iosMain/resources/**']"
     }
 
     sourceSets {
--- a/components/gradle.properties
+++ b/components/gradle.properties
@@ -5,7 +5,7 @@ kotlin.code.style=official
 # __KOTLIN_COMPOSE_VERSION__
 kotlin.version=1.8.20
 # __LATEST_COMPOSE_RELEASE_VERSION__
-compose.version=1.4.1
+compose.version=0.0.0-dev1099
 agp.version=7.3.1
 org.jetbrains.compose.experimental.jscanvas.enabled=true
 org.jetbrains.compose.experimental.macos.enabled=true
--- a/components/gradle.properties
+++ b/components/gradle.properties
@@ -5,7 +5,7 @@ kotlin.code.style=official
 # __KOTLIN_COMPOSE_VERSION__
 kotlin.version=1.8.20
 # __LATEST_COMPOSE_RELEASE_VERSION__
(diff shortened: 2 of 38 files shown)


pjBooms:
+ restructure sources in the way that resource management logic will be in a separate package

benchmarks/ios/animation-from-template/shared/build.gradle.kts
```diff
@@ -24,7 +24,6 @@ kotlin {
             baseName = "shared"
             isStatic = true
         }
-        extraSpecAttributes["resources"] = "['src/commonMain/resources/**', 'src/iosMain/resources/**']"
```
pjBooms, line 27: The changes in the samples and in the benchmarks should not be a part of this commit until we have a stable compose build that contains new resource management logic. We may create a separate PR for them that will be merged once we have such a version. And I would recommend to migrate all of them to regular framework and add a separate sample with CocoaPods usage.

components/resources/library/src/iosMain/kotlin/org/jetbrains/compose/resources/Resource.ios.kt
```diff
@@ -18,8 +18,10 @@ actual fun resource(path: String): Resource = UIKitResourceImpl(path)
 @ExperimentalResourceApi
 private class UIKitResourceImpl(path: String) : AbstractResourceImpl(path) {
     override suspend fun readBytes(): ByteArray {
-        val absolutePath = NSBundle.mainBundle.resourcePath + "/" + path
```
pjBooms, line 21: We have a demo project https://github.com/JetBrains/compose-multiplatform/tree/master/components/resources/demo that is used to test the functionality. It should be synchronized with the changes in the library. So we need to migrate iOS part of the demo to new functionality for this and test it with locally build. When a dev build is available (after merge of this PR) it is better to update compose version to a dev build of the project.

pjBooms (changes requested):

dima-avdeev-jb (approved):
Checked in this branch: https://github.com/JetBrains/compose-multiplatform/pull/3351
Thanks, it works!!!

components/gradle.properties
```diff
@@ -5,7 +5,7 @@ kotlin.code.style=official
 # __KOTLIN_COMPOSE_VERSION__
 kotlin.version=1.8.20
 # __LATEST_COMPOSE_RELEASE_VERSION__
-compose.version=1.4.1
+compose.version=0.0.0-dev1099
```
pjBooms, line 8: Change this to beta release version after release

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/experimental/uikit/tasks/SyncComposeResourcesForIosTask.kt
```diff
@@ -0,0 +1,98 @@
...
+            }
+            allResources
+        }
+
+    @get:OutputDirectory
+    val outputDir: DirectoryProperty = objects.directoryProperty()
+
+    @TaskAction
+    fun run() {
+        val outputDir = outputDir.get().asFile.apply { mkdirs() }
```
pjBooms, line 69: we need to clean the directory before copying to avoid copying resources that do not exist anymore

pjBooms (changes requested):

PR #3378, opened 2023-07-20
Add commonTest source set to Minesweeper example

pjBooms:
- for GameControllerTest to be able to run
+ add linker options for ability to run it on iOS simulator

### DIFF
--- a/examples/minesweeper/shared/build.gradle.kts
+++ b/examples/minesweeper/shared/build.gradle.kts
@@ -15,7 +15,16 @@ kotlin {
     jvm("desktop")
 
     ios()
-    iosSimulatorArm64()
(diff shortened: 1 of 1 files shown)


examples/minesweeper/shared/build.gradle.kts
```diff
@@ -15,7 +15,16 @@ kotlin {
     jvm("desktop")
 
     ios()
-    iosSimulatorArm64()
```
eymar, line 18: Maybe it's worth adding those flags to ios and macos targets too. 
Although, we will likely update to 1.5.0-beta soon, so it will be fixed automatically

eymar (approved):

PR #3380, opened 2023-07-20
Hardcode 1.8.21 as kotlin version in compose.html to make new version compatible with older kotlin

eymar:
…

### DIFF
--- a/html/buildSrc/gradle.properties
+++ b/html/buildSrc/gradle.properties
@@ -1,1 +1,1 @@
-kotlin.version=1.9.0
+kotlin.hardcodeVersion=1.8.21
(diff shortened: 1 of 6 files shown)


html/buildSrc/gradle.properties
```diff
@@ -1 +1 @@
-kotlin.version=1.9.0
+kotlin.hardcodeVersion=1.8.21
```
igordmn, line 1: Let's name it just `kotlin.version`, I'll disable propagation in CI

igordmn: Disabled for build, but kept it for tests

igordmn (approved):

PR #3384, opened 2023-07-20
Disable image viewer checks on CI

igordmn:
Because:
1. it uses experimental API, that was changed int 1.5.0-beta03 after 1.4.0
2. we have a policy to use only release versions in the examples

Temporarily, will made a fix in support/1.5.0 branch


igordmn:
Merging to unblock CI build

PR #3390, opened 2023-07-21
Components. Downgrade to Kotlin 1.8.22

igordmn:
Dowgrading to Kotlin 1.8 will allow to use `components` in Kotlin 1.8 and Kotlin 1.9 projects (iOS, JS targets). Now it is supported only in Kotlin 1.9 projects.

### DIFF
--- a/components/gradle.properties
+++ b/components/gradle.properties
@@ -3,7 +3,7 @@ android.useAndroidX=true
 android.enableJetifier=true
 kotlin.code.style=official
 # __KOTLIN_COMPOSE_VERSION__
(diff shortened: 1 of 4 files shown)


eymar:
This downgrage woud help with CfW (canvas/js) issue - https://kotlinlang.slack.com/archives/C01F2HV7868/p1689977371885429

But I'm wondering what is the original reason for this downgrade?

igordmn:
iOS projects also doesn't work with Kotlin 1.8. Downgrading helps.

PR #3392, opened 2023-07-21
Disable ProGuard's optimizations by default

### LINKED ISSUE #3387: java.lang.VerifyError when running release build
**Describe the bug**
I have just upgraded my Compose multiplatform project (desktop & android) to Kotlin 1.9.0 and Compose 1.5.0-dev1114. This works without problems when I run from the IDE or when I execute the gradle task 'runDistributable'. But when I try to execute 'runReleaseDistributable' the build part works fine too but the program crashes immediately with:

```
Exception in thread "main" java.lang.VerifyError: Instruction type does not match stack map
Exception Details:
  Location:
    androidx/compose/animation/core/AnimationSpecKt.estimateAnimationDurationMillis(FFFFF)J @604: dstore
  Reason:
    Type top (current frame, locals[68]) is not assignable to double (stack map, locals[68])

```

See the full output in the attachment. What I find strange too is that the location is pointing to AnimationSpecKt but the error happens already when I parse the command line parameters of the main program.

This all worked before with Kotlin 1.8.22 and Compose 1.5.0-dev1104.

**Affected platforms**
Select one of the platforms below:
- Desktop

**Versions**
- Kotlin version*: 1.9.0
- Compose Multiplatform version*: 1.5.0-dev1114
- OS version(s)* (required for Desktop and iOS issues): macOS Ventura 13.4.1
- OS architecture (x86 or arm64): x86
- JDK (for desktop issues): OpenJDK 17.0.7+7

**To Reproduce**
Sorry, no reproducer.

**Expected behavior**
The release should work the same way as the non-release version.

[output.txt.zip](https://github.com/JetBrains/compose-multiplatform/files/12122235/output.txt.zip)

AlexeyTsvetkov:
Resolves #3387

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
@@ -22,4 +22,5 @@ abstract class ProguardSettings @Inject constructor(
     val configurationFiles: ConfigurableFileCollection = objects.fileCollection()
     val isEnabled: Property<Boolean> = objects.notNullProperty(false)
     val obfuscate: Property<Boolean> = objects.notNullProperty(false)
+    val optimize: Property<Boolean> = objects.notNullProperty(false)
 }
--- a/tutorials/Native_distributions_and_local_execution/README.md
+++ b/tutorials/Native_distributions_and_local_execution/README.md
@@ -637,3 +637,14 @@ compose.desktop {
     }
 }
 ```
+
+ProGuard's optimization is disabled by default since Compose 1.5. To enable it, set the following property via Gradle DSL:
+```
+compose.desktop {
+    application {
+        buildTypes.release.proguard {
+            optimize.set(true)
+        }
+    }
+}
+```
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureJvmApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureJvmApplication.kt
@@ -267,6 +267,7 @@ private fun JvmApplicationContext.configureProguardTask(
     // That's why a task property is follows ProGuard design,
     // when our DSL does the opposite.
     dontobfuscate.set(settings.obfuscate.map { !it })
+    dontoptimize.set(settings.optimize.map { !it })
 
     dependsOn(unpackDefaultResources)
     defaultComposeRulesFile.set(unpackDefaultResources.flatMap { it.resources.defaultComposeProguardRules })
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractProguardTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractProguardTask.kt
@@ -38,6 +38,10 @@ abstract class AbstractProguardTask : AbstractComposeDesktopTask() {
     @get:Input
     val dontobfuscate: Property<Boolean?> = objects.nullableProperty()
 
+    @get:Optional
+    @get:Input
+    val dontoptimize: Property<Boolean?> = objects.nullableProperty()
+
     // todo: DSL for excluding default rules
     // also consider pulling coroutines rules from coroutines artifact
     // https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
@@ -107,6 +111,10 @@ abstract class AbstractProguardTask : AbstractComposeDesktopTask() {
         rootConfigurationFile.ioFile.bufferedWriter().use { writer ->
             if (dontobfuscate.orNull == true) {
                 writer.writeLn("-dontobfuscate")
+            }
+
+            if (dontoptimize.orNull == true) {
+                writer.writeLn("-dontoptimize")
             }
 
             writer.writeLn("""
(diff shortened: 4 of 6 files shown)


pjBooms (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
```diff
@@ -22,4 +22,5 @@ abstract class ProguardSettings @Inject constructor(
     val configurationFiles: ConfigurableFileCollection = objects.fileCollection()
     val isEnabled: Property<Boolean> = objects.notNullProperty(false)
     val obfuscate: Property<Boolean> = objects.notNullProperty(false)
+    val optimize: Property<Boolean> = objects.notNullProperty(false)
```
igordmn, line 25: `release` builds usually should be optimized.

Can we:
- exclude only problematic functions/classes
- write test that checks all classes, to find problematic (maybe not now, but in the future)

I wouldn't make this change only to speed up release and simplify the release process.

As an example, [Android builds](https://android.googlesource.com/platform/sdk/+/master/files/proguard-android-optimize.txt) enable optimizations in the release build, and there is the rule that each third-party library should contain a valid ProGuard rules.

mipastgt: But there is no such rule for desktop. With lots of external libraries it is often a PITA to get all the ProGuard rules right just in order to be able to do a release build. Maybe one should not carry over all Android habbits to the rest of the world.

mipastgt: This is indeed a dillemma. I have to admit that I do not have an Android background but I have been using JavaFX with GraalVM/native-image which conceptually has a similar problem and so I am wondering whether you ever considered something like the native-image agent (https://www.graalvm.org/22.0/reference-manual/native-image/Agent/) which I found to be very helpful and easier to use than guessing ProGuard rules for libraries which are not your own. Would some kind of a bridge between the agents output and ProGuard be possible? Just thinking.

igordmn: `native-image` looks interesting

AlexeyTsvetkov: > Can we:
> * exclude only problematic functions/classes
> * write test that checks all classes, to find problematic (maybe not now, but in the future)


In both cases most likely no.
I don't see any way to exclude just one class from optimization.
Also, simply forcing classloading does not trigger the verification error, which is described in the issue.
Also, I'm really not sure, if PG optimizations are actually beneficial with Hotspot. The desktop JVM is much better at optimizing code than mobile. Our primary motivation for supporting PG was to enable minification, not making bytecode faster

mipastgt: After having  a closer look at https://www.guardsquare.com/manual/configuration/optimizations I am a bit shocked to realize what PG is actually trying to do with my code by default. This really looks like fishing for trouble to me where all I really want is a reliable build combined with some reasonable code reduction. Packaging your app with the most recent JVM is probably more beneficial and safer than applying all this bit fiddling.

tutorials/Native_distributions_and_local_execution/README.md
```diff
@@ -637,3 +637,14 @@ compose.desktop {
     }
 }
 ```
+
+ProGuard's optimization is disabled by default since Compose 1.5. To enable it, set the following property via Gradle DSL:
+```
+compose.desktop {
+    application {
+        buildTypes.release.proguard {
+            optimize.set(true)
```
igordmn, line 646: Tested on Compose 1.4.1, Windows

[benchmark](https://github.com/JetBrains/compose-multiplatform/tree/igor.demin/benchmark-median1)

`./gradlew run`, OpenJDK 20
```
AnimatedVisibility: 26ms
LazyGrid: 101ms
VisualEffects: 33ms
```
`./gradlew run`, OpenJDK 18
```
AnimatedVisibility: 27ms
LazyGrid: 104ms
VisualEffects: 33ms
```

`./gradlew runDistributable`, OpenJDK 18:
```
AnimatedVisibility: 28ms
LazyGrid: 102ms
VisualEffects: 33ms
```

`./gradlew runReleaseDistributable`, OpenJDK 18:
```
AnimatedVisibility: 27ms
LazyGrid: 98ms
VisualEffects: 33ms
```

[benchmark](https://github.com/JetBrains/compose-multiplatform/tree/igor.demin/benchmark-median1-startup), startup + single frame, LazyGrid, median, 10-15 tries:
`./gradlew run`, OpenJDK 20
```
311ms
```
`./gradlew run`, OpenJDK 18
```
304ms
```

`./gradlew runDistributable`, OpenJDK 18:
```
306ms
```

`./gradlew runReleaseDistributable`, OpenJDK 18:
```
261ms
```

When we apply optimizations - no significant improvements when we render frames, but there are some improvement in the startup time.

So, it is preferable to keep optimizations working - [this PR](https://github.com/JetBrains/compose-multiplatform-core/pull/706) fixes this particular issue. Also I created [a task](https://github.com/JetBrains/compose-multiplatform/issues/3404) for the future - run all Compose tests after ProGuard optimizations.

As for disabling optimizations by default - it is debatable, let's just keep them enabled until there will be a strong argument against it.

pjBooms: Ok, let's enable optmizations again but keep it in Gradle DSL

PR #3408, opened 2023-07-25
Add pro guard optimize flag

AlexeyTsvetkov:


pjBooms (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
```diff
@@ -12,7 +12,7 @@ import org.jetbrains.compose.internal.utils.notNullProperty
 import org.jetbrains.compose.internal.utils.nullableProperty
 import javax.inject.Inject
 
-private const val DEFAULT_PROGUARD_VERSION = "7.2.2"
+private const val DEFAULT_PROGUARD_VERSION = "7.3.2"
```
igordmn, line 15: When I locallly switched to this version, I encountered other crashes. CI seems also can't build it.

It seeems switching to this version isn't easy.

ellet0: > When I locallly switched to this version, I encountered other crashes. CI seems also can't build it.
> 
> It seeems switching to this version isn't easy.

Fixed with:

```
-keep class androidx.compose.runtime.** { *; }
-keep class kotlinx.coroutines.** { *; }
```

This will exclude Kotlinx coroutines from minimization, the bundle size will be 29.4 MB for the example desktop app on macOS Apple silicon JAR file, previously 28.3 MB with the runtime error

Tested with Compose 1.6.10, Proguard 7.5.0 and Kotlin 2.0.0 with a project created by [kmp.jetbrains.com](https://kmp.jetbrains.com/)

mipastgt: Also lifting the default to 7.5.0 seems to be a must because it is the only version which is compatible with Kotlin 2.0.0.

ellet0: > Also lifting the default to 7.5.0 seems to be a must because it is the only version which is compatible with Kotlin 2.0.0.

You can try setting the Proguard version to 7.5.0, including those two rules in the comment above.

This might require more setup if you're using Material 3 (even on Proguard 7.2.2) or some additional libraries 

Don't forget to include a comment, creating an issue on your repository or a TODO indicating this is a workaround to make it easier to replace this solution or remove it.

mipastgt: I have my own projects running that way already. My comment was meant as a hint to Jetbrains because at the moment building desktop applications with Kotlin 2.0.0 is broken by default.

PR #3456, opened 2023-08-04
Add a test for composable getter in expect/actual val getter

eymar:


compose/integrations/composable-test-cases/testcases/expectActual/lib/src/commonMain/kotlin/Dependencies.kt
```diff
@@ -0,0 +1,10 @@
+import androidx.compose.runtime.Composable
+
+class Abc
+
+
+val Abc.commonIntVal: Int
+    @Composable get() = 1000
+
+expect val Abc.composableIntVal: Int
+    @Composable get
```
MatkovIvan, line 10: Do we need to test situation where expect/actual annotations do not match? (Missing `@Composable` in one part)

PR #3467, opened 2023-08-07
Handle ClassCastException for ComposeMultiplatformBuildService in Compose Gradle plugin

eymar:
ClassCastException might occur when our gradle plugin was loaded more than once.  See https://github.com/JetBrains/compose-multiplatform/issues/3459

We throw another exception with a bit more clear message.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
@@ -60,7 +60,7 @@ class ComposeCompilerKotlinSupportPlugin @Inject constructor(
             throw IllegalStateException(
                 "Failed to get ComposeMultiplatformBuildService instance." +
                         " Compose Gradle plugin was probably loaded more than once." +
-                        " Consider declaring it in the root build.gradle.kts. Original exception message: ${e.message}",
+                        " Consider declaring it in the root build.gradle.kts",
                 e
             )
         }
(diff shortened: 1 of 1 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -50,7 +51,21 @@ class ComposeCompilerKotlinSupportPlugin @Inject constructor(
...
+        val providedService = try {
+            service.get()
+        } catch (e: ClassCastException) {
+            // Compose Gradle plugin was probably loaded more than once
+            // See https://github.com/JetBrains/compose-multiplatform/issues/3459
+
+            throw IllegalStateException(
+                "Failed to get ComposeMultiplatformBuildService instance." +
+                        " Compose Gradle plugin was probably loaded more than once." +
+                        " Consider declaring it in the root build.gradle.kts. Original exception message: ${e.message}",
```
AlexeyTsvetkov, line 63: I'm not sure "Original exception message" is necessary.
`--stacktrace` or `--full-stacktrace` will print the nested exception.

AlexeyTsvetkov (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
eymar, line 63:
```suggestion
                        " Consider declaring it in the root build.gradle.kts",
```

PR #3496, opened 2023-08-15
Replace warnings with errors

AlexeyTsvetkov:
Compose Multiplatform 1.5.0 prohibits some Gradle configuration flags and Gradle DSL options.

## 1. Kotlin/Native compiler cache

By default, disabling the compiler caches for Kotlin/Native is now prohibited. The caches are disabled by the following flags in `gradle.properties` or `local.properties`:
* `kotlin.native.cacheKind=none`;
* `kotlin.native.cacheKind.<IOS_TARGET_PRESET>=none`; 

#### What can you do?
* 🟢 We recommend to remove `kotlin.native.cacheKind=none` from `gradle.properties` or `local.properties`. 
* ⚠️ If you are sure the flags are necessary, the automatic flag management and the error can be turned off by adding `compose.kotlin.native.manageCacheKind=false` to `gradle.properties`.

#### Detailed description

Disabling the compiler caches is usually not necessary because it significantly slows down incremental builds.
However, prior to Kotlin/Native 1.9.0, the compiler often could not compile Compose code with the cache. 
This has led to widespread use of `cacheKind=none` throughout many Compose Multiplatform projects. 

The root issue is partially fixed in Kotlin 1.9.0, and completely fixed in the upcoming Kotlin 1.9.20. 
Compose Multiplatform Gradle plugin now can manage the necessary caching flags by itself based on a Kotlin/Native version being used. 

To address the widespread use of the flags, Compose Multiplatform Gradle plugin now throws an error when a flag is explicitly set to `none`. 

## 2. CocoaPods resource configuration

⚠️ This section is only relevant for users of the Compose Multiplatform experimental resources library.
Users of third party libraries (such as Moko Resources or libres) should not be affected by this.

Now, explicitly setting the extra spec `resources` attribute is prohibited:
```
kotlin {
    cocoapods {
        // ...
        extraSpecAttributes["resources"] = "['src/commonMain/resources/**', 'src/iosMain/resources/**']"
    }
}
``` 

#### What can you do?

* 🟢 We recommend to remove `extraSpecAttributes["resources"]` from build scripts:

```
kotlin {
    cocoapods {
        // The following line should be removed!
        extraSpecAttributes["resources"] = "['src/commonMain/resources/**', 'src/iosMain/resources/**']"
    }
}
```
* ⚠️ Alternatively, the automatic resource management can be disabled by adding the following line to `gradle.properties`:
```
org.jetbrains.compose.ios.resources.sync=false
```

#### Detailed description

For more information about resource management improvements, see the description of the [corresponding PR](https://github.com/JetBrains/compose-multiplatform/pull/3340).

### DIFF
--- a/examples/chat/gradle.properties
+++ b/examples/chat/gradle.properties
@@ -6,11 +6,10 @@ org.gradle.jvmargs=-Xmx3g
 org.jetbrains.compose.experimental.jscanvas.enabled=true
 org.jetbrains.compose.experimental.macos.enabled=true
 org.jetbrains.compose.experimental.uikit.enabled=true
-kotlin.native.cacheKind=none
 kotlin.native.useEmbeddableCompilerJar=true
 kotlin.mpp.androidSourceSetLayoutVersion=2
 # Enable kotlin/native experimental memory model
 kotlin.native.binary.memoryModel=experimental
 kotlin.version=1.9.0
 agp.version=7.1.3
-compose.version=1.4.3
+compose.version=1.5.0-beta02
--- a/examples/chat/shared/build.gradle.kts
+++ b/examples/chat/shared/build.gradle.kts
@@ -45,7 +45,6 @@ kotlin {
             baseName = "shared"
             isStatic = true
         }
-        extraSpecAttributes["resources"] = "['src/commonMain/resources/**', 'src/iosMain/resources/**']"
     }
 
     sourceSets {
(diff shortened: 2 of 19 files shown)


igordmn:
> compose.kotlin.native.manageCacheKind

Should we rename it to "org.jetbrains.compose.kotlin.native.manageCacheKind"? To be consistent with the other property

dima-avdeev-jb (commented):
All fine, Approve!

PR #3522, opened 2023-08-17
Upgrade imageviewer to compose 1.5.0-beta02

m-sasha:
and fix all the issues that come with that.


dima-avdeev-jb:
Thanks!
We already upgrade imageviewer to 1.5.0-beta02 at `support/1.5.0` branch.
Better to target this PR there.

PR #3536, opened 2023-08-21
Add a workaround to examples to run the js app with kotlin 1.9.0

eymar:
The workaround can be removed after we switch to kotlin 1.9.10 (see https://youtrack.jetbrains.com/issue/KT-60852)


igordmn (approved):
> The workaround can be removed after we switch to kotlin 1.9.10 (see https://youtrack.jetbrains.com/issue/KT-60852)

Let's create an issue to not forget about it

PR #3541, opened 2023-08-21
Add example with cocoapods

AlexeyTsvetkov:


dima-avdeev-jb (approved):
Checked, it works!
Also I suggest to add more info in README
https://github.com/JetBrains/compose-multiplatform/pull/3548

PR #3592, opened 2023-08-30
Remove cast to KotlinJvmTarget for all targets of type jvm

AmrAfifiy:
The new android multiplatform plugin will have type `jvm` but doesn't inherit from the KotlinJvmTarget type, which makes the compose multiplatform gradle plugin not useable with the new android gradle plugin for multiplatforms builds.


m-sasha:
Thanks for the PR, we'll look at it.

AlexeyTsvetkov:
@AmrAfifiy do you have an example, where the issue could be reproduced? Generally LGTM, but I would like to add a test. I'm read to write the test myself, but I would like to clarify, what:
> makes the compose multiplatform gradle plugin not useable with the new android gradle plugin for multiplatforms builds.

means

jamesward:
Here is a fairly small reproducer: https://github.com/jamesward/hello-android/tree/compose-multiplatform

AlexeyTsvetkov (approved):

AlexeyTsvetkov:
@jamesward thanks for the reproducer! 

I have checked the fix locally. The test will be added as a separate PR.

@AmrAfifiy thanks for the contribution!

PR #3597, opened 2023-08-31
Fill kotlin version compatibility for 1.5.0 release

InsanusMokrassar:


m-sasha:
I don't think this is necessary. That table specifies the *minimum* compose version that supports each given Kotlin version. The minimum compose version that supports Kotlin 1.9.0 is still 1.4.3

PR #3608, opened 2023-09-01
Run Gradle tests in parallel

AlexeyTsvetkov:


.github/workflows/gradle-plugin.yml
```diff
@@ -22,7 +23,7 @@ jobs:
         run: |
           cd gradle-plugins
           ./gradlew assemble
-          ./gradlew :compose:check --continue
+          ./gradlew --continue :preview-rpc:test :compose:test :compose:testGradle-${{ matrix.gradle }}
```
dima-avdeev-jb, line 26: should we increase number of gradle workers?
`--max-workers     Configure the number of concurrent workers Gradle is allowed to use.`

Or it has no sence on GitHub Actions for now?

dima-avdeev-jb (approved):

.github/workflows/gradle-plugin.yml
```diff
@@ -10,6 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04, macos-12, windows-2022]
+        gradle: [7.3.3, 8.3]
```
dima-avdeev-jb, line 13: Nice idea, to configure versions in matrix. Cool!

dima-avdeev-jb (approved):

.github/workflows/gradle-plugin.yml
AlexeyTsvetkov: Does not make sense on GHA

PR #3628, opened 2023-09-05
Fix Error in snippet at tutorials/UI_Testing/README.md:9

igordmn:
It crashes because the tutorial checker uses desktop-template without uiTesting module

See https://teamcity.jetbrains.com/buildConfiguration/JetBrainsPublicProjects_Compose_Publish_2_AllGitHubRelease/4301157?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildProblemsSection=true&expandBuildChangesSection=true

Also adding "``` kotlin" hack to avoid compilation of some snippets (we use it in another places)

### DIFF
--- a/ci/templates/desktop-template/build.gradle.kts
+++ b/ci/templates/desktop-template/build.gradle.kts
@@ -17,6 +17,9 @@ dependencies {
     // (in a separate module for demo project and in testMain).
     // With compose.desktop.common you will also lose @Preview functionality
     implementation(compose.desktop.currentOs)
+
+    // Include the Test API
+    implementation(compose.desktop.uiTestJUnit4)
 }
 
 compose.desktop {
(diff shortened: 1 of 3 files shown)


ci/templates/desktop-template/build.gradle.kts
```diff
@@ -17,6 +17,9 @@ dependencies {
     // (in a separate module for demo project and in testMain).
     // With compose.desktop.common you will also lose @Preview functionality
     implementation(compose.desktop.currentOs)
+
+    // Include the Test API
+    implementation(compose.desktop.uiTestJUnit4)
```
igordmn, line 22: using `implementation` instead of `testImplementation` only for CI purposes

m-sasha (approved):

PR #3647, opened 2023-09-08
Added support to add files to jpackage resources argument

alchitry:
I wanted to add a driver (udev rule) file to my Linux install and to do this I needed to override the `postinst` script that jpackage generates. This can be done by providing a `postinst` script in the directory pointed to by the `--resource-dir` flag.

This is currently not exposed and is used only by the MacOS target. 

I added an override that allows a directory to be supplied to each OS target. It may be better if this could be specified for each `TargetFormat` type instead of per OS but I don't think it matters as I think jpackage will ignore the files it doesn't recognize.

See https://docs.oracle.com/en/java/javase/20/jpackage/override-jpackage-resources.html for the relavant reasons to override this.


alchitry:
I ended up writing a task to modify the generated deb instead as I think it is the more correct solution. I do think this addition could still be helpful for some use cases.

PR #3671, opened 2023-09-13
Update custom dialogs in the todo-lite example

MatkovIvan:


examples/todoapp-lite/shared/src/commonMain/kotlin/example/todoapp/lite/common/RootContent.kt
```diff
@@ -11,7 +13,7 @@ fun RootContent(modifier: Modifier = Modifier) {
     val state = model.state
 
     MainContent(
-        modifier = modifier,
+        modifier = modifier.background(MaterialTheme.colors.background),
```
MatkovIvan, line 16: Workaround for https://github.com/JetBrains/compose-multiplatform-core/pull/812
But I'd keep it even after this fix to align styling with material spec

pjBooms (approved):

PR #3680, opened 2023-09-14
Update versions and CHANGELOG.md

Walingar:


ci/compose-uber-jar/gradle.properties
```diff
@@ -1,2 +1,2 @@
-compose.version=1.5.1
+compose.version=1.5.10-beta01
```
pjBooms, line 1: Do we need to change Compose version to beta in samples? @igordmn

igordmn: We can change them, but we should create `support/1.5.10` branch, and merge the PR there.

`release/1.5.10` won't be merged back to master, `support/1.5.10` will be merged back to master (after the release).

igordmn: Changelog itself we should merge to master, not to `support/1.5.10`

Walingar: I thought that `support/1.5.10` is not needed. I mean I'll cherry-pick commits from here to master directly when beta will be released

Walingar: I can keep examples with 1.5.1, for me it's ok

igordmn:
The target branch should be "master"

Walingar:
Sure, will create new one

PR #3721, opened 2023-09-22
Support kotlin 1.9.20-Beta2

eymar:


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
```diff
@@ -16,7 +16,8 @@ internal object ComposeCompilerCompatibility {
         "1.9.0-RC" to "1.4.8-beta",
         "1.9.0" to "1.5.1",
         "1.9.10" to "1.5.2",
-        "1.9.20-Beta" to "1.5.2.1-Beta2"
+        "1.9.20-Beta" to "1.5.2.1-Beta2",
+        "1.9.20-Beta2" to "1.5.2.1-Beta3"
```
igordmn, line 20: In the future, we should name compiler versions the same as we name main compose versions (-beta01, -beta02)

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
```diff
@@ -16,7 +16,8 @@ internal object ComposeCompilerCompatibility {
...
     )
 
     fun compilerVersionFor(kotlinVersion: String): String {
```
igordmn, line 23: We have a failed test. Does it fail before the change?

igordmn (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
eymar: I'll check.
This change doesn't affect the version of the plugin used in the tests. So it's likely a flaky test/environemnt.

eymar: rerun helped :|

PR #3722, opened 2023-09-22
Add API to not apply the Compose Compiler plugin

eymar:
This a solution for https://github.com/JetBrains/compose-multiplatform/issues/3695

Usage:
```
compose {
    excludePlatform(KotlinPlatformType.js, KotlinPlatformType.native)
}
```

or 
```
compose {
    platformTypes.set(platformTypes.get() - setOf(KotlinPlatformType.js, KotlinPlatformType.native))
}
```

2nd sample seems to be too verbose, so maybe we can make `platformTypes` private.
___

**Note:** 
this solution will make it possible to disable the Compose Compiler plugin for all targets which have given `KotlinPlatformType`. 
A project can have many targets with the same KotlinPlatformType:

```
jvm("backend) {}
jvm("frontend") {}
// etc

```

`compose.excludePlatform(KotlinPlatformType.jvm)` will disable the compose compiler plugin for all jvm targets.
While we don't have a use case for now where it can be a problem, I wanted to highlight this.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
@@ -16,6 +16,8 @@ import org.jetbrains.kotlin.gradle.targets.js.ir.KotlinJsIrTarget
 
 class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
     private lateinit var composeCompilerArtifactProvider: ComposeCompilerArtifactProvider
+    private lateinit var applicableForPlatformTypes: Provider<Set<KotlinPlatformType>>
+
 
     override fun apply(target: Project) {
         super.apply(target)
@@ -26,6 +28,8 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
                 composeExt.kotlinCompilerPlugin.orNull ?:
                     ComposeCompilerCompatibility.compilerVersionFor(target.getKotlinPluginVersion())
             }
+
+            applicableForPlatformTypes = composeExt.platformTypes
 
             collectUnsupportedCompilerPluginUsages(target)
         }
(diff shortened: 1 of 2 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeExtension.kt
```diff
@@ -41,5 +43,26 @@ abstract class ComposeExtension @Inject constructor(
...
+    /**
+     * A set of kotlin platform types to which the Compose Compiler plugin will be applied.
+     * By default, it contains all KotlinPlatformType values.
+     * It can be used to disable the Compose Compiler plugin for one or more targets:
+     * ```
+     * platformTypes.set(platformTypes.get() - KotlinPlatformType.native)
+     * ```
+     */
+    @Suppress("MemberVisibilityCanBePrivate")
+    val platformTypes: SetProperty<KotlinPlatformType> = objects.setProperty(KotlinPlatformType::class.java).also {
```
eymar, line 55: make it private? WDYT?

eymar: when it's public it can be conveniet to use it like this:

```
platformTypes.set(setOf(KotlinPlatformType.jvm))
```

to enable jvm only and disable for all other

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerKotlinSupportPlugin.kt
```diff
@@ -62,15 +66,14 @@ class ComposeCompilerKotlinSupportPlugin : KotlinCompilerPluginSupportPlugin {
...
-            KotlinPlatformType.jvm -> true
-            KotlinPlatformType.js -> isApplicableJsTarget(kotlinCompilation.target)
-            KotlinPlatformType.androidJvm -> true
-            KotlinPlatformType.native -> true
-            KotlinPlatformType.wasm -> false
+    override fun isApplicable(kotlinCompilation: KotlinCompilation<*>): Boolean {
+        val applicableTo = applicableForPlatformTypes.get()
+
+        return when (val type = kotlinCompilation.target.platformType) {
+            KotlinPlatformType.js -> isApplicableJsTarget(kotlinCompilation.target) && applicableTo.contains(type)
```
igordmn, line 73: Let's deprecate `compose.web.targets`

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeExtension.kt
igordmn: Better to provide only one API for setting this.

I would choose `platformTypes` over `excludePlatform`. Better to not have an exclude API as the only API (it isn't flexible/extendable). And if we have override-style API, it superseeds exclude-style API.

> platformTypes.set(platformTypes.get() - setOf(KotlinPlatformType.js, KotlinPlatformType.native))

> 2nd sample seems to be too verbose

In Gradle 8 (the current Gradle) users can write::

> platformTypes = platformTypes.get() - setOf(KotlinPlatformType.js, KotlinPlatformType.native)

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeExtension.kt
```diff
@@ -41,5 +43,26 @@ abstract class ComposeExtension @Inject constructor(
      */
     val kotlinCompilerPluginArgs: ListProperty<String> = objects.listProperty(String::class.java)
 
+    /**
+     * A set of kotlin platform types to which the Compose Compiler plugin will be applied.
```
igordmn, line 47:
```suggestion
     * A set of kotlin platform types to which the Compose plugin will be applied.
```

This probably is useful for other Compose features, like disabling code generation for resources.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeExtension.kt
```diff
@@ -41,5 +43,26 @@ abstract class ComposeExtension @Inject constructor(
...
+     * By default, it contains all KotlinPlatformType values.
+     * It can be used to disable the Compose Compiler plugin for one or more targets:
```
igordmn, line 49:
```suggestion
     * It can be used to disable the Compose plugin for one or more targets:
```

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeExtension.kt
eymar: done

igordmn (approved):

PR #3725, opened 2023-09-23
Html: SVG - Add fillOpacity attribute

hfhbd:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill-opacity


eymar (approved):

eymar:
Thanks!

PR #3743, opened 2023-09-29
Enable Windows ARM64 publication

igordmn:
Everything is already configured but we didn't publish desktop-windows-arm64 "helper" dependency for another users that use "compose.desktop.currentOs" DSL.

The publication was configured in https://github.com/JetBrains/compose-multiplatform-core/pull/322 and in https://github.com/JetBrains/skiko/pull/610

Resolves https://youtrack.jetbrains.com/issue/COMPOSE-223/Finish-Windows-ARM-64-support

core PR: https://github.com/JetBrains/compose-multiplatform-core/pull/853

TODO: Test Compose on a real Windows ARM64 machine


igordmn:
Fixed in another PR

PR #3835, opened 2023-10-18
Update Kotlin, dependencies and build scripts

terrakok:


components/resources/demo/shared/build.gradle.kts
```diff
@@ -1,13 +1,21 @@
...
-version = "1.0-SNAPSHOT"
-
 kotlin {
-    android()
+    @OptIn(ExperimentalKotlinGradlePluginApi::class)
+    targetHierarchy.default()
+    androidTarget {
+        compilations.all {
+            kotlinOptions {
+                jvmTarget = "11"
```
MatkovIvan, line 15: Shouldn't we use `jvmToolchain(11)` instead?

terrakok: the jvmToolchain is not my favorite thing to use. we want to set java version only for the android target. so it seems easier to use explicit jvmTarget.

components/gradle.properties
```diff
@@ -1,21 +1,26 @@
...
+
+#Compose
 org.jetbrains.compose.experimental.jscanvas.enabled=true
 org.jetbrains.compose.experimental.macos.enabled=true
 org.jetbrains.compose.experimental.uikit.enabled=true
+compose.desktop.verbose=true
 
+#Kotlin
+kotlin.code.style=official
+kotlin.mpp.enableCInteropCommonization=true
```
igordmn, line 22: We have a different properties now. Were each property changed consciously?

igordmn (approved):

components/gradle.properties
terrakok: Yes. I set the properties by intention and removed outdated ones

terrakok: By fact, it is required only for an IDE code completion. So, I'll roll back it

dima-avdeev-jb: As I understand this property also affects publication process.
But as I correctly understand previous problems was with SkikoUIView class compatibility.
So, for now - it should be fine!

igordmn: > it is required only for an IDE code completion

> was with SkikoUIView class compatibility.

Let's keep the property only if we are sure if we need it

dima-avdeev-jb (approved):

PR #3854, opened 2023-10-23
Fixed a crash when selecting empty rows in codeviewer example

ASalavei:
Due to [known problem](https://partnerissuetracker.corp.google.com/issues/199919707), the application crashes when selecting lines with empty text.
The fix does 2 things:
- Prevents `SelectionContainer` from crashing
- Improves format of copied text by separating it into lines

### DIFF
--- a/examples/codeviewer/shared/src/commonMain/kotlin/org/jetbrains/codeviewer/ui/editor/EditorView.kt
+++ b/examples/codeviewer/shared/src/commonMain/kotlin/org/jetbrains/codeviewer/ui/editor/EditorView.kt
@@ -1,7 +1,15 @@
 package org.jetbrains.codeviewer.ui.editor
 
 import androidx.compose.foundation.background
-import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxHeight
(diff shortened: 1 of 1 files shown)


examples/codeviewer/shared/src/commonMain/kotlin/org/jetbrains/codeviewer/ui/editor/EditorView.kt
```diff
@@ -172,6 +180,7 @@ private fun codeString(str: String) = buildAnnotatedString {
         addStyle(AppTheme.code.value, strFormatted, Regex("[0-9]+"))
         addStyle(AppTheme.code.annotation, strFormatted, Regex("^@[a-zA-Z_]*"))
         addStyle(AppTheme.code.comment, strFormatted, Regex("^\\s*//.*"))
+        append("\n")
```
MatkovIvan, line 183: Should it be wrapped to some condition? Something like
```kt
append(if (strFormatted.isNotEmpty()) strFormatted else "\n")
```

 Also, I'd left a comment with a link to the issue.

ASalavei: It's better to leave as it is to keep separation between lines in copied text. Otherwise copied text will be joined in one single line.
Comments has been added.

dima-avdeev-jb (approved):

PR #3934, opened 2023-11-16
Add kotlin 2.0.0-Beta1 to ComposeCompilerCompatibility

eymar:
I decided to add the kotlin version name to the name of our artefact to make it more obvious.


igordmn (approved):
We also need to add it to https://github.com/JetBrains/kotlin-multiplatform-dev-docs/blob/master/topics/compose/compose-compatibility-and-versioning.md

PR #3956, opened 2023-11-20
update ImageViewer build file

shalva97:
As I understood according to https://kotlinlang.org/docs/whatsnew1920.html#set-up-the-target-hierarchy those `dependsOn` calls are no longer needed

Tested on iOS, MacOS and Android. It works as before


dima-avdeev-jb (approved):
Looks good to me

PR #3962, opened 2023-11-21
Disable decoys generation for k/js targets

eymar:
It should be merged along with https://github.com/JetBrains/compose-multiplatform-core/pull/912


eymar:
Alrady merged in https://github.com/JetBrains/compose-multiplatform/pull/4009

PR #3988, opened 2023-11-28
Update Key Events Tutorial

findusl:
As per this discussion on slack: https://kotlinlang.slack.com/archives/C01D6HTPATV/p1701186380386439

### DIFF
--- a/tutorials/Keyboard/README.md
+++ b/tutorials/Keyboard/README.md
@@ -166,3 +166,9 @@ fun App() {
 ```
 
 <img alt="window_keyboard" src="window_keyboard.gif" height="280" />
+
+## TextFields
+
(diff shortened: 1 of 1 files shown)


tutorials/Keyboard/README.md
```diff
@@ -166,3 +166,9 @@ fun App() {
 ```
 
 <img alt="window_keyboard" src="window_keyboard.gif" height="280" />
+
+## TextFields
+
+TextFields in Compose for Desktop consume special events of type `java.awt.event.KeyEvent.KEY_TYPED`. If you want to avoid consuming Events that have already been consumed by a TextField, you have to check for this special eventType, as TextFields will not consume `KeyEventType.KeyDown` or `KeyEventType.KeyUp` events on Desktop.
```
SebastianAigner, line 172: Heya! From your addition, it's not quite clear why users should care about this. From our private convo, you said this is relevant if you have keyboard shortcuts that shouldn't trigger while you're currently in a text field. Would you be able to provide a short code snippet that demonstrates that use case? That would also make it easier for my colleagues to see if what you're doing is idiomatic, and whether we can move forward with adding it to the tutorial. :)

SebastianAigner (changes requested):

PR #3991, opened 2023-11-29
Rename components library modules.

terrakok:
There was a problem with an android publication. Android artifactId has name "module_name"-"android" even though we explicitly renamed artifactId inside the configureMavenPublication block. It means that "components-ui-tooling-preview" android library rewrites "components-resources" android library on the maven. Because they have the same name "library-android".


dima-avdeev-jb (approved):

terrakok:
All artifacts will have the same name except the abdroid

PR #4027, opened 2023-12-11
Update kotlin version in components/resources library

eymar:
Reason: newer compose-core libs were built with kotlin 1.9.21. Previous kotlin/native version 1.9.10 is not compatible with it.


eymar:
Merging now, since it's a trivial change. 
Tested: built the library into mavenLocal.

PR #4028, opened 2023-12-11
Add K/Wasm target to components/resources library

eymar:
Changes:
- added k/wasm target to library and demo
- added libs.versions.toml with coroutines version

Tested:
- using demo project
- publishToMavenLocal

I'll setup the test separately.


components/resources/library/build.gradle.kts
```diff
@@ -122,12 +132,21 @@ kotlin {
...
             dependsOn(blockingTest)
         }
-        val jsMain by getting {
+        val webMain by creating {
             dependsOn(skikoMain)
         }
+        val jsMain by getting {
+            dependsOn(webMain)
+        }
+        val wasmJsMain by getting {
```
terrakok, line 141: We need to keep a mirroring of main and test source sets to avoid hidden problems.

components/resources/library/build.gradle.kts
```diff
@@ -122,12 +132,21 @@ kotlin {
             dependsOn(skikoTest)
             dependsOn(blockingTest)
         }
-        val jsMain by getting {
+        val webMain by creating {
```
terrakok, line 135: We need to keep a mirroring of main and test source sets to avoid hidden problems.
need to add the webTest

components/resources/library/build.gradle.kts
```diff
@@ -169,3 +188,10 @@ configureMavenPublication(
     artifactId = "components-resources",
     name = "Resources for Compose JB"
 )
+
+afterEvaluate {
+    // TODO(o.k.): remove this after we refactor jsAndWasmMain source set in skiko to get rid of broken "common" js-interop
+    tasks.configureEach {
+        if (name == "compileWebMainKotlinMetadata") enabled = false
```
terrakok, line 195: does it mean the resources library won't have the published web (wasm+browser) API?

components/resources/library/build.gradle.kts
eymar: Interesting. I didn't know about hidden problems if sources sets are not mirrored.
Do you know any YT issues or examples?
__

I'll add it.

components/resources/library/src/wasmJsTest/kotlin/org/jetbrains/compose/resources/TestUtils.wasmJs.kt
```diff
@@ -0,0 +1,7 @@
+package org.jetbrains.compose.resources
+
+import kotlinx.coroutines.CoroutineScope
+
+actual fun runBlockingTest(block: suspend CoroutineScope.() -> Unit) {
+    TODO()
```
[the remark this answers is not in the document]
eymar: Sure. It's in progress here - https://github.com/JetBrains/compose-multiplatform/pull/4031

components/resources/library/build.gradle.kts
eymar: the API will be published.
The consequence of this is that projects with jsAndWasm common source set (if it exists) will need to have this workaround too.
But I'll refactor skiko soon to get rid of this. Currently the problem originates in skiko, where I used common js-interop (common meaning in jsAndWasm), but it's not correct as I learned later.

components/resources/library/src/wasmJsTest/kotlin/org/jetbrains/compose/resources/TestUtils.wasmJs.kt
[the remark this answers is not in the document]
eymar: done

components/resources/library/build.gradle.kts
eymar: done

components/resources/library/build.gradle.kts
eymar: Adding here what we discussed separately:
With this workaround, APIs declared in webMain sourceset (common for js and wasm) won't be available in jsAndWasm source sets in user projects. It's okay for now, since the target is experimental. And in most cases that API to be used in platform-specific code. 

We aim to fix it properly.

terrakok (approved):

components/gradle.properties
```diff
@@ -8,13 +8,13 @@ android.useAndroidX=true
 
 #Versions
 kotlin.version=1.9.21
-compose.version=1.6.0-dev1323
+compose.version=0.0.0-dev1332
```
eymar, line 11:
```suggestion
compose.version=1.6.0-dev1334
```

PR #4030, opened 2023-12-11
Update gradle plugin to make sure that skiko.wasm is unpacked for web tests too

eymar:


eymar:
It's required for: https://github.com/JetBrains/compose-multiplatform/pull/4031

PR #4031, opened 2023-12-11
run kwasm tests in resources library

eymar:


eymar:
This a follow up to: https://github.com/JetBrains/compose-multiplatform/pull/4028

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestUtils.kt
```diff
@@ -2,7 +2,9 @@ package org.jetbrains.compose.resources
 
 import kotlinx.coroutines.CoroutineScope
 
-expect fun runBlockingTest(block: suspend CoroutineScope.() -> Unit)
+expect class TestReturnType
```
eymar, line 5: The test return type is different on platforms:
- void for jvm
- Promise<*> for k/js and k/wasm

Therefore we need to distinguish between them to ensure the compilation works.

terrakok (approved):

PR #4039, opened 2023-12-14
Fix CHANGELOG Jetpack Compose versions

igordmn:
We have never based Compose Multiplatform on Jetpack Compose 1.5.4/1.5.3/1.5.2:
image
(the left branch is jb-main)

We only redirect Android target to it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,10 @@
 ## Dependencies
 This version of Compose Multiplatform is based on the next Jetpack Compose libraries:
 * [Compiler 1.5.4](https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.4)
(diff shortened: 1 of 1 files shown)


eymar:
we have cherry-picked the commits from 1.5.2-1.5.4:
https://github.com/JetBrains/compose-multiplatform-core/commits/release/1.5.10 (5 commits on the top)
It was added only to the release branch, not jb-main.
https://github.com/JetBrains/compose-multiplatform-core/blob/release/1.5.11/gradle.properties#L100 (updated here too)

So technically 1.5.10 and 1.5.11 are based on JC 1.5.4, right?

igordmn:
Indeed, thanks! Then we can keep it as it is.

In the future, better to do a merge commit to have better discoverability and to avoid manual errors.

PR #4061, opened 2023-12-22
Update README.md

amir1376:
there is no `version` in `nativeDistributions` block which is causes to change version of project itself


dima-avdeev-jb (approved):
Thanks!

PR #4062, opened 2023-12-23
UIKitViewController with Compose 1.6.0 dev versions

dima-avdeev-jb:
This PR needs for future documentation changes.
@elijah-semyonov made a second implementation UIKitViewController to interop with UIViewController. And this is preferred way to interop with SwiftUI inside Compose.

Also may helps with Issue https://github.com/JetBrains/compose-multiplatform/issues/3942

### DIFF
--- a/examples/interop/ios-swiftui-in-compose/gradle.properties
+++ b/examples/interop/ios-swiftui-in-compose/gradle.properties
@@ -1,7 +1,5 @@
 kotlin.code.style=official
 xcodeproj=./iosApp
 org.gradle.jvmargs=-Xmx3g
-# Enable kotlin/native experimental memory model
-kotlin.native.binary.memoryModel=experimental
 kotlin.version=1.9.21
-compose.version=1.5.11
+compose.version=1.6.0-dev1347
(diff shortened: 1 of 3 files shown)


kocheick:
I have another kind of the same problem.
I want to render this simple Composable on iOS
```
@Composable
fun CallOutController() = ComposeUIViewController {
    Text("GREEN LIGHT")
}
```
on the iOS side I have a `class MapDelegate(val uiImage: UIImage? = null) : NSObject(), MKMapViewDelegateProtocol`... and inside the following method `override fun mapView(mapView: MKMapView, viewForAnnotation: MKAnnotationProtocol): MKAnnotationView? {....}`
I'm trying to render the Composable in this view via **`ComposeUIViewController`** but it does not render

this won't render:
```
annotationView.detailCalloutAccessoryView = UIView().apply {
                addSubview(  CallOutController().view    )
            }
```
this works fine as it is native
```
annotationView.detailCalloutAccessoryView = UIImageView( UIImage.imageNamed("icon.jpg") )
```

PR #4088, opened 2024-01-02
Fix Qualifier density

igordmn:
```
    LDPI(120), // 120/160 = 0.75
    MDPI(160), // 160/160 = 1.0
    HDPI(240), // 240/160 = 1.5
    XHDPI(320), // 320/160 = 2.0
    XXHDPI(480), // 480/160 = 3.0
    XXXHDPI(640); // 640/160 = 4.0
```

### DIFF
--- a/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/Qualifier.kt
+++ b/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/Qualifier.kt
@@ -41,7 +41,7 @@ enum class DensityQualifier(val dpi: Int) : Qualifier {
         fun selectByDensity(density: Float) = when {
             density <= 0.75 -> LDPI
             density <= 1.0 -> MDPI
-            density <= 1.33 -> HDPI
+            density <= 1.5 -> HDPI
             density <= 2.0 -> XHDPI
             density <= 3.0 -> XXHDPI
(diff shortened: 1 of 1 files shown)


imknown:
What about `tvdpi`?
> https://developer.android.com/training/multiscreen/screendensities#TaskProvideAltBmp
> 
> Resources for screens somewhere between mdpi and hdpi; approximately ~213 dpi. This isn't considered a "primary" density group. It is mostly intended for televisions, and most apps don't need it—providing mdpi and hdpi resources is sufficient for most apps, and the system scales them as appropriate. If you find it necessary to provide tvdpi resources, size them at a factor of 1.33 * mdpi. For example, a 100x100 pixel image for mdpi screens is 133x133 pixels for tvdpi.

``` diff
LDPI(120), // 120/160 = 0.75
MDPI(160), // 160/160 = 1.0
+ TVDPI(213), // 213/160 ≈ 1.33
HDPI(240), // 240/160 = 1.5
XHDPI(320), // 320/160 = 2.0
XXHDPI(480), // 480/160 = 3.0
XXXHDPI(640); // 640/160 = 4.0
```

terrakok:
There is lot of DPIs to support. At the moment we added basic common dpi for all platforms.
In the future we maybe add more (or rather [nnndpi](https://developer.android.com/guide/topics/resources/providing-resources))

PR #4090, opened 2024-01-03
Fix the bug that AsyncImage demo cannot update new images

hehua2008:
"val image: T? by produceState<T?>()" should use "load" as a key, or else it won't be re-launched for the new "load". And "painterFor = { remember { BitmapPainter(it) } }" should use "it" as a key for the "remember", or else the "painterFor" won't update for the new "image" loaded by "produceState".


hehua2008:
Please fix this bug to avoid confusing tutorials learners!

PR #4136, opened 2024-01-17
Option to pack jars as uber JAR, support Proguard for uber JAR

badmannersteam:
## Proposed changes
1. Added support to join JARs to the uber JAR with ProGuard, disabled by default:
```
compose.desktop {
    application {
        buildTypes.release.proguard {
            joinOutputJars.set(true)
        }
    }
}
```
2. All 'release' tasks now really depend on ProGuard, as stated in [tutorial](https://github.com/JetBrains/compose-multiplatform/tree/master/tutorials/Native_distributions_and_local_execution#minification--obfuscation).

## Testing
- A new auto test

- Manual:
1. Test on Windows/macOs/Linux
2. Test the new Gradle parameter `joinOutputJars`:
```
compose.desktop {
    application {
        buildTypes.release.proguard {
            joinOutputJars.set(true)
        }
    }
}
```
`false` (by default) should generate multiple jars (except for `package*UberJarForCurrentOS`)
`true` should generate a single jar in a result distribution
3. Test debug tasks:
```
run
runDistributable
createDistributable
packageUberJarForCurrentOS
```
4. Test release tasks:
```
runRelease
runReleaseDistributable
createReleaseDistributable
packageReleaseUberJarForCurrentOS
```
The jars should be reduced in size (because Proguard is enabled in the release mode)

This should be test by QA.

## Issues fixed
Fixes https://github.com/JetBrains/compose-multiplatform/issues/4129

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureJvmApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureJvmApplication.kt
@@ -7,7 +7,6 @@ package org.jetbrains.compose.desktop.application.internal
 
 import org.gradle.api.DefaultTask
 import org.gradle.api.file.DuplicatesStrategy
-import org.gradle.api.file.FileCollection
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.JavaExec
 import org.gradle.api.tasks.Sync
@@ -16,6 +15,7 @@ import org.gradle.jvm.tasks.Jar
 import org.jetbrains.compose.desktop.application.dsl.TargetFormat
 import org.jetbrains.compose.desktop.application.internal.validation.validatePackageVersions
 import org.jetbrains.compose.desktop.application.tasks.*
+import org.jetbrains.compose.desktop.tasks.AbstractJarsFlattenTask
 import org.jetbrains.compose.desktop.tasks.AbstractUnpackDefaultComposeApplicationResourcesTask
 import org.jetbrains.compose.internal.utils.*
 import org.jetbrains.compose.internal.utils.OS
@@ -26,7 +26,6 @@ import org.jetbrains.compose.internal.utils.ioFile
 import org.jetbrains.compose.internal.utils.ioFileOrNull
 import org.jetbrains.compose.internal.utils.javaExecutable
 import org.jetbrains.compose.internal.utils.provider
-import java.io.File
 
 private val defaultJvmArgs = listOf("-D$CONFIGURE_SWING_GLOBALS=true")
 internal const val composeDesktopTaskGroup = "compose desktop"
@@ -220,11 +219,18 @@ private fun JvmApplicationContext.configurePackagingTasks(
         }
     }
 
+    val flattenJars = tasks.register<AbstractJarsFlattenTask>(
+        taskNameAction = "flatten",
+        taskNameObject = "Jars"
+    ) {
+        configureFlattenJars(this, runProguard)
+    }
+
     val packageUberJarForCurrentOS = tasks.register<Jar>(
         taskNameAction = "package",
         taskNameObject = "uberJarForCurrentOS"
     ) {
-        configurePackageUberJarForCurrentOS(this, runProguard)
+        configurePackageUberJarForCurrentOS(this, flattenJars)
     }
 
     val runDistributable = tasks.register<AbstractRunDistributableTask>(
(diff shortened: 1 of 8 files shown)


terrakok:
@badmannersteam could you add integration tests for that?

badmannersteam:
@terrakok probably yes, but what combinations they should check?

terrakok:
I would expect a multi-module project + proguard as uberjar and a regular multi-module project

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureJvmApplication.kt
```diff
@@ -430,27 +434,41 @@ private fun JvmApplicationContext.configureRunTask(
...
         jar.project.files({
             files.map { if (it.isZipOrJar()) jar.project.zipTree(it) else it }
         })
 
-
-    jar.useAppRuntimeFiles { (runtimeJars, _) ->
-        from(flattenJars(runtimeJars))
+    if (runProguard != null) {
+        jar.dependsOn(runProguard)
+        jar.from(flattenJars(project.fileTree(runProguard.flatMap { it.destinationDir })))
```
badmannersteam, line 459: After some tests i've found that `from()` method executes eagerly before the execution phase when there are no proguard processed files yet, so it remembers the empty files list.

I've tried to play with `fileTree()`, `zipTree()`, `files()` and closures for creating the lazy evaluated file tree of zip trees from files from `destinationDir`, but with no luck.

The only working solution that I found was this:
```kotlin
val fs: FileSystemOperations
val archive: ArchiveOperations

fun flattenJars(files: FileCollection) = files.map { if (it.isZipOrJar()) archive.zipTree(it) else it }

val proguardDestinationDir = jar.project.fileTree(runProguard.flatMap { it.destinationDir })
val uberjarFlattenedSourcesDir = appTmpDir.dir("uberjar")
jar.doFirst {
    fs.clearDirs(uberjarFlattenedSourcesDir)
    fs.copy {
        it.duplicatesStrategy = DuplicatesStrategy.EXCLUDE
        it.from(flattenJars(proguardDestinationDir))
        it.into(uberjarFlattenedSourcesDir)
    }
}
jar.from(uberjarFlattenedSourcesDir)
```
It works fine, but has some disadvantages:
- Uses temp directory for proguard processed jars unpacking and flattening instead of `zipTree`
- I don't know where to place and inject `FileSystemOperations` and `ArchiveOperations` better.
Usage of corresponding `Project` methods is forbidden because of gradle configuration cache.

Maybe you can suggest something?

badmannersteam:
Added some tests.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureJvmApplication.kt
terrakok: yes, you are right, but I'd recommend to register a custom task to flat an uberjar before the jar task. And there you may `inject` needed operations.

badmannersteam: Fixed by a separate task.

badmannersteam:
@terrakok seems like that all for now.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureJvmApplication.kt
badmannersteam: After testing in prod i've found one more problem - on Windows unpacking to the file system removes some files because of case insensitive FS.
As a workaround we can use `-dontusemixedcaseclassnames` in proguard config, but adding it to the plugin defaults looks like a bad idea.
How else we can fix it?
Maybe completely replace the `Jar` task with smthng more powerful like a ShadowJar plugin?

badmannersteam:
I'll check the tests today and also want to fix the issue on Windows, that was mentioned above (by flattering into the intermediate jar instead of directory).

terrakok:
@badmannersteam could you check failed tests and rebase the branch on top of the master?

terrakok (approved):

badmannersteam:
@terrakok @igordmn 
I apologize for being intrusive, but are there any chances that this will be merged before the next release?
Just want to know for proper planning.

igordmn:
Most probably yes, it will be in the next release (1.6.10).

I plan to review it soon.

igordmn (approved):
Thanks!

LGTM.

Few things:
1. I changed the title/description by our new PR format
2. I moved README to [support/1.6.10](https://github.com/JetBrains/compose-multiplatform/commit/034c34dede2465b5eb1be2dd9bda5303a1560e45), which will be merged to `master` on the release date

PR #4146, opened 2024-01-19
Mark all resources public API as experimental.

terrakok:
And all public but only for generated code API as @InternalResourceApi

### DIFF
--- a/components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/FontResources.android.kt
+++ b/components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/FontResources.android.kt
@@ -5,6 +5,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.text.font.*
 
+@ExperimentalResourceApi
 @Composable
 actual fun Font(resource: FontResource, weight: FontWeight, style: FontStyle): Font {
     val environment = LocalComposeEnvironment.current.rememberEnvironment()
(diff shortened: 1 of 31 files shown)


pjBooms (approved):

components/resources/library/src/macosMain/kotlin/org/jetbrains/compose/resources/ResourceReader.macos.kt
```diff
@@ -5,6 +5,7 @@ import kotlinx.cinterop.usePinned
 import platform.Foundation.NSFileManager
 import platform.posix.memcpy
 
+@OptIn(ExperimentalResourceApi::class)
 actual suspend fun readResourceBytes(path: String): ByteArray {
```
igordmn, line 9: Do we need to mark these actuals with `@InternalResourceApi` as the `expect` is marked with it?

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/FontResources.android.kt
```diff
@@ -5,6 +5,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.text.font.*
 
+@ExperimentalResourceApi
```
igordmn, line 8: `MalformedXMLException` isn't marked with it.

Also `sealed class Resource`, but marking it Experimental can be too much, not sure 🤔.

igordmn (approved):

PR #4162, opened 2024-01-23
Configure android resources after the agp is applied

terrakok:


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResClassTask.kt
```diff
@@ -33,6 +35,13 @@ abstract class GenerateResClassTask : DefaultTask() {
             val rootResDir = resDir.get().asFile
             logger.info("Generate resources for $rootResDir")
 
+            rootResDir.walkTopDown()
+                .filter { file -> file.name == DS_STORE_FILE }
+                .forEach { file ->
+                    logger.info("Delete $DS_STORE_FILE: ${file.path}")
+                    file.delete()
```
pjBooms, line 42: I am not sure that it is ok to delete files that we don't create. Mat we just ignore hidden files?

pjBooms (changes requested):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResClassTask.kt
```diff
@@ -105,6 +104,9 @@ abstract class GenerateResClassTask : DefaultTask() {
             .map { it.attributes.getNamedItem("name").nodeValue }
         return ids.toSet()
     }
+
+    private fun File.listFilesWithoutDsStore(): List<File> =
+        listFiles()?.filter { it.name != ".DS_Store" }.orEmpty()
```
pjBooms, line 109: I think we can filter any hidden file, not only .DStore (File.isHidden can be used)

pjBooms (approved):

PR #4177, opened 2024-01-25
Accumulate characters when parsing xml element

fbarthelery:
Documentation for [XmlParserDelegate](https://developer.apple.com/documentation/foundation/xmlparserdelegate/1412539-parser) says that `parser(_:foundCharacters:)` may be called multiple times with only a part of the element.

This seems to occurs with accentuated characters in the element like :
```
    <string name="lbl_create_room">Créer une table</string>
```
Before this change this will return `éer une table` instead of `Créer une table`


terrakok:
I cherry-picked the commit there
https://github.com/JetBrains/compose-multiplatform/pull/4207

PR #4202, opened 2024-01-30
Add common test samples to widgets-gallery

m-sasha:


examples/widgets-gallery/shared/src/commonMain/kotlin/org/jetbrains/compose/demo/widgets/ui/screens/AppBars.kt
```diff
@@ -69,7 +74,7 @@ private fun TopAppBarsDemo() {
     TopAppBar(
         title = {
             Icon(
-                painterResource(ImageResource("composeRes/images/ic_twitter.xml")),
+                painterResource(DrawableResource("composeRes/images/ic_twitter.xml")),
```
terrakok, line 77: I guess we have to move the resource to `composeResources/drawable/ic_twitter.xml` and use the `Res.drawable.ic_twitter` accessor here

And for other places the same comment

terrakok (approved):

PR #4205, opened 2024-01-30
Generate initializer functions in the Res file to avoid the MethodTooLargeException

terrakok:


MatkovIvan (commented):
Should we add a reproduction of `MethodTooLargeException` case to `test-projects`?

terrakok:
> Should we add a reproduction of `MethodTooLargeException` case to `test-projects`?

I don't think so. the test will require a lot of time to run.

terrakok:
> Should we add a reproduction of `MethodTooLargeException` case to `test-projects`?

done

pjBooms (approved):

PR #4220, opened 2024-02-01
Improve handling of special characters in string resources

terrakok:
Introduced a function to process and replace certain escaped symbols like '\n', '\t', and '\uXXXX' in the strings extracted from compose string resources.


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/StringResources.kt
```diff
@@ -203,4 +205,34 @@ private suspend fun loadStringArray(
...
+ * Replaces
+ *
+ * '\n' -> new line
+ *
+ * '\t' -> tab
+ *
+ * '\uXXXX' -> unicode symbol
+ *
+ * '\\' -> '\'
+ *
```
pjBooms, line 221: What about quotes? Will be they supported in an additional commit?

pjBooms (approved):

terrakok:
Fixes: https://github.com/JetBrains/compose-multiplatform/pull/4220

PR #4240, opened 2024-02-06
Add a type name to the resource initializers

terrakok:
to avoid "Overload resolution ambiguity" for different resources with a same name

https://github.com/JetBrains/compose-multiplatform/issues/4237


pjBooms (approved):

terrakok:
fixes https://github.com/JetBrains/compose-multiplatform/issues/4237

PR #4249, opened 2024-02-06
Update CHANGELOG.md

igordmn:


CHANGELOG.md
```diff
@@ -95,7 +95,7 @@ This version of Compose Multiplatform is based on the next Jetpack Compose libra
 
 ## Highlights
 
-- Resource library improvements ([demo project](https://github.com/JetBrains/compose-multiplatform/tree/c31c761e09212eaa13014f4d0d2a6516511f859a/gradle-plugins/compose/src/test/test-projects/misc/commonResources))
+- Resource library improvements ([an example project](https://github.com/JetBrains/compose-multiplatform/tree/c31c761e09212eaa13014f4d0d2a6516511f859a/gradle-plugins/compose/src/test/test-projects/misc/commonResources))
```
terrakok, line 98: why don't we point to the real demo? https://github.com/JetBrains/compose-multiplatform/tree/master/components/resources/demo

igordmn: It is not a self-sufficient project.

From the beginning, the target audience of `components/../demo` was contributors to `components`.

If we need to provide an example for all users, we need to provide a self-sufficient project in `examples`. As a temporary measure I provided the test project.

igordmn: > It is not a self-sufficient project.

To be able to use it, users have to open the whole `components` project and depend on the sources instead of artifacts. The configuration there also differ from the external usage of the library.

terrakok: the test project is a bad example because it uses a simplified project structure and it wasn't designed as a showcase

igordmn: The `components/../demo` is also bad as it isn't self-sufficient. For now, we have to choose between two evils. And write a good one for the release.

PR #4265, opened 2024-02-07
Support Rtl in SplitPane

ahmedhosnypro:
This is related to issue: https://github.com/JetBrains/compose-multiplatform/issues/4258

changes
- DesktopSplitPane.kt: placable.place() -> placable.placeRelative

- SplitePaneDSL.kt: change the delta direction to follow the layout direction
```kotlin
 @Composable
    override fun Modifier.markAsHandle(): Modifier = this.run {
        val layoutDirection = LocalLayoutDirection.current
        pointerInput(containerScope.splitPaneState) {
            detectDragGestures { change, _ ->
                change.consume()
                containerScope.splitPaneState.dispatchRawMovement(
                    if (containerScope.isHorizontal)
                        if (layoutDirection == LayoutDirection.Ltr) change.position.x else -change.position.x
                    else change.position.y
                )
            }
        }
```

the problem with .onPointerEvent() Modifier, or onDrag also,  is
    whenever the layout direction is Ltr or Rtl:
        moving to right always produce positive change, [expected negative if dir =Rtl]
        moving to left always produce negative change, [expected positive if dir =Rtl]

the calculation of postion will fail if layoutDir is Rtl, because  positionPercentage will be out of range
```kotlin
fun dispatchRawMovement(delta: Float) {
        val movableArea = maxPosition - minPosition
        if (movableArea > 0) {
            positionPercentage =
                ((movableArea * positionPercentage) + delta).coerceIn(0f, movableArea) / movableArea
        }
    }
```

### DIFF
--- a/components/SplitPane/library/src/commonMain/kotlin/org/jetbrains/compose/splitpane/SplitPaneDSL.kt
+++ b/components/SplitPane/library/src/commonMain/kotlin/org/jetbrains/compose/splitpane/SplitPaneDSL.kt
@@ -48,7 +48,6 @@ interface SplitPaneScope {
 @ExperimentalSplitPaneApi
 interface HandleScope {
     /** allow mark composable as movable handle */
-    @Composable
     fun Modifier.markAsHandle(): Modifier
 }
 
--- a/components/SplitPane/library/src/commonMain/kotlin/org/jetbrains/compose/splitpane/SplitPaneDSL.kt
+++ b/components/SplitPane/library/src/commonMain/kotlin/org/jetbrains/compose/splitpane/SplitPaneDSL.kt
@@ -86,7 +86,7 @@ internal class HandleScopeImpl(
     private val containerScope: SplitPaneScopeImpl
 ) : HandleScope {
     @Composable
-    override fun Modifier.markAsHandle(): Modifier = this.run {
+    override fun Modifier.markAsHandle(): Modifier = composed {
         val layoutDirection = LocalLayoutDirection.current
         pointerInput(containerScope.splitPaneState) {
             detectDragGestures { change, _ ->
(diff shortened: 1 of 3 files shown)


components/SplitPane/library/src/commonMain/kotlin/org/jetbrains/compose/splitpane/SplitPaneDSL.kt
```diff
@@ -46,6 +48,7 @@ interface SplitPaneScope {
 @ExperimentalSplitPaneApi
 interface HandleScope {
     /** allow mark composable as movable handle */
+    @Composable
```
igordmn, line 51:
```suggestion
```
One of the rules of creating Modifier's is to not make them Composable.

components/SplitPane/library/src/commonMain/kotlin/org/jetbrains/compose/splitpane/SplitPaneDSL.kt
```diff
@@ -83,12 +86,18 @@ interface SplitterScope {
...
     private val containerScope: SplitPaneScopeImpl
 ) : HandleScope {
-    override fun Modifier.markAsHandle(): Modifier = this.pointerInput(containerScope.splitPaneState) {
-        detectDragGestures { change, _ ->
-            change.consume()
-            containerScope.splitPaneState.dispatchRawMovement(
-                if (containerScope.isHorizontal) change.position.x else change.position.y
-            )
+    @Composable
+    override fun Modifier.markAsHandle(): Modifier = this.run {
```
igordmn, line 90:
```suggestion
    override fun Modifier.markAsHandle(): Modifier = composed {
```
You can access `LocalLayoutDirection` this way.

Better though to convert it to [Modifier.Node](https://developer.android.com/jetpack/compose/custom-modifiers), but it is not easy - `composed` will do for now.

igordmn (commented):
Thanks!

components/SplitPane/library/src/commonMain/kotlin/org/jetbrains/compose/splitpane/SplitPaneDSL.kt
ahmedhosnypro: done, also modified it in the Implementaion

igordmn (approved):

PR #4271, opened 2024-02-08
Add possibility to modify the macOs minimum version

prof18:
Hi all 👋 

I recently tried to publish my macOS app to the App Store and the publishing failed because I wasn't including an Intel version
Screenshot 2024-02-04 at 17 31 52

The alternative could be publishing a Universal binary, but it's not quite supported now (see #1599). But by setting the minimum version of macOS to 12, it's possible to upload only arm version.

So, I've added the possibility of changing the minimum macOS version.


dima-avdeev-jb:
Thanks!

dima-avdeev-jb (approved):

eymar (approved):

PR #4278, opened 2024-02-10
Gradle Plugin: Force kotlinx-coroutines version 1.8.0-RC2 for web targets in user projects

eymar:
Removed such the version substituion from the resources library. The versions will be subtituted now by gradle plugin.


eymar:
This change is not related to 1 failing test:
```
ResourcesTest[Gradle '7.4'] > testCopyFontsInAndroidApp() FAILED
    java.io.IOException at TempDirectory.java:262
```

PR #4294, opened 2024-02-13
Fix resource accessors compilation when there are huge number of resource files.

terrakok:
Instead of object properties there are being generated extension properties in different files.

fixes https://github.com/JetBrains/compose-multiplatform/issues/4285


igordmn (approved):
Approving the general idea after online discussion. The code still need to be reviewed

pjBooms (approved):

Issue #4295, opened 2024-02-13
Error `Name contains illegal characters` in generated `Res.kt` while building jsTarget.


suhaibroomy:
**Describe the bug**
Error `Name contains illegal characters` in generated `Res.kt` while building jsTarget when the project name contains hyphen. This happens since the Plugin gets the project name and appends it to the generated files package name within ``.  This works fine on androidTargets but generates an error with JsTarget

**Affected platforms**
Select one of the platforms below:
- Web (K/Wasm) - Canvas based API

**Versions**
- Kotlin version*:  1.9.22
- Compose Multiplatform version*:  1.6.0-beta02
- OS version(s)* (required for Desktop and iOS issues): 
- OS architecture (x86 or arm64):  arm64
- JDK (for desktop issues): 

**To Reproduce**

1. Create a Compose multiplatform project
2. Add Js target
3. Set project name to something with hyphen
4. Add a font (or any resource) in commonResources
5. Run ./gradlew assemble

**Expected behaviour**
There should be no error

luca992:
is there a way to workaround this until the next stable release is up with the merged pr?

terrakok:
Yes. Just give a project name without dashes

luca992:
@terrakok a workaround that doesn't involve renaming directories/ changing module names that are already published on maven

terrakok:
You are not able to publish a library with compose resources yet. Could you explain your case?

luca992:
I'm using compose resources to load files in `commonTest`. This in combination with #4442 is preventing me from being able to publish library builds while being able to run the tests

terrakok:
https://github.com/JetBrains/compose-multiplatform/pull/4300

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-4295">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #4296, opened 2024-02-13
Fix package name of generated `Res` file when project is building for JsTarget (#4295)

suhaibroomy:
Fix package name of generated `Res` file when project is building for JsTarget (#4295)


terrakok:
Thank you for the contribution. You are supposed to provide tests cover the issue

suhaibroomy:
@terrakok Thanks for review. I have added a test case that verifies the issue is fixed. Let me know if it works or if I need to add some more test cases

terrakok:
Reworked tests: https://github.com/JetBrains/compose-multiplatform/pull/4300

PR #4383, opened 2024-02-28
Update CHANGELOG.md

igordmn:


CHANGELOG.md
```diff
@@ -131,6 +131,7 @@ _Changes since 1.5.12_
 - [Added WASM to `components.uiToolingPreview` library](https://github.com/JetBrains/compose-multiplatform/pull/4286)
 - [Fix "The cursor is invisible in compose web"](https://github.com/JetBrains/skiko/pull/846)
 - [Use an alternative implementation of `Image.toBitmap`](https://github.com/JetBrains/compose-multiplatform-core/pull/917)
+- Specifying `plugin:androidx.compose.compiler.plugins.kotlin:generateDecoys=true` is no longer needed for JS target
```
igordmn, line 134: Do we have a PR? If we have, let's add more information to it. Like removing [this code](https://kotlinlang.slack.com/archives/C01F2HV7868/p1709067758907739?thread_ts=1709058089.672359&cid=C01F2HV7868).

eymar:
```
Specifying `plugin:androidx.compose.compiler.plugins.kotlin:generateDecoys=true` is no longer needed for JS target
```
Having that flag would actually break the compilation. So it's better to say that please make sure you don't have that flag anymore if you use compose-runtime 1.6.0 or newer.

The PR was not specificly about decoys: https://github.com/JetBrains/compose-multiplatform/pull/4009

PR #4384, opened 2024-02-28
Update CHANGELOG.md

eymar:


CHANGELOG.md
```diff
@@ -53,6 +53,9 @@ _Changes since 1.5.12_
 #### Desktop
 - [Remove deprecated APIs in `TooltipArea` and `PointerEvent`](https://github.com/JetBrains/compose-multiplatform-core/pull/1029)
 
+#### Web (k/js)
+- Decoys generation was disabled. It means the libraries which depend on earlier Compose Multiplatform version are not compatible with 1.6.0 anymore. Projects which apply  Compose Compiler plugin manually need to remove this argument: `plugin:androidx.compose.compiler.plugins.kotlin:generateDecoys=true`
```
igordmn, line 57:
```suggestion
- Libraries which depend on earlier Compose Multiplatform version are not compatible with 1.6.0 anymore. This is because decoys generation was disabled. Projects which apply Compose Compiler plugin manually need to remove this argument: `plugin:androidx.compose.compiler.plugins.kotlin:generateDecoys=true`.
```

We should make the accent on incompatibility.

igordmn (approved):

Issue #4237, opened 2024-02-05
[Bug] `Overload resolution ambiguity` in Compose Resources Gradle plugin


ChristianKatzmann:
**Describe the bug**
If we define two resources with different type but same name, the build fails with an `Overload resolution ambiguity` as the initializer functions introduced in #4205 does not contain the type in its name.

```
e: ui/build/generated/compose/resourceGenerator/kotlin/ui/generated/resources/Res.kt:30:55 Overload resolution ambiguity:
private fun get_powered_by(): DrawableResource defined in ui.generated.resources in file Res.kt
private fun get_powered_by(): StringResource defined in ui.generated.resources in file Res.kt
e: ui/build/generated/compose/resourceGenerator/kotlin/ui/generated/resources/Res.kt:84:53 Overload resolution ambiguity:
private fun get_powered_by(): DrawableResource defined in package.ui.generated.resources in file Res.kt
private fun get_powered_by(): StringResource defined in ui.generated.resources in file Res.kt
e: ui/build/generated/compose/resourceGenerator/kotlin/ui/generated/resources/Res.kt:110:1 Conflicting overloads: private fun get_powered_by(): DrawableResource defined in ui.generated.resources in file Res.kt, private fun get_powered_by(): StringResource defined in ui.generated.resources in file Res.kt
e: ui/build/generated/compose/resourceGenerator/kotlin/ui/generated/resources/Res.kt:298:1 Conflicting overloads: private fun get_powered_by(): DrawableResource defined in ui.generated.resources in file Res.kt, private fun get_powered_by(): StringResource defined in ui.generated.resources in file Res.kt
```

_This log is from my own project, redacted._

**Affected platforms**
All

**Versions**
- Kotlin version*: 1.9.22
- Compose Multiplatform version*: 1.6.0-dev1405 
- OS version(s)* (required for Desktop and iOS issues): macOS Sonoma 14.2.1
- OS architecture (x86 or arm64): arm64
- JDK (for desktop issues): Zulu 17

**To Reproduce**
Define two resources with different type but same name.

**Expected behavior**
The generated `Res.kt` file compiles.

ChristianKatzmann:
I was able to reproduce the issue in ChristianKatzmann/compose-multiplatform@b63a963.

okushnikov:
[...]

PR #4404, opened 2024-03-01
[resources] Init resource accessors lazily and in external function to avoid MethodTooLargeException

terrakok:
fixes https://github.com/JetBrains/compose-multiplatform/issues/4194


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/ResourcesSpec.kt
```diff
@@ -188,42 +188,50 @@ private fun getChunkFileSpec(
...
-                                            add("),\n")
-                                        }
-                                    }
-                                    add(")\n")
-                                }
-                            }
-                            .add(")")
-                            .build().toString()
-                    )
+                    .delegate("\nlazy·{ init_$resName() }")
```
ChristianKatzmann, line 193: Include `type` as well, to prevent something like #4237?

pjBooms (approved):

PR #4410, opened 2024-03-01
Make desktop preview task fully configuration cache compliant

eygraber:
The issue was in a case where there wasn't a direct dependency on the skiko artifact, the task would attempt to get one at task execution time. This moves that to task configuration time.


eygraber:
@igordmn any idea on when this could get looked at?

igordmn:
Thanks! Lgtm.

I changed the description to comply our new guidelines that aren't yet enforced. Does it looks okay?

eygraber:
Description looks ok. Do you want to add that an integration test was added, or is that section only for manual QA testing?

igordmn:
> Do you want to add that an integration test was added, or is that section only for manual QA testing?

Thanks, added. This is for all testing.

PR #4433, opened 2024-03-05
[example] Migrate the image viewer on the new compose resources

terrakok:


igordmn (approved):

examples/imageviewer/desktopApp/build.gradle.kts
```diff
@@ -28,7 +24,7 @@ compose.desktop {
             packageName = "ImageViewer"
             packageVersion = "1.0.0"
```
igordmn, line 25: `examples/imageviewer/shared/src/desktopMain/resources/ic_imageviewer_round.png` is under `resources` instead of `composeResources`.

Is it expected, or it is not possible to have platform-specific resources yet?

terrakok: because the desktop window API requires java resources. it is the same situation as app icon for an android app - system specific thing

igordmn: The window API [accepts](https://github.com/JetBrains/compose-multiplatform/blob/19331a8a02491dd75e2631ba2a301b489b3a620d/examples/imageviewer/shared/src/desktopMain/kotlin/example/imageviewer/view/ImageViewer.desktop.kt#L60) common `Painter`, which is also the type of the resource function `painterResource`.

igordmn: I saw that we moved some icons to `desktop-icons` - they we can't use indeed. But the situation seems different for `ic_imageviewer_round`?

terrakok: we may but don't do. so, I moved it to a platform specific dir

igordmn (approved):

PR #4452, opened 2024-03-08
update compiler version to 1.5.10

shishkin-pavel:


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
```diff
@@ -23,6 +23,7 @@ internal object ComposeCompilerCompatibility {
         "1.9.20" to "1.5.3",
         "1.9.21" to "1.5.4",
         "1.9.22" to "1.5.8.1",
+        "1.9.23" to "1.5.10",
```
eymar, line 26: Also let's udpate here: https://github.com/JetBrains/compose-multiplatform/blob/shish/compiler/release/1.5.10-for-compose-1.6.1/gradle-plugins/gradle.properties#L15-L17

shishkin-pavel: done

gradle-plugins/gradle.properties
```diff
@@ -12,9 +12,9 @@ compose.version=1.6.0-dev1383
 # The latest version of Compose Compiler used by Gradle plugin. Used only in tests/CI.
 compose.tests.compiler.version=1.5.8.1
```
eymar, line 13:
```suggestion
compose.tests.compiler.version=1.5.10
```

eymar (approved):

PR #4490, opened 2024-03-15
Provide a better message for erroneous values folder files

writtmeyer:
The task generateComposeResClass created an error message that didn't help the user to understand the problem if files other than strings.xml were in the values folder. This fix improves on the message so that users can quickly fix the underlying issue.

The old message was "Unknown resource type: 'values'." But not values is the problem, but a file like - in my case - colors.xml which I copied from an existing Android project (well, yeah, I shouldn't have done that, but anyway, the message was not helpful).

The new message would be "Forbidden file name 'colors.xml'! Only a file for string resources named 'strings.xml' is valid in the 'values' folder." I also added a check for this in the ResourcesTest's `testGeneratedAccessors()` method.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResClassTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResClassTask.kt
@@ -96,10 +96,14 @@ internal abstract class GenerateResClassTask : DefaultTask() {
             return null
         }
 
-        if (typeString == "values" && file.name.equals("strings.xml", true)) {
-            val stringIds = getStringIds(file)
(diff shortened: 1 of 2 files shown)


terrakok (approved):

terrakok:
@writtmeyer check the tests, please.

writtmeyer:
I have serious trouble using [act](https://github.com/nektos/act) to run this as an action locally on my machines (Mac and Ubuntu). And running the tests itself on both systems doesn't run into these failures. And as far as I see it only the test log, but not the log of the specific failing gradle runs executed within the failing test method are available. Thus I'm really not sure as to what the reason for these failures is.

terrakok:
Since any xml files were supported in the values directory I will close the issue

PR #4519, opened 2024-03-20
Support plural string resource

paxbun:
Ports a part of Unicode's ICU in pure Kotlin and implements Android-style plural string resource support. Fixes JetBrains/compose-multiplatform#425.

# Changes

- Added `org.jetbrains.compose.resources.intl.{PluralCategory, PluralRule, PluralRuleList}`, which parses and evaluates scripts in Unicode's Locale Data Markup Langauge.
- Copied `plurals.xml` from Unicode's [CLDR](https://github.com/unicode-org/cldr/blob/release-44-1/common/supplemental/plurals.xml).
- Added `GeneratePluralRuleListsTask`, which parses `plurals.xml` and generates required Kotlin source codes.
- Added `PluralStringResource`, `pluralStringResource`, or `getPluralString`, corresponding to `StringResource`, `stringResource`, or `getString`.
- Modified `ResourcesSpec.kt` so the generated `Res` class exposes `Res.plurals`.

# Potential Further Improvements

- [ ] Allow configuring the default language in the `compose.resources {}` block (#4482) to determine the default pluralization rule (or just presume English as default)
- [ ] Move the parser logic to the Gradle plugin and generate pluralization rules in `Res` only for languages used in `composeResources`


terrakok:
@paxbun Thank you for the great contribution! I made a backward PR to your branch with some refactorings. And I would like to ask to change a logic a bit: not to crash an app when user hasn't provided any pluralization:
```kotlin
kotlin.IllegalStateException: String ID=`new_message` does not have the pluralization MANY for quantity 0!
```
I guess it's better to use a some default and print a warning to the log. what do you think?

paxbun:
@terrakok Thanks for the comment! I merged your changes and renamed the properties of GeneratePluralRuleListsTask (cause they're now files, not directories).

> I guess it's better to use a some default and print a warning to the log. what do you think?

[The current implementation of Android SDK](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/base/core/java/android/content/res/ResourcesImpl.java;l=339-341?q=ResourcesImpl.java) also throws an Exception when there is no proper pluralization, although mine did not fallback to `other` when the right one was not found (I've just added the fallback logic).

We can return an empty string, but in that case we also have to care about ones with format arguments because `String.replaceWithArgs` will throw an exception for an empty string.

terrakok (approved):

terrakok:
I will do it manually

terrakok:
BTW, you forgot to `handleSpecialCharacters` 
```kotlin
val pluralCategory = PluralCategory.fromString(
    element.getAttribute("quantity"),
) ?: return@mapNotNull null
val rawString = element.textContent.orEmpty()
pluralCategory to handleSpecialCharacters(rawString) // <--- here

```
could you fix it?

paxbun:
Of course! Here is the PR as you requested: #4543

PR #4529, opened 2024-03-22
update compose compiler version for k2.0.0-Beta5

shishkin-pavel:


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
```diff
@@ -26,6 +26,7 @@ internal object ComposeCompilerCompatibility {
         "1.9.23" to "1.5.10.1",
         "2.0.0-Beta1" to "1.5.4-dev1-kt2.0.0-Beta1",
         "2.0.0-Beta4" to "1.5.9-kt-2.0.0-Beta4",
+        "2.0.0-Beta4" to "1.5.11-kt-2.0.0-Beta5",
```
igordmn, line 29:
```suggestion
        "2.0.0-Beta5" to "1.5.11-kt-2.0.0-Beta5",
```
is it correct?

igordmn (approved):

PR #4576, opened 2024-04-04
Get resource files as URI

terrakok:
Adds a public `Res.getUri(path: String): String` function.
It lets external libraries a way to read resource files by a platform dependent Uri.
E.g.: video players, image loaders or embedded web browsers.

```kotlin
val uri = Res.getUri("files/my_video.mp4")
```

fixes https://github.com/JetBrains/compose-multiplatform/issues/4360

### DIFF
--- a/gradle-plugins/compose/src/test/test-projects/misc/jvmOnlyResources/expected/Res.kt
+++ b/gradle-plugins/compose/src/test/test-projects/misc/jvmOnlyResources/expected/Res.kt
@@ -3,12 +3,13 @@
     org.jetbrains.compose.resources.ExperimentalResourceApi::class,
 )
 
-package me.app.jvmonlyresources.generated.resources
+package app.group.resources_test.generated.resources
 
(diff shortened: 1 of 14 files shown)


gradle-plugins/compose/src/test/test-projects/misc/jvmOnlyResources/expected/Res.kt
```diff
@@ -23,6 +24,16 @@ internal object Res {
...
 
+    /**
+     * Returns the URI string of the resource file at the specified path.
+     *
+     * Example: `val uri = Res.getUri("files/key.bin")`
+     *
+     * @param path The path of the file in the compose resource's directory.
+     * @return The URI string of the file.
+     */
+    public fun getUri(path: String): String = getResourceUri("" + path)
```
pjBooms, line 35: Should not we add test that calls getUri (we may check that resulting URI contains original path)?

pjBooms (approved):

PR #4600, opened 2024-04-10
drop old desktop-only resource API for the new multiplatform resource API

elect86:
Credits to @m-sasha for the help


m-sasha:
Thanks for the PR, but we'll update the tutorial ourselves when the new resources API is considered stable.

PR #4623, opened 2024-04-16
Delete the experimental mark from the stable resources library API

terrakok:
And remove internal testing methods


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/Resource.kt
```diff
@@ -14,7 +14,6 @@ annotation class InternalResourceApi
  * @property id The ID of the resource.
  * @property items The set of resource items associated with the resource.
  */
-@ExperimentalResourceApi
```
igordmn, line 17: Not for this PR, but because we are removing experimental, we have to support API backward compatibility, and better to apply https://github.com/Kotlin/binary-compatibility-validator plugin for that

igordmn (approved):
LGTM. Needs a second review before the merge.

PR #4628, opened 2024-04-16
Hot fix integration with cocoapods

terrakok:
Get the cocoapods extension from the kotlin instead project object and configure task dependency lazy


igordmn (approved):

MatkovIvan (commented):
Good for hotfix, but I'd request some test case (as separate PR) for this to avoid future regressions.

PR #4631, opened 2024-04-16
[gradle] Add integration test for compose resources with cocoapods

terrakok:


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -532,4 +533,100 @@ class ResourcesTest : GradlePluginTestBase() {
             check.logContains("Generation Res class is disabled")
         }
     }
+
+    @Test
+    fun iosResources() {
+        Assumptions.assumeTrue(currentOS == OS.MacOS)
```
igordmn, line 539: Because of the recent issue can we write a test that have cocoapods and test the usual `run` task on all systems?

terrakok: I don't think we have to. Because the behavior is tested on the KGP side.
We don't have platform specific logic here

igordmn (approved):

PR #4651, opened 2024-04-18
[resources] Add functions to retrieve bytes from drawable or font resources.

terrakok:
Implemented two new experimental functions:
```kotlin
/**
 * Retrieves the byte array of the drawable resource.
 *
 * @param environment The resource environment, which can be obtained from [rememberResourceEnvironment] or [getSystemResourceEnvironment].
 * @param resource The drawable resource.
 * @return The byte array representing the drawable resource.
 */
@ExperimentalResourceApi
suspend fun getDrawableResourceBytes(
    environment: ResourceEnvironment,
    resource: DrawableResource
): ByteArray {...}

/**
 * Retrieves the byte array of the font resource.
 *
 * @param environment The resource environment, which can be obtained from [rememberResourceEnvironment] or [getSystemResourceEnvironment].
 * @param resource The font resource.
 * @return The byte array representing the font resource.
 */
@ExperimentalResourceApi
suspend fun getFontResourceBytes(
    environment: ResourceEnvironment,
    resource: FontResource
): ByteArray {...}
```

fixes https://github.com/JetBrains/compose-multiplatform/issues/4360


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
```diff
@@ -110,6 +109,23 @@ private fun svgPainter(resource: DrawableResource): Painter {
...
 
+/**
+ * Retrieves an ByteArray using the specified drawable resource.
+ *
+ * @param resource The drawable resource to be used.
+ * @return The ByteArray loaded from the resource.
+ */
+@ExperimentalResourceApi
+@Composable
+fun getDrawableResourceBytes(resource: DrawableResource): ByteArray {
```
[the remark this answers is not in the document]
terrakok: But we don't try to return resource we return bytearray. and it is user's responsibility to handle an empty array. no?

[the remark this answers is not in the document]
igordmn: > and it is user's responsibility to handle an empty array. no?

In our case yes, but it is not obvious by the name of the function and its behaviour on non-web platforms that user should do that. It looks like an usual function that returns a valid byte array straight away - in such function user shouldn't check an empty value.

So, to show the user that they should handle this case, better to represent it in a type system, so user can know about the special case in compilation time.

Besides, what if we add some image format, zero byte array for which is actually a valid value? (I didn't check, maybe it is valid for BMP?) We would come to a situation when the user can't do anything with this empty value, as they skip zero-bytes values, handling the loading case.

Anyway, let's discuss the non-composable option below, it doesn't require handling a special case.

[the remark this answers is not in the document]
terrakok: Yes, we don't provide our internal environment classes outside the library, because it seems should to be a part of runtime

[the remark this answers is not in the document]
igordmn: > it seems should to be a part of runtime

Ability to change environment seem belong to runtime, yes.

But I am not sure about the class itself. The runtime environment class won't contain everything that is needed for resources (or it won't have a separate class for it, just a bunch of CompositionLocal's). Also, resurces library can depend on extra environment, that isn't handled by Compose runtime.

Let's discuss over a call.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/FontResources.kt
```diff
@@ -32,4 +33,20 @@ expect fun Font(
...
+/**
+ * Retrieves the byte array of the font resource.
+ *
+ * @param environment The optional resource environment.
+ * @param resource The font resource.
+ * @return The byte array representing the font resource.
+ */
+@ExperimentalResourceApi
+suspend fun getFontResourceBytes(
+    environment: ResourceEnvironment = getResourceEnvironment(),
```
igordmn, line 47: Having `environment` on the first place forces users to use `resource` name explicitly. Is that was the intention?

I don't have a strong opinion here. On one hand, it is better for users to make a conscience choice to use the system environment, on the other hand it is verbose.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/FontResources.kt
```diff
@@ -32,4 +33,20 @@ expect fun Font(
     resource: FontResource,
```
igordmn, line 33: It is late and not for this PR, but `Font` function isn't by Compose guidelines.

If it returns a value, it should start with lower case.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/StringArrayResources.kt
```diff
@@ -48,6 +48,20 @@ fun stringArrayResource(resource: StringArrayResource): List<String> {
...
+/**
+ * Loads a list of strings using the specified string array resource.
+ *
+ * @param environment The resource environment.
+ * @param resource The string array resource to be used.
+ * @return A list of strings representing the items in the string array.
+ *
+ * @throws IllegalStateException if the string array with the given ID is not found.
+ */
+suspend fun getStringArray(
```
igordmn, line 60: To be consistent with other types, maybe we need `suspend fun vectorResource(ResourceEnvironment`, `suspend fun imageResource(ResourceEnvironment`, `suspend fun font(ResourceEnvironment`

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/FontResources.kt
terrakok: It's done to mirror the Android approach:
```kotlin
val firaSansFamily = FontFamily(
    Font(R.font.firasans_light, FontWeight.Light),
    Font(R.font.firasans_regular, FontWeight.Normal),
    Font(R.font.firasans_italic, FontWeight.Normal, FontStyle.Italic),
    Font(R.font.firasans_medium, FontWeight.Medium),
    Font(R.font.firasans_bold, FontWeight.Bold)
)
```

https://developer.android.com/develop/ui/compose/text/fonts

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/FontResources.kt
terrakok: Yes, it was done to avoid parameters after a vararg in
```kotlin
suspend fun getString(
    environment: ResourceEnvironment,
    resource: StringResource,
    vararg formatArgs: Any
): String
```

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/FontResources.kt
igordmn: `FontFamily` and `Font` aren't Composable functions, they are data classes that have a link to a resource. This link is resolved later, when it is used inside Composable function.

If we want to maintain the Compose API and maintain Compose guidelines at the same time, we need to do the same.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/StringArrayResources.kt
terrakok: I wouldn't add public APIs without explicit needs.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/PluralStringResources.kt
```diff
@@ -99,9 +115,32 @@ suspend fun getPluralString(resource: PluralStringResource, quantity: Int, varar
...
+ *
+ * @param environment The resource environment.
+ * @param resource The string resource to be used.
+ * @param quantity The quantity of the pluralization to use.
+ * @param formatArgs The arguments to be inserted into the formatted string.
+ * @return The loaded string resource.
+ *
+ * @throws IllegalArgumentException If the provided ID or the pluralization is not found in the resource file.
+ */
+suspend fun getPluralString(
```
igordmn, line 132:
```suggestion
@ExperimentalResourceApi
suspend fun getPluralString(
```

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/PluralStringResources.kt
```diff
@@ -43,7 +43,23 @@ fun pluralStringResource(resource: PluralStringResource, quantity: Int): String
...
+ * Loads a string using the specified string resource.
+ *
+ * @param environment The resource environment.
+ * @param resource The string resource to be used.
+ * @param quantity The quantity of the pluralization to use.
+ * @return The loaded string resource.
+ *
+ * @throws IllegalArgumentException If the provided ID or the pluralization is not found in the resource file.
+ */
+suspend fun getPluralString(
```
igordmn, line 58:
```suggestion
@ExperimentalResourceApi
suspend fun getPluralString(
```

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/FontResources.kt
igordmn: In this case, I would suggest to do the same for other functions - split them into 2:
```
@ExperimentalResourceApi
suspend fun getFontResourceBytes(
    environment: ResourceEnvironment,
    resource: FontResource
): ByteArray

@ExperimentalResourceApi
suspend fun getFontResourceBytes(
    environment: ResourceEnvironment,
): ByteArray = getFontResourceBytes(getSystemResourceEnvironment(), resource)
```

Otherwise we can't write `getDrawableResourceBytes(Res.drawable.title)`, only `getDrawableResourceBytes(resource = Res.drawable.title)`

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
```diff
@@ -5,11 +5,12 @@ import androidx.compose.runtime.*
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.text.intl.Locale
 
-internal data class ResourceEnvironment(
-    val language: LanguageQualifier,
-    val region: RegionQualifier,
-    val theme: ThemeQualifier,
-    val density: DensityQualifier
+@ExperimentalResourceApi
+data class ResourceEnvironment internal constructor(
```
MatkovIvan, line 9: [Don't use data classes in an API﻿](https://kotlinlang.org/docs/jvm-api-guidelines-backward-compatibility.html#don-t-use-data-classes-in-an-api)

MatkovIvan (approved):

PR #4682, opened 2024-04-23
ci: update Dockerfile used to run web tests

kropp:
update to the latest LTS and add Xvfb to run browser in non-headless mode


eymar:
Thank you!

Issue #4197, opened 2024-01-29
Provide a way to programmatically change the local `ResourceEnvironment` (`*Qualifier`)


AzimMuradov:
I need to have an ability to programmatically switch between sets of resources (based on their qualifiers). I need this feature to provide my users a way for changing the app language/interface size/theme/etc.

Fortunately, it should be quite easy to implement. The only thing needed is to make [`LocalComposeEnvironment`](https://github.com/JetBrains/compose-multiplatform/blob/master/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt#L42) `public` and maybe add some utility functions (`WithLanguage`/`WithRegion`/...).

Basically, I want to be able to programmatically and locally change `ResourceEnvironment`:

```kotlin
internal data class ResourceEnvironment(
    val language: LanguageQualifier,
    val region: RegionQualifier,
    val theme: ThemeQualifier,
    val density: DensityQualifier
)
```

If it's ok as a feature, I'm more than happy to help!

terrakok:
You can override the density via the CompositionLocalProvider
```
CompositionLocalProvider(
    LocalDensity provides Density(1.5f)
) {
    //compose
}
```

for the other environment values the Compose doesn't provide providers yet.
this should be implemented in the Compose and the library will use it

ryanmitchener:
I just wanted to voice my support for this enhancement. This is particularly helpful for allowing users to select their language in an application. I'm not sure if Compose will ever implement support for this because Android has alternative methods of changing the language from inside the application via configurations and/or per-app language preferences. Those existing methods do not work for iOS or Desktop and configuring the application language in-app is more prevalent on Desktop vs Android/iOS.

At the very least it would be helpful to add a function that allows passing in a language qualifier or something similar. That would allow 3rd party libraries to wrap the resources library and work around this limitation.

```kotlin
fun stringResource(resource: StringResource, language: LanguageQualifier)
```

juhaodong:
Very Important for us!

aniluMango:
We have an existing ios app. which is created with custom localization handling. that not fitted in compose resources

N7ghtm4r3:
Hi, I tried to use `Locale.setDefault(Locale.forLanguageTag("tag"))` and on `Desktop` works correctly, obviuosly using the  
`composeResources` api with the specific language tag for values folder (values-en l, etc...)
I don't know if this can work on mobile platforms or on web also

[...]

PR #4703, opened 2024-04-26
overridable resource locale

psuzn:
Add a way to override local for the resources (partially address #4197).

Similar to how we override the density with the LocalDensity composition provider it adds LocalResourceLocale to override Locale for Resource. 

**Usage**

```kotlin
 CompositionLocalProvider(LocalResourceLocale provides Locale("np")) {
  Text(stringResource(Res.string.hello))
}
```

### DIFF
--- a/components/resources/demo/shared/src/commonMain/composeResources/values-np/strings.xml
+++ b/components/resources/demo/shared/src/commonMain/composeResources/values-np/strings.xml
@@ -1,0 +1,3 @@
+<resources>
+    <string name="hello">😊 नमस्कार संसार!</string>
+</resources>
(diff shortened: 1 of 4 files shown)


psuzn:
@terrakok, can you review this, please?

toaster217:
Will this be available in 1.6.10?

MatkovIvan:
> Will this be available in 1.6.10?

No

kaidotarma:
Would love to see this merged. :)

kaidotarma:
@terrakok @MatkovIvan Are there any possible issues this change might cause? 
If not, it would great to have the possibility to handle the app language in Compose.
I would be more than happy to help move this forward.

terrakok:
The feature should be implemented in the core/runtime modules, not in the library. Because it is a more general thing. You may track the progress here: https://github.com/JetBrains/compose-multiplatform/issues/4347
Sorry for the long reply

PR #4746, opened 2024-05-01
Add jetsnack app to examples

JSMonk:


Schahen (approved):
Thank you fro porting!

PR #4783, opened 2024-05-08
Fix resources with cocoapods integration

terrakok:
The PR sets a static path to the compose resources for the cocoapods integration because there maybe only one integration framework for a gradle module.

Fixes https://github.com/JetBrains/compose-multiplatform/issues/4720

### Fixes - Resources
- _(prerelease fix)_ Fix resources with cocoapods integration

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/IosResources.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/IosResources.kt
@@ -81,19 +81,24 @@ internal fun Project.configureSyncIosComposeResources(
     plugins.withId(COCOAPODS_PLUGIN_ID) {
         (kotlinExtension as ExtensionAware).extensions.getByType(CocoapodsExtension::class.java).apply {
             framework { podFramework ->
-                val syncDir = podFramework.getFinalResourcesDir().get().asFile.relativeTo(projectDir)
-                val specAttr = "['${syncDir.path}']"
-                extraSpecAttributes["resources"] = specAttr
-                project.tasks.named("podInstall").configure {
-                    it.doFirst {
-                        if (extraSpecAttributes["resources"] != specAttr) {
-                            error("""
+                // DIRTY HACK: we don't have an API to configure PodspecTask.
+                // We need to use the cocoapods DSL only but to provide a correct resources path we need lazy read a framework baseName.
+                // Which may be configured be the same DSL
+                podFramework.project.afterEvaluate {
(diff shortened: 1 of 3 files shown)


igordmn (approved):
Thanks!

Please add Release Notes

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/IosResources.kt
```diff
@@ -81,19 +81,24 @@ internal fun Project.configureSyncIosComposeResources(
...
-                val specAttr = "['${syncDir.path}']"
-                extraSpecAttributes["resources"] = specAttr
-                project.tasks.named("podInstall").configure {
-                    it.doFirst {
-                        if (extraSpecAttributes["resources"] != specAttr) {
-                            error("""
+                // DIRTY HACK: we don't have an API to configure PodspecTask.
+                // We need to use the cocoapods DSL only but to provide a correct resources path we need lazy read a framework baseName.
+                // Which may be configured be the same DSL
+                podFramework.project.afterEvaluate {
```
MatkovIvan, line 87: I'd extract this apply/framework/afterEvaluate chain to separate function to localize this "hack" there and reduce nesting blocks here.
Just suggestion/nitpick, not blocking

MatkovIvan (approved):
Moving cocoapod configuration to `afterEvaluate` looks valid to me

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/IosResources.kt
```diff
@@ -113,7 +113,7 @@ private fun Framework.isCocoapodsFramework() = name.startsWith("pod")
 private fun Framework.getFinalResourcesDir(): Provider<Directory> {
     val providers = project.providers
     return if (isCocoapodsFramework()) {
-        project.layout.buildDirectory.dir("compose/ios/$baseName/$IOS_COMPOSE_RESOURCES_ROOT_DIR/")
+        project.layout.buildDirectory.dir("compose/cocoapods/$IOS_COMPOSE_RESOURCES_ROOT_DIR/")
```
MatkovIvan, line 116: Does it work for case with multiple pods/frameworks now?

terrakok: we cannot produce more then one a framework from a gradle module with a podspec

igordmn (approved):

MatkovIvan (approved):

PR #4796, opened 2024-05-10
Update gradle plugin: Deprecate experimental configuration for Compose for Web

eymar:
Since 1.6.10 Compose for Web goes to Alpha and experimental configuration is not needed anymore. We'll configure the web targets by default when they're added to projects.

## Testing
- I built the plugin to mavenLocal. And used it in a couple of our samples.
- After gradle sync completes, I observe the Deprecation warning in build.gradle.kts on `compose.experimental.web` usages (see a screenshot below)
Screenshot 2024-05-10 at 15 41 14

Then I remove deprecated API usages and test that the project works without it:
- run `./gradlew clean` just in case
- run the app: `./gradlew wasmJsBrowserRun` and `./gradlew jsBrowserRun` - both run fine
- build the production distribution: `./gradlew wasmJsBrowserDistribution`
- `cd ..../build/dist/wasmJs/productionExecutable`  and run `python -m http.server`, open a browser at `http://localhost:8000` - the app should work the same way it works with gradle tasks above

This should be tested by QA

### Highlights - Web
- Compose for Web goes to Alpha!
Some experimental Compose Multiplatform Gradle plugin APIs for web app configuration were deprecated. Their usage is not required anymore.


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/experimental/internal/checkExperimentalTargets.kt
```diff
@@ -27,7 +27,6 @@ private val TargetType.gradlePropertyName get() = "org.jetbrains.compose.experim
 private val EXPERIMENTAL_TARGETS: Set<TargetType> = setOf(
     TargetType("macos", presets = listOf("macosX64", "macosArm64")),
     TargetType("jscanvas", presets = listOf("jsIr", "js")),
-    TargetType("wasm", presets = listOf("wasm", "wasmJs")),
```
eymar, line 30: Note: I keep jscanvas above! 
We'll try to support k/js at least until Safari enables WasmGC. But sill it's better to keep it as an experimental target for now.

Schahen (approved):
Thank you for taking care of this!

PR #4819, opened 2024-05-16
[resources] Check cached deferreds and drop them if they are cancelled.

terrakok:
Before the fix we could cancel a coroutine and the cancelled deferred was saved in cache.

### Fixes - Resources
- _(prerelease fix)_ Fix a cached empty resource on a Compose for Web if the resource loading was canceled during progress

### DIFF
--- a/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/AsyncCache.kt
+++ b/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/AsyncCache.kt
@@ -1,0 +1,31 @@
+package org.jetbrains.compose.resources
+
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.async
+import kotlinx.coroutines.coroutineScope
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+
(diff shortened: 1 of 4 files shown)


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/AsyncCache.kt
```diff
@@ -0,0 +1,31 @@
...
+        val deferred = mutex.withLock {
+            var cached = cache[key]
+            if (cached == null || cached.isCancelled) {
+                //LAZY - to free the mutex lock as fast as possible
+                cached = async(start = CoroutineStart.LAZY) { load() }
+                cache[key] = cached
+            }
+            cached
+        }
+        deferred.await()
```
igordmn, line 24: We still have an issue when 2 composables starts to load simultaneously:

1. one starts to load
2. the second awaits
3. we exit from the first one
4. the second cancels

igordmn: To solve the issue, we probably need this:
1. use global scope
2. count ref count
3. cancel if ref count is zero

igordmn (approved):
Agreed that with current state, the bug isn't realistic, but will look at it in master branch

PR #4824, opened 2024-05-17
CfW: Prevent UnpackSkikoWasmRuntime task execution when it's not needed

eymar:
Fixes: https://github.com/JetBrains/compose-multiplatform/issues/4823

In https://github.com/JetBrains/compose-multiplatform/pull/4796 we intentionally started to configure the web app for all k/js and k/wasm targets.  The configuration involves adding a dependency on skiko-wasm runtime and unpacking it. 
Some projects don't need skiko-wasm-runtime (like those based on compose.html or just compose.runtime). 

**Solution:**
We check if there is a dependency on org.jetbrains.compose.ui libraries (including transitive dependencies). If we find it, then we enable skikoUnpack task. Otherwise it's disabled.

## Testing
- Build the gradle plugin locally (with this change)
- Used it in our html landing example: https://github.com/JetBrains/compose-multiplatform/blob/master/examples/html/landing
- Run `./gradlew jsBrowserDistribution`, check `.../compose-multiplatform/examples/html/landing/build/dist/js/productionExecutable` and see NO skiko.* files added there
- Then add `implementation(compose.foundation)` dependency in build.gradle.jts and run `./gradlew clean jsBrowserDistribution` again - the build/dist contains skiko.* now

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
@@ -31,6 +31,7 @@ internal fun Project.configureWeb(
 internal fun Collection<KotlinJsIrTarget>.configureWebApplication(
     project: Project
 ) {
+    if (project.properties["org.jetbrains.compose.web.skipUnpackSkiko"] == "true") return
     val skikoJsWasmRuntimeConfiguration = project.configurations.create("COMPOSE_SKIKO_JS_WASM_RUNTIME")
     val skikoJsWasmRuntimeDependency = skikoVersionProvider(project).map { skikoVersion ->
         project.dependencies.create("org.jetbrains.skiko:skiko-js-wasm-runtime:$skikoVersion")
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
@@ -23,15 +23,35 @@ internal fun Project.configureWeb(
     composeExt: ComposeExtension,
 ) {
     val webExt = composeExt.extensions.getByType(WebExtension::class.java)
+
+    // here we check all dependencies (including transitive)
+    // If there is compose.ui, then skiko is required!
+    val shouldRunUnpackSkiko = project.provider {
+        var dependsOnComposeUi = false
+        project.configurations.matching {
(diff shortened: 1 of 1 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -23,13 +23,34 @@ internal fun Project.configureWeb(
...
+            val isWasmOrJs = it.name.contains("js", true) ||
+                    it.name.contains("wasm", true)
+
+            it.isCanBeResolved && isWasmOrJs
+        }.all {
+            val match = it.incoming.artifacts.resolvedArtifacts.get().any {
+                it.id.componentIdentifier.toString().contains("org.jetbrains.compose.ui")
+            }
+
+            dependsOnComposeUi = dependsOnComposeUi || match
```
terrakok, line 41: I'd ask use explicit parameters instead of 'it'

terrakok (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
eymar: done

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -23,13 +23,34 @@ internal fun Project.configureWeb(
...
+    val shouldRunUnpackSkiko = project.provider {
+        var dependsOnComposeUi = false
+        project.configurations.matching { configuration ->
+            val isWasmOrJs = configuration.name.contains("js", true) ||
+                    configuration.name.contains("wasm", true)
+
+            configuration.isCanBeResolved && isWasmOrJs
+        }.all { configuration ->
+            val match = configuration.incoming.artifacts.resolvedArtifacts.get().any { artifact ->
+                artifact.id.componentIdentifier.toString().contains("org.jetbrains.compose.ui")
```
igordmn, line 38: it also includes "ui-unit". can we check the exact match with ui module?

igordmn (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
eymar: done

igordmn (approved):

PR #4845, opened 2024-05-22
Update Compose for Web status in README.md

artemyto:
Seems like Compose for Web is in Alpha now: https://github.com/JetBrains/compose-multiplatform/pull/4796


MatkovIvan:
It's done in #4857

PR #4893, opened 2024-05-29
[resources] Use cached value synchronously on web.

terrakok:
The change speeds resources web rendering up by the reading a cached value instantly by request (it was being dispatched to the end of the UI queue in `LaunchedEffect`)

### Features - Resources
- Speed resources web rendering up by the reading a cached value instantly


igordmn:
> Highlights - Resources

Better just "Features - Resources"

PR #4894, opened 2024-05-29
[resources] Improve efficiency of font caching in Skiko

terrakok:
Introduced a cache containing previously read fonts from Skiko. This improvement prevents re-reading font files already cached, instead providing an empty byte array to enhance performance. This change is especially beneficial for repeated calls to the same font resources.

## Testing
 - Run the compose resources web sample
 - Open a network log
 - Few times open and close a tab with font samples
 - Check that fonts were loaded only once

### Highlights - Resources
- Reuse cached fonts in Skiko

### DIFF
--- a/components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
+++ b/components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
@@ -4,6 +4,8 @@ import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.text.font.*
 import androidx.compose.ui.text.platform.Font
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import kotlin.io.encoding.Base64
 import kotlin.io.encoding.ExperimentalEncodingApi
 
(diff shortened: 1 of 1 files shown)


components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
```diff
@@ -29,12 +31,27 @@ private val emptyFontBase64 =
...
     val resourceReader = LocalResourceReader.current
     val fontFile by rememberResourceState(resource, weight, style, { defaultEmptyFont }) { env ->
         val path = resource.getResourceItemByEnvironment(env).path
-        val fontBytes = resourceReader.read(path)
+
+        // Skiko has own internal font's cache,
+        // so we can provide an empty byte array since the font was cached already
+        val fontBytes = cacheMutex.withLock {
+            if (fontsCachedBySkiko.contains(path)) {
+                ByteArray(0)
```
igordmn, line 47: I looked at the sources, and `getData` is called by Compose (not Skia) cache [here](https://github.com/JetBrains/compose-multiplatform-core/blob/f9b90c58dfb166be398917f1e43326f6566e50ff/compose/ui/ui-text/src/skikoMain/kotlin/androidx/compose/ui/text/platform/PlatformFont.skiko.kt#L211). It uses `ExpireAfterAccessCache` with 1-minute expiration date.

So, we can't guarantee that this data is stored forever.

igordmn (changes requested):

PR #4909, opened 2024-05-31
[gradle] Generate method to create drawable or font resource accessors by a file path.

terrakok:
Now we can get resource accessors by the path.

```kotlin
Icon(
  painterResource(Res.drawable.byPath("drawable/my_icon.xml")),
  contentDescription = null
)
```

The new API doesn't work with qualifiers and environment. It associates a concrete file with a new resource instance.
If there are two icons: `drawable-night/my_icon.xml` and `drawable-light/my_icon.xml`, users are supposed to select a right icon by their own.

It doesn't work for strings resources. There is no a such thing as a full qualified key for the string in a text representation. (as path for other resource types) We need to combine an xml file path + key inside. It is not what we want.

I understand that someone would like to have something like a runtime search in their string resources based on a string key and a current environment but it is a huge performance problem. That's why we converted XML files to the internal format and use generated classes instead. To iterate by files in the runtime is not possible.

Note: there is a way to generate a special map: `key -> value` and use for the search but it requires more resources to maintain it then brings a profit.

So, for the case it's possible to save strings to the regular txt file and read it as string on app's side.

Fixes https://github.com/JetBrains/compose-multiplatform/issues/4880

### Features - Resources
- Generate method to create drawable or font resource accessors by a file path.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
@@ -134,13 +134,13 @@ internal fun getResFileSpec(
                 FunSpec.builder("readBytes")
                     .addKdoc(
                         """
-                    Reads the content of the resource file at the specified path and returns it as a byte array.
-                    
-                    Example: `val bytes = Res.readBytes("files/key.bin")`
-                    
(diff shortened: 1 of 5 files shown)


terrakok:
After an internal discussion we agreed that the right way is to add utility functions to convert ByteArrays to ImageBitmap/XmlElement/SvgElement in the compose core library.

https://github.com/JetBrains/compose-multiplatform/issues/4925

To use strings as keys for resources is different story and will be discussed separately.

Issue #4938, opened 2024-06-07
Compose Resource incorrectly reads string chunks from a file  


AChep:
**Describe the bug**
App crashes during base64 decoding of the resources. 

![image](https://github.com/JetBrains/compose-multiplatform/assets/990982/96fc0290-b44e-4f92-bf4d-50410b9f4e18)

![image](https://github.com/JetBrains/compose-multiplatform/assets/990982/ce3e0d49-e178-446f-ab97-084abd944862)

If I go to the file itself and read the given range
```
> dd if=strings.commonMain.cvr ibs=1 skip=23401 count=83
string|feat_item_show_barcode_text|RW5jb2RlIGEgdGV4dCBhcyBkaWZmZXJlbnQgYmFyY29kZXMu
83+0 records in
0+1 records out
83 bytes copied, 7.6769e-05 s, 1.1 MB/s
```

we can see that the range is correct. However the record data is different to what Compose resources reads:
```
RW5jb2RlIGEgdGV4dCBhcyBkaWZmZXJlbnQgY
RW5jb2RlIGEgdGV4dCBhcyBkaWZmZXJlbnQgYmFyY29kZXMu
```

if we investigate further we see that the `ResourceReader.desktop.kt` has read the following sequence of bytes:

```
[115, 116, 114, 105, 110, 103, 124, 102, 101, 97, 116, 95, 105, 116, 101, 109, 95, 115, 104, 111, 119, 95, 98, 97, 114, 99, 111, 100, 101, 95, 116, 101, 120, 116, 124, 82, 87, 53, 106, 98, 50, 82, 108, 73, 71, 69, 103, 100, 71, 86, 52, 100, 67, 66, 104, 99, 121, 66, 107, 97, 87, 90, 109, 90, 88, 74, 108, 98, 110, 81, 103, 89, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

if I pause there and execute two different commands:
```kotlin
String(getResourceAsStream(path).use { it.readAllBytes() }).slice(offset.toInt()..(offset+size).toInt())
// prints string|feat_item_show_barcode_text|RW5jb2RlIGEgdGV4dCBhcyBkaWZmZXJlbnQgYmFyY29kZXMu which is correct
String(getResourceAsStream(path).use { it.skipNBytes(offset); it.readNBytes(size.toInt()) })
[...]

PR #4939, opened 2024-06-07
fix(Resources): Use readNBytes() to read a chunk of resources  #4938

AChep:
Fixes https://github.com/JetBrains/compose-multiplatform/issues/4938 The idea is that both `.skip()` and `.read()` may skip and read less bytes that requested.

Sections:
- Fixes

Subsections:
- Desktop
- Resources


terrakok:
Thank you for the fix. I did it by my own for both targets: android + jvm.
https://github.com/JetBrains/compose-multiplatform/pull/4943

PR #4943, opened 2024-06-07
[resources] Read exactly requested count of bytes from InputStream on jvm platforms.

terrakok:
In some cases the skip and read methods may handle less bytes then expected. The PR fixes it by proper API on the JVM and manual check on the Android.

Fixes https://github.com/JetBrains/compose-multiplatform/issues/4938

## Testing
I manually checked it on the project from the issue.

### Fixes - Resources
- Read exactly requested count of bytes from InputStream on jvm platforms.


AChep:
You should check for the end of a file in the Android implementation. Currently if you request to read N bytes, but there are less than that it will get stuck in the loop.

AChep:
The changes in other Compose files are also probably not needed 😃

igordmn (approved):

PR #4957, opened 2024-06-11
File associations

### LINKED ISSUE #773: Desktop: Add support for file associations
## Feature Request

The current desktop packaging tasks do not include options for setting up [file associations](https://docs.oracle.com/en/java/javase/16/jpackage/support-application-features.html#GUID-8668A806-8A80-435F-970F-7B2BF65863E4), a particularly valuable feature for desktop apps. 

## Possible Workaround

The `AbstractJPackageTask` allows us to add custom args with the `freeArgs` property, which get passed-through directly to `jpackage`. So to add our file associations, we should be able to follow the documentation linked above and manually add the args we need.

```kotlin
// build.gradle.kts
plugins {
    kotlin("multiplatform")
    id("org.jetbrains.compose")
}
compose.desktop {
    application {
		...
    }
}

tasks.withType<AbstractJPackageTask>().all {
    // add custom args to jpackage to set up file associations
    freeArgs.add("--file-associations")
    freeArgs.add(project.projectDir.resolve("ext.properties"))
}
```

## macOS Problems

The above workaround does not seem to work on macOS (I have not tried it on other OSs). macOS handles file associations differently than Windows or Linux, and has a particularly verbose "registration" process that involves adding a bunch of additional information to the package's `Info.plist` file. Normally, `jpackage` would add all this info to the `Info.plist` file itself in accordance with the `--file-associations` property, but the Compose for Desktop Gradle plugin is [writing that file itself](https://github.com/JetBrains/compose-jb/blob/master/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/InfoPlistBuilder.kt). This means that even if we add those flags manually using the `freeArgs` property, the application will still not be registered to open custom file types, because the Compose Gradle plugin is not adding the necessary properties to the plist it writes.

In addition, the CfD plugin adds `freeArgs` before it configures itself, and it specifies `--resource-dir` for its own use when writing the `Info.plist`, so we cannot [provide our own resource dir](https://docs.oracle.com/en/java/javase/16/jpackage/override-jpackage-resources.html#GUID-1B718F8B-B68D-4D46-B1DB-003D7729AAB6) to write the plist by hand ourselves either.

## Solution

Rather than writing the `Info.plist` file within the CfD Gradle plugin, it should let `jpackage` do it. 

## Potential Issues

There may already be info CfD is manually adding to `Info.plist` that `jpackage` is not adding (which may have been the reason for manually writing it in the first place). Removing the step to manually write the plist may cause issues with some of the existing Gradle plugin properties, and deeper research into customizing `jpackage` usage or other workarounds may need to be added to fully support them.

zhelenskiy:
Add file associations support to Compose Desktop

Fixes #773

## Testing

Tested on the [sample project](https://github.com/zhelenskiy/file-associations-demo).
Behaviours per OSs:
- MacOS Sonoma: associations work for distributables.
- Windows 11: associations work after the installation of the MSI.
- Kubuntu: associations do not work, but everything else works fine. However, IDEA also does not have associations there, so I assume this is fine.

I didn't write any unit tests because I don’t know which of them you are expecting me to write. So, I'm looking forward to your feedback and suggestions.

This should be tested by QA

### Highlight - Desktop
- Introduction of the new DSL function in `nativeDistributions` block:
  ```kotlin
  fun fileAssociation(mimeType: String, extension: String, description: String): Unit
  ```


igordmn:
@terrakok, could you also review the new API in this PR? I am also reviewing it and the implementation.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
```diff
@@ -368,6 +378,31 @@ abstract class AbstractJPackageTask @Inject constructor(
...
+                .groupBy { it.extension }
+                .mapValues { (extension, associations) ->
+                    associations.mapIndexed { index, association ->
+                        propertyFilesDirJava.resolve("FA${extension}${if (index > 0) index.toString() else ""}.properties")
+                            .apply {
+                                writeText(
+                                    """
+                                        mime-type=${association.mimeType}
+                                        extension=${association.extension}
+                                        description=${association.description}
```
igordmn, line 394:
```suggestion
                                        description=${association.description}
                                        icon=${association.iconPath}
```

Let's also add support for settings icons (or create a follow-up issue for adding it).

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
```diff
@@ -32,4 +32,9 @@ abstract class JvmApplicationDistributions : AbstractDistributions() {
     fun windows(fn: Action<WindowsPlatformSettings>) {
         fn.execute(windows)
     }
+    
+    internal val fileAssociations: MutableSet<FileAssociation> = mutableSetOf()
+    fun fileAssociation(mimeType: String, extension: String, description: String) {
```
igordmn, line 37: Let's move it to https://github.com/JetBrains/compose-multiplatform/blob/67da8d12f9b18f6aefe68a449cb2ed8c3eb8646f/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/PlatformSettings.kt#L19

Other common properties are there, and we can specify them separately.

igordmn, line 37: Let's add a doc for this function to https://github.com/JetBrains/compose-multiplatform/tree/master/tutorials/Native_distributions_and_local_execution#platform-specific-options

(1-2 points below `installationPath`

igordmn (commented):
Thanks!

A few changes needed, an API review from @terrakok, and it is good to merge.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
zhelenskiy: But is it common? It works generally only on desktop, which is JVM. Or should it also work on MacOS Native?

igordmn: I meant common for desktop platforms. `PlatformSettings` is only for desktop.

Native distribution doesn't support MacOs Native, but if there will be one, it will read `PlatformSettings`

zhelenskiy: I have never tried creating MacOS Kotlin/Native. Does it use user Info.plist as the iOS target does, or a generated one as Desktop/JVM one does for MacOS? If the first option takes place, we cannot add support for MacOS native for the feature.

gradle-plugins/compose/src/test/test-projects/application/macOptions/Expected-Info.plist
```diff
@@ -1,37 +1,37 @@
...
-  <string>Unknown</string>
-  <key>CFBundleVersion</key>
-  <string>1.0.0</string>
-  <key>NSHumanReadableCopyright</key>
-  <string>Copyright (C) CURRENT_YEAR</string>
-  <key>NSSupportsAutomaticGraphicsSwitching</key>
-  <string>true</string>
-  <key>NSHighResolutionCapable</key>
-  <string>true</string>
+  <dict>
```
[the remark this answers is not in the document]
zhelenskiy: It was "user-defined extra", but I updated it as well to look better.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
zhelenskiy: As icons are platform-specific, how would you like to see the api: writing a single `fileAssociation` function with `windowsIcon` etc. or writing `fileAssociation` function with `iconFile` for every platform?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
zhelenskiy: But then I'll have to specify the associations for each platform separately.
image

igordmn: Yes. This was a limitation before for other properties as well (packageVersion, for example). We shouldn't fix this limitation in this feature, otherwise we'll have inconsistent API.

If you need to specify it for all platforms, you can extract an extension function `AbstractPlatformSettings.appFileAssociations()`

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
igordmn: As suggested before, we need to move it to `AbstractPlatformSettings`. This way, you should specify `fileAssociation` for each platform separately, or extract own function, in which you can apply own logic of choosing the icon.

See similar property `AbstractPlatformSettings.iconFile`

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
zhelenskiy: But icon format is different for each window type. Maybe, I can add an extension which will add the default option for each platform?

igordmn: This is an issue, agree. We have the same issue with the application icon, but it is less painful, as it is a single icon.

A simple extension won't work, as it is called on the configuration time, not on the task execution time (the host OS can be different from the OS for which we build in the future).

We have to provide some `Provider<>` and set an appropriate extension when  `package*` task is called. Not sure how it should look like - we need to try.

Alternatively we can:
1. Delegate this to the user:
```kotlin
fun AbstractPlatformSettings.appFileAssociations(iconExtension) {
   fileAssociation(..., iconFile = project.file("icons/txt.$iconExtension"))
}

windows.appFileAssociations(iconExtension = "ico")
macOs.appFileAssociations(iconExtension = "icns")
linux.appFileAssociations(iconExtension = "png")
```

2. Add extension automatically if it isn't set:
```kotlin
fun AbstractPlatformSettings.appFileAssociations(iconExtension) {
   fileAssociation(..., iconFile = project.file("icons/txt"))
}
```

I would choose 2 as the simplest solution to implement and use. With this approach, I also don't mind to have it on `nativeDistributions` level (I looked at the code more closely, and it seems both levels are fine, with their own cons, which should be solved not in this feature).

@terrakok, @zhelenskiy, what do you think?

gradle-plugins/compose/src/test/test-projects/application/macOptions/Expected-Info.plist
terrakok, line 1: Please, add a test for the new DSL

terrakok (changes requested):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/InfoPlistBuilder.kt
```diff
@@ -5,14 +5,63 @@
...
+                ) { (key, value) ->
+                    "${indentForLevel(nestingLevel + 1)}<key>${key.name}</key>\n${value.toString(nestingLevel + 1)}"
+                }
+            override fun toString() = super.toString()
+
+            constructor(vararg elements: Pair<InfoPlistKey, InfoPlistValue>) : this(elements.toMap())
+        }
+
+        data class InfoPlistStringValue(val value: String) : InfoPlistValue() {
+            override fun toString(nestingLevel: Int): String = if (value.isEmpty()) "${indentForLevel(nestingLevel)}<string/>" else "${indentForLevel(nestingLevel)}<string>$value</string>"
```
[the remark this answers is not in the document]
zhelenskiy: I just want to be consistent with other tags.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/FileAssociation.kt
```diff
@@ -0,0 +1,9 @@
+package org.jetbrains.compose.desktop.application.dsl
+
+import java.io.Serializable
+
+data class FileAssociation(
```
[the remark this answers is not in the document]
igordmn: It may be fine for Gradle plugin, as it isn't used as a transitive library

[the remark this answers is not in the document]
zhelenskiy: I'll make it internal to make sure nobody can access it outside the plugin.

zhelenskiy:
Except for DSL test and docs everything else is fixed. Can you close [the comment](https://github.com/JetBrains/compose-multiplatform/pull/4957#discussion_r1648681063), @terrakok? It is not closable.

What do you think about the API. It supports:
* Ability to make platform-specific file associations.
* Ability to make platform-independent associations without copying.
* Ability to provide or not an icon.
* For platform-specific association icons you need to specify only one property `iconFile` while for universal you need to specify `windowsIconFile`, etc.

The demo project is also updated according to the new API.

PS. This version is only tested on MacOS for now. I'll check the rest a bit later.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
```diff
@@ -32,4 +33,13 @@ abstract class JvmApplicationDistributions : AbstractDistributions() {
...
     }
+    
+    fun fileAssociation(
+        mimeType: String, extension: String, description: String,
+        linuxIcon: File? = null, windowsIcon: File? = null, macOSIcon: File? = null,
+    ) {
+        linux.fileAssociation(mimeType, extension, description, linuxIcon)
+        windows.fileAssociation(mimeType, extension, description, windowsIcon)
+        macOS.fileAssociation(mimeType, extension, description, macOSIcon)
+    }
```
igordmn, line 44:
```suggestion
    fun fileAssociation(
        mimeType: String, extension: String, description: String,
        linuxIconFile: File? = null, windowsIconFile: File? = null, macOSIconFile: File? = null,
    ) {
        linux.fileAssociation(mimeType, extension, description, linuxIconFile)
        windows.fileAssociation(mimeType, extension, description, windowsIconFile)
        macOS.fileAssociation(mimeType, extension, description, macOSIconFile)
    }
```

igordmn (approved):
Looks good to me after resolving 2 my open comments.

The end review and merging I delegate to @terrakok.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
zhelenskiy: I extracted it to a new section as it is quite large because of an example.

zhelenskiy:
I had to change resource name generation to make it stable.

zhelenskiy:
I tested the current version on Windows, MacOS, Ubuntu, Kubuntu. On Windows, MacOS everything is ok, while on both Linux distributives there are problems (in Kubuntu there is no icon, in Ubuntu, association doesn't work), but the same thing is with idea as well.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt
```diff
@@ -244,6 +248,40 @@ abstract class AbstractJPackageTask @Inject constructor(
     @get:Optional
     val javaRuntimePropertiesFile: RegularFileProperty = objects.fileProperty()
 
+    @get:Input
+    @get:Optional
+    internal val fileAssociations: SetProperty<FileAssociation> = objects.setProperty(FileAssociation::class.java)
```
terrakok, line 253: The property is not optional as I see. It is configured in anyway

terrakok (approved):

zhelenskiy:
Everything seems fixed now.

terrakok (commented):
val fileAssociationFiles = fileAssociations.**orNull**.orEmpty()

do we need it iyet?

tutorials/Native_distributions_and_local_execution/README.md
```diff
@@ -511,6 +511,60 @@ compose.desktop {
 }
 ```
 
+## File associations
```
igordmn, line 514: Thanks!

Could you move the docs to another PR to `support/1.7.0` branch? It will be reviewed by tech writers and be merged at the release date)

gradle-plugins/compose/src/test/test-projects/application/macOptions/build.gradle
```diff
@@ -28,12 +28,36 @@ compose.desktop {
         mainClass = "MainKt"
         nativeDistributions {
             packageName = "TestPackage"
+            fileAssociation(
+                    "text/kotlin",
+                    "kot",
+                    "Kotlin Source File0",
+                    project.file("Kotlin_icon_big.png"),
+                    project.file("Kotlin_icon_big.ico"),
```
igordmn, line 36: I checked it on Windows, and it doesn't seem to work:
image


It has the app icon instead.

JDK: `openjdk 19.0.2 2023-01-17`

igordmn: Never mind. It works.

I put icons in the root folder instead of the module folder

zhelenskiy: Can you give the example standalone?

igordmn: Do you need if it works? :) I was mistaken.

zhelenskiy: I just had out-of-date message history, no need, of cource.

igordmn:
Looks good to me, thanks!

@terrakok, please merge, if everything is okay to you too.

tutorials/Native_distributions_and_local_execution/README.md
zhelenskiy: Done

PR #4958, opened 2024-06-12
Make sure the web app distribution doesn't contain a duplicate skiko.wasm

eymar:
https://youtrack.jetbrains.com/issue/CMP-1114

**Testing:**
- changed an existing test to test where added a new check for the content of the app distribution


Schahen (commented):
LGTM

PR #4959, opened 2024-06-12
Delete outdated build services

terrakok:
We had two build services:
 1) to check unsupported compose compiler plugins were applied
 2) to check a native cache kind configuration

Both of them are outdated and we may get rid of them.

Fixes https://github.com/JetBrains/compose-multiplatform/issues/4815

### Fixes - Gradle Plugin
- Delete outdated build services


eymar (approved):
We discussed it today. Here is why I think these changes are okay:
- checking for 3rd party compiler plugin is not needed anymore, since we already have one compiler plugin for all platforms in Kotlin 2. The check might be still relevant for projects with  kotlin < 2.0, but given that Build Service causes issues in some cases, the check can be dropped now, without waiting for full kotlin 2.0 adoption
- check for native caches flags. We used to force all users to disable caches to make Compose work. Then when k/native implemented certain fixes, the caches started to work, and we wanted the users to drop their settings where they disabled the caches. Now, most projects probably dropped it. And the check is not relevant anymore.

igordmn (approved):

PR #4969, opened 2024-06-14
[resources] Update resource density-based lookup to be equal with the android logic

terrakok:
In general, Android prefers scaling down a larger original image to scaling up a smaller original image: https://developer.android.com/guide/topics/resources/providing-resources#BestMatch

Fixes https://github.com/JetBrains/compose-multiplatform/issues/4368

### Highlights - Resources
- If there is no resource with suitable density, use resource with the most suitable density, otherwise use default (similar to the Android logic)


igordmn (approved):

igordmn:
> Update resource filtering and add density-based lookup to be equal with the android logic

Better to be more specific. Something like:

```
If there is no resource with suitable density, use resource with the closest lower density, otherwise use default (similar to the Android logic)
```

igordmn (approved):
Looks good, could you also [change](https://github.com/JetBrains/compose-multiplatform/pull/4969#issuecomment-2173653537) the PR description?

Issue #2921, opened 2023-03-25
Automatic Preview Updates


mgroth0:
Is there any way to make the Desktop Preview update automatically as you type your code? Even with macros? Is there at least a keyboard shortcut? Clicking the little icon over and over is not very efficient, and I can't tell if there is any alternative.

sergeshustoff:
Could you please at least add a button somewhere in the preview window to refresh it manually? Sometimes it requires some scrolling to find the preview button and refresh after some changes. Preview itself is quite usefull, but requires extra steps to use)

okushnikov:
Please check the <a href="https://youtrack.jetbrains.com/issue/CMP-2921">following ticket</a> on YouTrack for follow-ups to this issue. GitHub issues will be closed in the coming weeks.

PR #4977, opened 2024-06-16
Refresh desktop preview (#2921)

sergeshustoff:
Add refresh action to desktop preview window
Add global action for refreshing desktop preview (keymap can be assigned) 
Partially addresses #2921 by allowing user to refresh already shown preview more easily after changing the previewed UI code


natanfudge:
This very simple change will improve the preview experience dramatically.

PR #5054, opened 2024-07-01
Add file associations support documentation

### LINKED ISSUE #773: Desktop: Add support for file associations

zhelenskiy:
Docs for file associations support (https://github.com/JetBrains/compose-multiplatform/pull/4957) which fixes #773.

### DIFF
--- a/tutorials/Native_distributions_and_local_execution/README.md
+++ b/tutorials/Native_distributions_and_local_execution/README.md
@@ -546,6 +546,7 @@ compose.desktop {
                 )
             }
 
+            // OS-agnostic file association for .kot files, with icons for all three operating systems.
             fileAssociation(
                 mimeType = "text/kotlin",
                 extension = "kot",
--- a/tutorials/Native_distributions_and_local_execution/README.md
+++ b/tutorials/Native_distributions_and_local_execution/README.md
@@ -511,6 +511,60 @@ compose.desktop {
 }
 ```
 
+## File associations
+
+File associations can be added with `fileAssociation` function.
+These associations can optionally include custom icons.
+Similar to [app icon](#app-icon), these custom icons also have OS-specific formats: `.icns` for macOS, `.ico` for Windows and `.png` for Linux.
+
+It is possible to specify both OS-specific and OS-agnostic file associations.
+For OS-agnostic you can specify icon files for all three OSes in-place,
+while for OS-specific there is a shorthand with a single icon file parameter.
+
+```kotlin
+compose.desktop {
+    application {
+        mainClass = "MainKt"
+
+        nativeDistributions {
+            targetFormats(TargetFormat.Dmg, TargetFormat.Msi, TargetFormat.Deb)
+
+            packageName = "file-associations-demo"
+            packageVersion = "1.0.0"
+
+            macOS {
+                fileAssociation(
+                    mimeType = "text/kotlin",
+                    extension = "kott",
+                    description = "Kotlin Source File",
+                    iconFile = project.file("Kotlin_icon_big.icns"),
+                )
+                fileAssociation(
+                    mimeType = "text/kotlin",
+                    extension = "ko",
+                    description = "Kotlin Source File",
+                )
+            }
+
+            fileAssociation(
+                mimeType = "text/kotlin",
+                extension = "kot",
+                description = "Kotlin Source File",
+                macOSIconFile = project.file("Kotlin_icon_big.icns"),
+                windowsIconFile = project.file("Kotlin_icon_big.ico"),
+                linuxIconFile = project.file("Kotlin_icon_big.png"),
+            )
+
+            fileAssociation(
+                mimeType = "text/kotlin",
+                extension = "kottt",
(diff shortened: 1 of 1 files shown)


igordmn (commented):
LGTM, will send to copyediting after approve from Zamula

tutorials/Native_distributions_and_local_execution/README.md
```diff
@@ -511,6 +511,60 @@ compose.desktop {
...
+                    iconFile = project.file("Kotlin_icon_big.icns"),
+                )
+                fileAssociation(
+                    mimeType = "text/kotlin",
+                    extension = "ko",
+                    description = "Kotlin Source File",
+                )
+            }
+
+            fileAssociation(
```
zamulla, line 549:
```suggestion
            // OS-agnostic file association for .kot files, with icons for all three operating systems.
            fileAssociation(
```

tutorials/Native_distributions_and_local_execution/README.md
```diff
@@ -511,6 +511,60 @@ compose.desktop {
...
+                mimeType = "text/kotlin",
+                extension = "kot",
+                description = "Kotlin Source File",
+                macOSIconFile = project.file("Kotlin_icon_big.icns"),
+                windowsIconFile = project.file("Kotlin_icon_big.ico"),
+                linuxIconFile = project.file("Kotlin_icon_big.png"),
+            )
+
+            fileAssociation(
```
zamulla, line 558:
```suggestion
            // OS-agnostic file association for .kottt files, without icons.
            fileAssociation(
```

tutorials/Native_distributions_and_local_execution/README.md
```diff
@@ -511,6 +511,60 @@ compose.desktop {
...
 
+## File associations
+
+File associations can be added with `fileAssociation` function.
+These associations can optionally include custom icons.
+Similar to [app icon](#app-icon), these custom icons also have OS-specific formats: `.icns` for macOS, `.ico` for Windows and `.png` for Linux.
+
+It is possible to specify both OS-specific and OS-agnostic file associations.
+For OS-agnostic you can specify icon files for all three OSes in-place,
+while for OS-specific there is a shorthand with a single icon file parameter.
```
zamulla, line 522:
```suggestion
while for OS-specific there is a shorthand with a single `iconFile` parameter.
```

zamulla (changes requested):
Most importantly I think we would benefits from comments in the code, other suggestions are phrasing polish.

zamulla (approved):
Thanks, looks good 👍

igordmn (approved):

PR #5080, opened 2024-07-08
Update compose compiler version for kotlin 1.9.24

eymar:
This fixes https://youtrack.jetbrains.com/issue/CMP-1612 and https://youtrack.jetbrains.com/issue/CMP-1602/Html-js-error-content-is-not-a-function

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
@@ -24,7 +24,7 @@ internal object ComposeCompilerCompatibility {
         "1.9.21" to "1.5.4",
         "1.9.22" to "1.5.8.1",
         "1.9.23" to "1.5.13.5",
-        "1.9.24" to "1.5.14",
+        "1.9.24" to "1.5.14.1-beta02",
         "2.0.0-Beta1" to "1.5.4-dev1-kt2.0.0-Beta1",
         "2.0.0-Beta4" to "1.5.9-kt-2.0.0-Beta4",
         "2.0.0-Beta5" to "1.5.11-kt-2.0.0-Beta5",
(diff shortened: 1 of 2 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposeCompilerCompatibility.kt
```diff
@@ -24,7 +24,7 @@ internal object ComposeCompilerCompatibility {
         "1.9.21" to "1.5.4",
         "1.9.22" to "1.5.8.1",
         "1.9.23" to "1.5.13.5",
-        "1.9.24" to "1.5.14",
+        "1.9.24" to "1.5.14.1-beta02",
```
igordmn, line 27: Please, create a task with the target "1.7.0" to make it stable

igordmn:
We still have a crash on Kotlin 1.9.23 and Kotlin 2.0.0. How difficult it will be to maintain backward compatibility with these Kotlin's?

Alternatively:
- For 1.9.23 we need to release a new Compiler (not sure if it worth it).
- For Kotlin 2.0.0 we can decide that it should be updated to a hotfix version of Kotlin 2.0.10

eymar:
> We still have a crash on Kotlin 1.9.23 and Kotlin 2.0.0. How difficult it will be to maintain backward compatibility with these Kotlin's?
> 
> Alternatively:
> For 1.9.23 we need to release a new Compiler (not sure if it worth it).
> For Kotlin 2.0.0 we can decide that it should be updated to a hotfix version of Kotlin 2.0.10


- For 1.9.23 it's up to us to publish a compiler plugin with a fix. (Probably we don't need to do that, as everyone can use 1.9.24) 
- For 2.0.0, since it's already published, the only way to make 1.7.0- compatible with 2.0.0 is to remove a newly added function in compose runtime (web). Compose compiler (2.0.0) looks for that function, and if it's present, it starts using it (the problem is that 2.0.0 Compose compiler, the web pieces "don't know" about the new function). 


IMO, removing the function for only 1 platform is not a future-proof solution, because this way the platforms behaviour might diverge and it might cause other issues. 
Suggesting users to switch to 2.0.10 is the best solution IMO, especially considering the Alpha status of CfW.

igordmn:
> Suggesting users to switch to 2.0.10 is the best solution IMO, especially considering the Alpha status of CfW.

Agree, I didn't take Alpha status into account.

igordmn (approved):

PR #5101, opened 2024-07-17
[resources] Downscale drawable resources if it came from upper dpi.

terrakok:
Now drawables from upper DPIs will be downscalled to the expected size. (the same behavior as on Android)
**BEFORE**

**AFTER**

Fixes https://youtrack.jetbrains.com/issue/CMP-5657

### Fixes - Resources
- Now drawables from upper DPIs will be downscalled to the expected size. (the same behavior as on Android)

### DIFF
--- a/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
+++ b/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
@@ -56,10 +56,17 @@ private val emptyImageBitmap: ImageBitmap by lazy { ImageBitmap(1, 1) }
 @Composable
 fun imageResource(resource: DrawableResource): ImageBitmap {
     val resourceReader = LocalResourceReader.currentOrPreview
-    val imageBitmap by rememberResourceState(resource, resourceReader, { emptyImageBitmap }) { env ->
-        val path = resource.getResourceItemByEnvironment(env).path
-        val cached = loadImage(path, resourceReader) {
-            ImageCache.Bitmap(it.toImageBitmap())
+    val resourceEnvironment = rememberResourceEnvironment()
+    val imageBitmap by rememberResourceState(
+        resource, resourceReader, resourceEnvironment, { emptyImageBitmap }
+    ) { env ->
+        val item = resource.getResourceItemByEnvironment(env)
+        val resourceDensityQualifier = item.qualifiers.firstOrNull { it is DensityQualifier } as? DensityQualifier
+        val resourceDensity = resourceDensityQualifier?.dpi ?: DensityQualifier.MDPI.dpi
+        val screenDensity = resourceEnvironment.density.dpi
+        val path = item.path
+        val cached = loadImage(path, "$path-${screenDensity}dpi", resourceReader) {
+            ImageCache.Bitmap(it.toImageBitmap(resourceDensity, screenDensity))
         } as ImageCache.Bitmap
         cached.bitmap
     }
@@ -82,7 +89,7 @@ fun vectorResource(resource: DrawableResource): ImageVector {
     val density = LocalDensity.current
     val imageVector by rememberResourceState(resource, resourceReader, density, { emptyImageVector }) { env ->
         val path = resource.getResourceItemByEnvironment(env).path
-        val cached = loadImage(path, resourceReader) {
+        val cached = loadImage(path, path, resourceReader) {
             ImageCache.Vector(it.toXmlElement().toImageVector(density))
         } as ImageCache.Vector
         cached.vector
@@ -102,7 +109,7 @@ private fun svgPainter(resource: DrawableResource): Painter {
     val density = LocalDensity.current
     val svgPainter by rememberResourceState(resource, resourceReader, density, { emptySvgPainter }) { env ->
         val path = resource.getResourceItemByEnvironment(env).path
-        val cached = loadImage(path, resourceReader) {
+        val cached = loadImage(path, path, resourceReader) {
             ImageCache.Svg(it.toSvgElement().toSvgPainter(density))
         } as ImageCache.Svg
         cached.painter
@@ -126,7 +133,7 @@ suspend fun getDrawableResourceBytes(
     return DefaultResourceReader.read(resourceItem.path)
 }
 
-internal expect fun ByteArray.toImageBitmap(): ImageBitmap
+internal expect fun ByteArray.toImageBitmap(resourceDensity: Int, targetDensity: Int): ImageBitmap
 internal expect fun ByteArray.toXmlElement(): Element
 internal expect fun ByteArray.toSvgElement(): SvgElement
 
@@ -145,6 +152,7 @@ internal fun dropImageCache() {
 
 private suspend fun loadImage(
     path: String,
+    cacheKey: String,
     resourceReader: ResourceReader,
     decode: (ByteArray) -> ImageCache
-): ImageCache = imageCache.getOrLoad(path) { decode(resourceReader.read(path)) }
+): ImageCache = imageCache.getOrLoad(cacheKey) { decode(resourceReader.read(path)) }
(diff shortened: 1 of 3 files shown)


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
```diff
@@ -56,10 +56,17 @@ private val emptyImageBitmap: ImageBitmap by lazy { ImageBitmap(1, 1) }
...
-        val path = resource.getResourceItemByEnvironment(env).path
-        val cached = loadImage(path, resourceReader) {
-            ImageCache.Bitmap(it.toImageBitmap())
+    val resourceEnvironment = rememberResourceEnvironment()
+    val imageBitmap by rememberResourceState(
+        resource, resourceReader, resourceEnvironment, { emptyImageBitmap }
+    ) { env ->
+        val item = resource.getResourceItemByEnvironment(env)
+        val resourceDensityQualifier = item.qualifiers.firstOrNull { it is DensityQualifier } as? DensityQualifier
+        val resourceDensity = resourceDensityQualifier?.dpi ?: DensityQualifier.MDPI.dpi
```
igordmn, line 65: Does Android do anything if the image just in `drawable` and device density is LDPI?

If it doesn't downscale, it should be `?: DensityQualifier.LDPI.dpi` (or better just `null`, with additional checks)

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/ImageResources.skiko.kt
```diff
@@ -4,12 +4,34 @@ import androidx.compose.ui.graphics.ImageBitmap
...
+    if (resourceDensity > targetDensity) {
+        val scale = targetDensity.toFloat() / resourceDensity.toFloat()
+        val targetH = image.height * scale
+        val targetW = image.width * scale
+        val srcRect = Rect.Companion.makeWH(image.width.toFloat(), image.height.toFloat())
+        val dstRect = Rect.Companion.makeWH(targetW, targetH)
+
+        targetImage = Surface.makeRasterN32Premul(targetW.toInt(), targetH.toInt()).run {
+            val paint = Paint().apply { isAntiAlias = true }
+            canvas.drawImageRect(image, srcRect, dstRect, paint)
```
igordmn, line 26: For resizing images we also need to set SamplingMode
image

What will be the best for downscaling? In my experience, `nearest` produces acceptable output only with integers (2x, 3x, etc).

image

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
terrakok: On android the default density is MDPI. So, it works as expected:
image

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/ImageResources.skiko.kt
terrakok: I checked it visually and the default mode seems very close to the android downscaling result

terrakok: image

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
igordmn: > On android the default density is MDPI

Is the `drawable` folder is for the default density MDPI?

I mean, from the user perspective, I would expect it to be density-independent, that requires to not do any downscale/upscale with images in it.

If `drawable` is really for MDPI in Android, let's keep the current behaviour.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
```diff
@@ -56,10 +56,17 @@ private val emptyImageBitmap: ImageBitmap by lazy { ImageBitmap(1, 1) }
...
+        val screenDensity = resourceEnvironment.density.dpi
+        val path = item.path
+        val cached = loadImage(path, "$path-${screenDensity}dpi", resourceReader) {
```
igordmn, line 68:
```suggestion
        val cached = loadImage(path, "$path-${resourceDensity}-${screenDensity}dpi", resourceReader) {
```
Different original `resourceDensity`'s produce different quality results

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/ImageResources.skiko.kt
```diff
@@ -4,12 +4,34 @@ import androidx.compose.ui.graphics.ImageBitmap
 import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.graphics.toComposeImageBitmap
 import androidx.compose.ui.unit.Density
-import org.jetbrains.skia.Data
-import org.jetbrains.skia.Image
+import org.jetbrains.skia.*
```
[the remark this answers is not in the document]
terrakok: done

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/ImageResources.skiko.kt
terrakok: done

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
terrakok: Yes. I see that android downscales png from the `drawable` folder on a LDPI device:
image
original image
image

terrakok: we don't support nodpi yet.

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/ImageResources.skiko.kt
igordmn: I checked it on desktop (`resourceDensity = XXXHDPI, screenDensity = HDPI`). It doesn't looks good:

[downscaleResults.zip](https://github.com/user-attachments/files/16266314/downscaleResults.zip)

I would prefer we change it to `LINEAR`, others are too slow.

```
canvas.drawImageRect(image, srcRect, dstRect, SamplingMode.LINEAR, paint, true)
```

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/ImageResources.skiko.kt
```diff
@@ -22,7 +28,10 @@ internal actual fun ByteArray.toImageBitmap(resourceDensity: Int, targetDensity:
         val dstRect = Rect.Companion.makeWH(targetW, targetH)
 
         targetImage = Surface.makeRasterN32Premul(targetW.toInt(), targetH.toInt()).run {
-            val paint = Paint().apply { isAntiAlias = true }
+            val paint = Paint().apply {
+                isAntiAlias = true
+                imageFilter = ImageFilter.makeImage(image, srcRect, dstRect, SamplingMode.MITCHELL)
```
igordmn, line 33: In `ImageFilter.makeImage` I am not sure what `srcRect, dstRect` mean, maybe we just pass it to `drawImageRect`? It seems on a lower level

igordmn: Probably, I understood, it will draw the image on any draw operation. So, `drawImageRect` will draw 2 images. Let's just use `drawImageRect`

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
terrakok: no. because the resourceDensity is already presented in the path

igordmn (approved):

PR #5105, opened 2024-07-22
Fix skiko.js unpacking for k/js target

eymar:
K/JS and K/Wasm have differences in the packaging logic, and therefore we need to account for it when unpacking Skiko files.

Fixes [CMP-5649](https://youtrack.jetbrains.com/issue/CMP-5649)

## Testing
- Added a test, which checks the state of k/js distribution

This should be tested by QA

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
@@ -85,6 +85,7 @@ internal fun configureWebApplication(
 
     targets.forEach { target ->
         target.compilations.all { compilation ->
+            // `wasmTargetType` is available starting with kotlin 1.9.2x
             if (target.wasmTargetType != null) {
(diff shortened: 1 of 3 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -77,8 +79,16 @@ internal fun configureWebApplication(
     }
 
     project.tasks.withType(IncrementalSyncTask::class.java) {
+        if (it.name.contains("wasmJs", ignoreCase = true)) {
+            it.dependsOn(unpackRuntime)
+            it.from.from(unpackedRuntimeDir)
+        }
```
Schahen, line 85: Can we leave small explanation on why we are actually doing this? 
The other question - any chance those "wasmJs" tasks have some narrower type we can rely on rather then IncrementalSyncTask

eymar: refactored and added the comments. Thanks to @ilgonmic

Schahen (approved):

PR #5115, opened 2024-08-05
Benchmarks refactoring

pjBooms:
Fixes [the issue](https://youtrack.jetbrains.com/issue/CMP-5894/Refactor-Compose-benchmarks-to-count-GC-between-frames): count GC between frames.

## Testing

Run the benchmarks locally.

### DIFF
--- a/benchmarks/kn-performance/src/commonMain/kotlin/MeasureComposable.kt
+++ b/benchmarks/kn-performance/src/commonMain/kotlin/MeasureComposable.kt
@@ -1,17 +1,17 @@
 import androidx.compose.runtime.Composable
-import androidx.compose.ui.ComposeScene
-import androidx.compose.ui.unit.Constraints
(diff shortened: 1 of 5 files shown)


elijah-semyonov (approved):

benchmarks/kn-performance/src/commonMain/kotlin/MeasureComposable.kt
```diff
@@ -1,69 +1,96 @@
...
+            scene.render(canvas, it * nanosPerFrame)
+        }
 
         val frames = MutableList(frameCount) {
             BenchmarkFrame(Duration.INFINITE, Duration.INFINITE)
         }
 
-        var nanoTime = 0L
+        var nextVSync = Duration.ZERO
+        var missedFrames = 0;
```
haitaka, line 64: Should this value be reported? Or does it somehow indirectly affects the reported values?

PR #5121, opened 2024-08-20
[gradle] Add dependency shortcuts for Compose Material3 adaptive components

terrakok:
### Features - Gradle Plugin
- New dependency shortcuts: `compose.material3Adaptive`, `compose.material3AdaptiveLayout`, `compose.material3AdaptiveNavigation`

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
@@ -19,19 +19,23 @@ import org.jetbrains.compose.desktop.DesktopExtension
 import org.jetbrains.compose.desktop.application.internal.configureDesktop
 import org.jetbrains.compose.desktop.preview.internal.initializePreview
 import org.jetbrains.compose.experimental.dsl.ExperimentalExtension
-import org.jetbrains.compose.experimental.internal.*
-import org.jetbrains.compose.internal.*
+import org.jetbrains.compose.experimental.internal.configureExperimentalTargetsFlagsCheck
+import org.jetbrains.compose.internal.KOTLIN_MPP_PLUGIN_ID
+import org.jetbrains.compose.internal.mppExt
+import org.jetbrains.compose.internal.mppExtOrNull
 import org.jetbrains.compose.internal.utils.currentTarget
 import org.jetbrains.compose.resources.ResourcesExtension
 import org.jetbrains.compose.resources.configureComposeResources
 import org.jetbrains.compose.web.WebExtension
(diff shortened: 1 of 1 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -19,19 +19,23 @@ import org.jetbrains.compose.desktop.DesktopExtension
...
-import org.jetbrains.kotlin.com.github.gundy.semver4j.SemVer
 import org.jetbrains.kotlin.gradle.dsl.KotlinCompile
 import org.jetbrains.kotlin.gradle.dsl.KotlinJsCompile
-import org.jetbrains.kotlin.gradle.plugin.*
+import org.jetbrains.kotlin.gradle.plugin.KotlinDependencyHandler
+import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
+import org.jetbrains.kotlin.gradle.plugin.getKotlinPluginVersion
 
 internal val composeVersion get() = ComposeBuildConfig.composeVersion
+private const val material3AdaptiveVersion = "1.0.0-alpha01"
```
igordmn, line 38: This will require updating it every time we build a new version. 

I suggest to not merge it, considering that:
1. it complicates the release process, as there are too many implicit things we need to keep in mind
2. `adaptive` will probably will be aligned with material3 when it is released
3. if we decide to do it anyway, we need to automate it on CI
4. we are thinking about moving from DSL versioning to `libs.version.toml`
5. we have other non-aligned libs like navigation/lifecycle, we mention versions for them in the changelog

igordmn (changes requested):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
igordmn: > if we decide to do it anyway, we need to automate it on CI

We can read it from `compose-multiplatform-core` repo after [CMP-5665](https://youtrack.jetbrains.com/issue/CMP-5665) Change versioning (get rid of 0.0.0 and 1.7.0-devXXX)

terrakok:
OK. I close the issue until we decide the way how to provide the versions.
Current way is:
```
dependencies {
  implementation("org.jetbrains.compose.material3.adaptive:adaptive:1.0.0-alpha01")
  implementation("org.jetbrains.compose.material3.adaptive:adaptive-layout:1.0.0-alpha01")
  implementation("org.jetbrains.compose.material3.adaptive:adaptive-navigation:1.0.0-alpha01")
}
```

kropp:
We are deprecating all dependency aliases in #5462

PR #5122, opened 2024-08-23
Support test resources in Compose UI tests.

terrakok:
1) The PR adds a support test resources in Compose multiplatform projects.
2) The PR adds a support multi-module resources in JVM-only projects.

Fixes https://youtrack.jetbrains.com/issue/CMP-1470
Fixes https://youtrack.jetbrains.com/issue/CMP-5963

### Features - Resources
- Added support of test resources in Compose Multiplatform projects
- Added support of multi-module resources in JVM-only projects


components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/AndroidContextProvider.kt
```diff
@@ -37,11 +38,34 @@ fun PreviewContextConfigurationEffect() {
...
+ * Example usage:
+ * ```
+ * fun configureTestEnvironment() {
+ *     setAndroidInstrumentedContext{
+ *         InstrumentationRegistry.getInstrumentation().context
+ *     }
+ * }
+ * ```
+ */
+@ExperimentalResourceApi
```
igordmn, line 56: Let's create an issue for removing Experimental with target = 1.8

igordmn (approved):

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/AndroidContextProvider.kt
```diff
@@ -37,11 +38,34 @@ fun PreviewContextConfigurationEffect() {
...
+fun setAndroidInstrumentedContext(context: Context) {
```
igordmn, line 57:
```suggestion
fun setResourcesAndroidInstrumentedContext(context: Context) {
```

Better to not use a generic name for a resource function

igordmn:
Adding a second reviewer, as we have a new API `setAndroidInstrumentedContext`

igordmn:
@eymar, could you review the new API?

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/AndroidContextProvider.kt
```diff
@@ -37,11 +38,34 @@ fun PreviewContextConfigurationEffect() {
...
+/**
+ * Sets the Android instrumented context.
+ *
+ * @param context The Android context to be set as the instrumented context.
+ *                This context will be used by a ResourceReader to read test assets.
+ *
+ * Example usage:
+ * ```
+ * fun configureTestEnvironment() {
+ *     setAndroidInstrumentedContext{
```
eymar, line 50: The example calls the function like it takes a lambda. But the actul parameter type is just Context. Probably a typo? And can be fixed by replacing {...} with (...)

gradle-plugins/compose/src/test/test-projects/misc/testResources/src/commonTest/kotlin/CommonUiTest.kt
```diff
@@ -0,0 +1,39 @@
...
+                assertEquals("Compose Resources App", mainStr)
+                assertEquals("Common test", testStr)
+                assertNotEquals("Compose Resources App", testStr)
+            }
+        }
+    }
+
+    @Test
+    fun checkTestFileResource() {
+        runTest {
```
eymar, line 31: If this test would be on web eventually, `runTest` result should be returned from the function:

```kotlin
@Test
fun fooTest() = runTest { ... }
```

In the future (after 1.8), this requirement might affect `runComposeUiTest` too.

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/AndroidContextProvider.kt
eymar: This API is android only (and it makes sense). 
I'm wondering how would developers write a common test in this case?

Do they need to create some expect/actual on their side?

terrakok: yes, the only way to read test resources on the android is provide the special context.
So, users are supposed to do it by their own

(or we can implement it in `compose-ui-tests` module, I guess)

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -683,4 +680,22 @@ class ResourcesTest : GradlePluginTestBase() {
...
+                check.logContains("Configure main resources for 'desktop' target")
+                check.logContains("Configure test resources for 'desktop' target")
+                check.logContains("Configure main resources for 'iosX64' target")
+                check.logContains("Configure test resources for 'iosX64' target")
+                check.logContains("Configure main resources for 'iosArm64' target")
+                check.logContains("Configure test resources for 'iosArm64' target")
+                check.logContains("Configure main resources for 'iosSimulatorArm64' target")
+                check.logContains("Configure test resources for 'iosSimulatorArm64' target")
+
+                check.taskSuccessful(":desktopTest")
```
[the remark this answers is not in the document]
terrakok: I run common "check" task, so ios tests should be run on macos machines: https://github.com/JetBrains/compose-multiplatform/actions/runs/10564736433/job/29267782682?pr=5122#step:4:360

I don't run web tests because it requires Chrome.  
I don't run android tests because it requires new version of the resource's library.

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/AndroidContextProvider.kt
terrakok: @igordmn the fuunction is in resource's package. I would say it is enough.
we already provide `PreviewContextConfigurationEffect` function with almost the same logic.

eymar (approved):

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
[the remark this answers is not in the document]
terrakok: Web tests with resources don't work at all (a known test framework issue)

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/AndroidContextProvider.kt
terrakok: https://youtrack.jetbrains.com/issue/CMP-6547

igordmn (approved):

PR #5123, opened 2024-08-26
Fix a resource reading on iOS 12

terrakok:
Use deprecated `seekToFileOffset` on iOS 12 to fix a crash on old Apple devices

Fixes https://youtrack.jetbrains.com/issue/CMP-5967

### Fixes - Resources
- _(prerelease fix)_ Fix a resource reading on iOS 12

### DIFF
--- a/components/resources/library/src/iosMain/kotlin/org/jetbrains/compose/resources/ResourceReader.ios.kt
+++ b/components/resources/library/src/iosMain/kotlin/org/jetbrains/compose/resources/ResourceReader.ios.kt
@@ -1,12 +1,19 @@
 package org.jetbrains.compose.resources
 
+import kotlinx.cinterop.BetaInteropApi
+import kotlinx.cinterop.ObjCObjectVar
 import kotlinx.cinterop.addressOf
+import kotlinx.cinterop.alloc
+import kotlinx.cinterop.memScoped
+import kotlinx.cinterop.ptr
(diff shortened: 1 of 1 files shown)


components/resources/library/src/iosMain/kotlin/org/jetbrains/compose/resources/ResourceReader.ios.kt
```diff
@@ -30,7 +41,11 @@ internal actual fun getPlatformResourceReader(): ResourceReader = object : Resou
 
     private fun readData(path: String, offset: Long, size: Long): NSData {
         val fileHandle = NSFileHandle.fileHandleForReadingAtPath(path) ?: throw MissingResourceException(path)
-        fileHandle.seekToOffset(offset.toULong(), null)
+        if (available(OS.Ios to OSVersion(major = 13))) {
+            fileHandle.seekToOffset(offset.toULong(), null)
```
ASalavei, line 45: I do understand that the code will be tricky, but handling errors when obj-c methods provide them, is a good practice:
```
        memScoped {
            val errorWrapper = alloc<ObjCObjectVar<NSError?>>()
            fileHandle.seekToOffset(offset.toULong(), errorWrapper.ptr).also {
                errorWrapper.value?.let {
                    error(it.localizedDescription) // Throw specific exception here
                }
            }
        }
```

ASalavei (approved):

PR #5128, opened 2024-09-02
Fix cocoapods integration

terrakok:
Use a correct file to configure the cocoapods resources integration instead of a relative path.

Fixes https://youtrack.jetbrains.com/issue/CMP-4303

### Fixes - Resources
- _(prerelease fix)_ Fix cocoapods resources integration which leaded to a lack resources in ios apps

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/IosResources.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/IosResources.kt
@@ -86,8 +86,8 @@ internal fun Project.configureSyncIosComposeResources(
     plugins.withId(COCOAPODS_PLUGIN_ID) {
         (kotlinExtension as ExtensionAware).extensions.getByType(CocoapodsExtension::class.java).apply {
             framework { podFramework ->
-                val syncDir = podFramework.getFinalResourcesDir().get().asFile.relativeTo(projectDir)
-                val specAttr = "['${syncDir.path}']"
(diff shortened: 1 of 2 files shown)


igordmn:
> Describe a change for adding it to https://github.com/JetBrains/compose-multiplatform/blob/master/CHANGELOG.md

Please, fix Release Notes

igordmn (approved):

tamimattafi:
Hello @terrakok!
I'm still having this issue on Kotlin 2.0.20 and CMP 1.7.0-beta02.

When I build Android, everything works as expected. However, iOS does crash due to missing resources.
```k
Uncaught Kotlin exception: org.jetbrains.compose.resources.MissingResourceException: Missing resource with path: ..../drawable/icon.xml
```

This crash goes away if I close Xcode, run `pod install`, then open the workspace again (As you mentioned [here](https://github.com/JetBrains/compose-multiplatform/issues/4720#issuecomment-2100898893)). 

But this solution with `pod install` isn't consistent, as it gets altered once I open Android Studio and do some operations that require clean and sync tasks.

My resources are in a separate module `<root>:presentation:resources`, which has composed plugin applied, this module is used inside `<root>:app:composeApp`, which is my shared module. It's linked to `iosApp` using CocoaPods (As a dynamic framework, `isStatic = false`)

I started facing this issue when I upgraded from Kotlin 1.9.24 and CMP 1.6.11. I used to have a script in Build Phase to copy resources, and it worked just fine. But I removed this script as I suppose multimodule support is now available since 2.0, and it became redundant.

Is there something I might have missed during migration?

PR #5131, opened 2024-09-06
Update images in Jetsnack example: host them locally and use Resources library

eymar:
We used to rely on an external service to fetch the images. But they are not available anymore


eymar:
Merging it to let the community use it properly. 

(https://youtrack.jetbrains.com/issue/CMP-6650)

PR #5132, opened 2024-09-09
Add TextField Scroller Sample

KunMinX:
Currently, there is a lack of examples on how to use a Scroller with TextField in the tutorials\Desktop_Components directory. Providing the correct example here will prevent developers from misusing it, which could lead to issues such as the cursor not automatically scrolling when it goes out of bounds.


kropp:
We are no longer maintaining tutorials in this repository

PR #5134, opened 2024-09-10
Add a redundancy warning to skiko.js for k/wasm target

eymar:
skiko.js used to be required for k/wasm projects. But some time ago it was refactored to use skiko.mjs, so skiko.js is not needed anymore (for k/wasm, still needed for k/js). But there are k/wasm  projects still including skiko.js in index.html which leads to extra loading time. The intention is to notify them without causing 404 errors for now.

## Testing
- updated a test
- tested manually

Screenshot 2024-09-10 at 15 34 41

### Highlights - Web
- `skiko.js` is redundant in case of K/Wasm Compose for Web applications and it can be removed from index.html files to not load redundant files. We are going to remove `skiko.js` from the k/wasm distribution in the future releases. `skiko.js` is still needed in case of K/JS Compose for Web apps.


ilgonmic (approved):

igordmn:
Changed:
```
We are going to remove `skiko.js` from the k/wasm distribution in the future releases
```
to:
```
skiko.js will be removed from the k/wasm distribution in the future releases
```

igordmn:
Applied to Release notes:
https://github.com/JetBrains/compose-multiplatform/pull/5142#discussion_r1781075051

PR #5138, opened 2024-09-14
Support K2 mode in the IJ plugin

rock3r:
This PR makes the IJ plugin run in K2 mode with the new Analysis APIs, instead of the old K1 APIs. It also includes cleaning up a bunch of compiler warnings, and some mistakes I saw when migrating to K2.

I have tested it on my own machine and it seems to work, both in IJ 243 EAP1 and 242.1, but I'd recommend you try it out too. Also, I don't have any Compose Web-based project to test the related changes with.

> [!IMPORTANT]
> Requires #5137 to be merged first.

### DIFF
--- a/idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/PreviewStateService.kt
+++ b/idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/PreviewStateService.kt
@@ -20,8 +20,8 @@ import javax.swing.JComponent
 import javax.swing.event.AncestorEvent
 import javax.swing.event.AncestorListener
 
-@Service(Service.Level.APP)
-class PreviewStateService() : Disposable {
+@Service(Service.Level.PROJECT)
+class PreviewStateService : Disposable {
     private val previewListener = CompositePreviewListener()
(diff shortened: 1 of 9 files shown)


idea-plugin/src/main/kotlin/org/jetbrains/compose/desktop/ide/preview/PreviewStateService.kt
```diff
@@ -35,7 +32,7 @@ class PreviewStateService(private val myProject: Project) : Disposable {
     init {
         val projectRefreshListener = ConfigurePreviewTaskNameCacheInvalidator(configurePreviewTaskNameCache)
         ExternalSystemProgressNotificationManager.getInstance()
-            .addNotificationListener(projectRefreshListener, myProject)
+            .addNotificationListener(projectRefreshListener, this)
```
rock3r, line 35: I'm not sure what is the expected behaviour here — the service was marked as an app-level service and the project used as disposable, but using Project as disposable is bad. It feels like the service should be project-level, so this subscription would maintain the same scope, but I don't know if it's the right way to go about it.

kropp (approved):

PR #5150, opened 2024-10-02
Fix of CMP-6802

InsanusMokrassar:
https://youtrack.jetbrains.com/issue/CMP-6802

**Bug Description**

In case if composition must be disposed, it clears all the data in container instead of data it has created

**Affected Platforms**

Select one or multiple affected platforms below:

- [x] Web (K/JS) - HTML library

**Versions**
- Compose Multiplatform version*: 1.7.0-rc01
- Kotlin version*: 2.0.20
- OS (name, version, arch): Any
- Browser (for Web issues): Any

**Reproduction Steps**

1. Clone https://github.com/InsanusMokrassar/IssueHtmlDomNodeWrapperReproducer_3d1c3e12
2. Call `jsBrowserDevelopmentExecutableDistribution` gradle task
3. Run test index.html in browser
4. Await while all body will be cleared (even script tag)

**Expected Behavior**

Only create by compose data cleared

### DIFF
--- a/html/compose-compiler-integration/src/jsTest/kotlin/AnonymousObjectsInComposable.kt
+++ b/html/compose-compiler-integration/src/jsTest/kotlin/AnonymousObjectsInComposable.kt
@@ -18,7 +18,7 @@ class AnonymousObjectsInComposable {
             content.Abc()
         }
 
-        assertEquals("<div>Abc</div>", document.body!!.firstElementChild!!.outerHTML)
+        assertEquals("<div>Abc</div>", document.body!!.lastElementChild!!.outerHTML)
     }
 
     @Test
@@ -30,7 +30,7 @@ class AnonymousObjectsInComposable {
             HasLocalClassWithComposable()
         }
 
-        assertEquals("<div>Abc2</div>", document.body!!.firstElementChild!!.outerHTML)
+        assertEquals("<div>Abc2</div>", document.body!!.lastElementChild!!.outerHTML)
     }
 
     @Test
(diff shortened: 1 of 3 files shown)


eymar:
We keep the [reported problem](https://youtrack.jetbrains.com/issue/CMP-6802) open for feedback, and keep the implementation remains the same. 

As for mermaid use case, I'll post the comment in the YT ticket. It's possible to integrate mermaid with Compose HTML without changes in the implementation of Compose HTML library.

For example:

```kotlin
@Composable
fun MermaidChart(
    chartSyntax: String,
    modifier: String = ""
) {
    var svgContent by remember { mutableStateOf("") }
    var errorMessage by remember { mutableStateOf<String?>(null) }

    LaunchedEffect(chartSyntax) {
        if (chartSyntax.isBlank()) {
            svgContent = ""
            errorMessage = null
            return@LaunchedEffect
        }

        try {
            val renderId = "mermaid-" + kotlin.random.Random.nextInt(1_000_000)
            val promise = js("window.mermaid.render(renderId, chartSyntax)") as Promise<dynamic>
            val result = promise.await()

            svgContent = result.svg as String
            errorMessage = null
        } catch (e: dynamic) {
            console.error("Mermaid parsing error:", e)
            errorMessage = e?.message?.toString() ?: "Invalid diagram syntax"
        }
    }

    Div(
        attrs = {
            if (modifier.isNotEmpty()) classes(modifier)
            prop<HTMLDivElement, String>({ element, value -> element.innerHTML = value }, svgContent)
        }
    )

    errorMessage?.let { msg ->
        Div(attrs = { style { color(Color.red) } }) {
            Text("Error: $msg")
        }
    }
}
```

PR #5152, opened 2024-10-02
Update Android target SDK to 35

MatkovIvan:
Should fix
```
Dependency 'androidx.compose.foundation:foundation-android:1.8.0-alpha02' requires libraries and applications that depend on it to compile against version 35 or later of the Android APIs.
```


igordmn (approved):

terrakok (approved):

avently:
Are you sure about this change? Because it will require everyone to update to SDK 15 which may not be a good idea. There is at least one year until Google Play will require an apk built against SDK 35.

Isn't updating the compile SDK to 35 enough (not target, just compile)?

PR #5167, opened 2024-11-25
Add to support Tajik language

zhimbura:
**Describe proposed changes and the issue being fixed**  
Added support for the Tajik language (`tg`) in the pluralization rules generation. Updated the rules based on CLDR to include pluralization logic specific to the Tajik language. This change ensures that the `tg` locale is handled correctly during resource generation.

Fixes: N/A (no linked issue)

---

## Testing

Unfortunately, I was unable to test the changes because the project cannot be indexed in my environment. It would be great if someone from the team could verify the correctness of this fix, including:
- Running the generation task using `./gradlew :resources:library:generatePluralRuleLists`.
- Checking the correctness of the generated pluralization rules for the Tajik language.

---

### Resources - Gradle Plugin
- **Feature:** Added support for the Tajik language (`tg`) in pluralization rules.


terrakok:
This file was generated automatically by GeneratePluralRuleListsTask.

We doesn't modify it manually.
Please, check if the new locale is supported by the Unicode (https://github.com/unicode-org/cldr) and then rerun the generation task.

zhimbura:
@terrakok 
There is 
https://youtrack.jetbrains.com/issue/CMP-7126/Add-support-for-the-Tajik-language-tg-in-the-pluralization-rules-generation.

terrakok:
The latest Unicode plurals don't support `tg` locale: https://github.com/unicode-org/cldr/blob/release-46/common/supplemental/plurals.xml

PR #5170, opened 2024-12-03
Refactor Compose Multiplatform benchmarks to remove code duplication.

pjBooms:
- kn+performance and jvm-vs-kotlin-native benchmarks are merged into one multiplatform where they measure both CPU time and missed frames + frames percentiles

- added multiplatform run arguments processing for benchmarks configuration

- deleted standalone samples (moved LazyList to the infrastructure)

Fixes [CMP-5790](https://youtrack.jetbrains.com/issue/CMP-5790/Refactor-CMP-benchmarks-to-remove-code-duplication)


benchmarks/multiplatform/build.gradle.kts
```diff
@@ -21,26 +23,21 @@ kotlin {
...
-                    "-linker-option", "-framework", "-linker-option", "Metal"
-                )
             }
         }
     }
 
     wasmJs {
-        browser()
         binaries.executable()
+        browser {
```
elijah-semyonov, line 39: Was something intended to be added there?

elijah-semyonov (approved):
LGTM

eymar (approved):
Thank you!
Can we swtich to kotlin 2.1.0? So wasmJs target would not fail to compile?

benchmarks/multiplatform/build.gradle.kts
pjBooms: Yes, it was. Fixed  :)

PR #5177, opened 2024-12-09
Drop support for versions lower than Gradle 8.0

Omico:
Since Compose Multiplatform 1.7.1, the minimum AGP version has been raised to 8.1.0. This means the minimum Gradle version is 8.0. Thus, we can safely remove this function.
https://www.jetbrains.com/help/kotlin-multiplatform-dev/whats-new-compose-170.html#minimum-agp-version-raised-to-8-1-0
https://developer.android.com/build/releases/past-releases/agp-8-1-0-release-notes

### DIFF
--- a/.github/workflows/gradle-plugin.yml
+++ b/.github/workflows/gradle-plugin.yml
@@ -16,9 +16,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, macos-14, windows-2022]
-        gradle: [8.0, 8.10.2]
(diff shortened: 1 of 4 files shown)


Omico:
@terrakok, I need some help. Is this still relevant? This prevents compilation, if we raise the Kotlin language version to 1.8. 
https://github.com/JetBrains/compose-multiplatform/blob/e91231144edc7a7ad4cac1878d9cbf407de6f2a4/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/tasks/AbstractJPackageTask.kt#L329-L344

PR #5178, opened 2024-12-09
Compose HTML: add an `AttrsScope.attr` overload for boolean attributes

ShreckYe:
The `AttrsScope.attr` member function supports bolean attributes via "For boolean attributes cast boolean value to String and pass it as value.". This feature turns out to be frequently used in this [compose-html-material](https://github.com/huanshankeji/compose-html-material) project I have been working on. So I'd like to propose to add this to the core library.

Reformatted AttributesTests.kt with IntelliJ IDEA BTW.

### DIFF
--- a/html/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
+++ b/html/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
@@ -246,3 +246,10 @@ open class AttrsScopeBuilder<TElement : Element>(
 private val setClassList: (HTMLElement, Array<out String>) -> Unit = { e, classList ->
     e.classList.add(*classList)
 }
+
+/**
+ * Adds a boolean attribute.
+ * @see AttrsScope.attr
+ */
+fun AttrsScope<*>.attr(attr: String, value: Boolean = true) =
+    attr(attr, value.toString())
--- a/html/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
+++ b/html/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
@@ -97,9 +97,16 @@ interface AttrsScope<out TElement : Element> : EventsListenerScope {
      * @param attr - the name of the attribute
      * @param value - the value of the attribute
      *
-     * For boolean attributes cast boolean value to String and pass it as value.
+     * For boolean attributes, use the other overload with a boolean [value] parameter, or cast boolean value to String and pass it as value.
      */
     fun attr(attr: String, value: String): AttrsScope<TElement>
+
+    /**
+     * Adds a boolean attribute.
+     * @see AttrsScope.attr
+     */
+    fun attr(attr: String, value: Boolean = true) =
+        attr(attr, value.toString())
 
     /**
      * [prop] allows setting values of element's properties which can't be set using [attr].
@@ -198,7 +205,7 @@ open class AttrsScopeBuilder<TElement : Element>(
      * @param attr - the name of the attribute
      * @param value - the value of the attribute
      *
-     * For boolean attributes cast boolean value to String and pass it as value.
+     * For boolean attributes, use the other overload with a boolean [value] parameter, or cast boolean value to String and pass it as value.
      */
     override fun attr(attr: String, value: String): AttrsScope<TElement> {
         attributesMap[attr] = value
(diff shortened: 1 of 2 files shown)


html/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
```diff
@@ -246,3 +246,10 @@ open class AttrsScopeBuilder<TElement : Element>(
 private val setClassList: (HTMLElement, Array<out String>) -> Unit = { e, classList ->
     e.classList.add(*classList)
 }
+
+/**
+ * Adds a boolean attribute.
+ * @see AttrsScope.attr
+ */
+fun AttrsScope<*>.attr(attr: String, value: Boolean = true) =
+    attr(attr, value.toString())
```
eymar, line 255: Any specific reason making it an extension function?
What do you think about adding this function to the `AttrsScope` itself?

Would you like to add a test for this new function in AttributesTests?

ShreckYe: Thanks for replying. Adding it as a member function looks good to me too. And sure I'd love to add some tests for this.

ShreckYe:
Hi, I have finished the requested changes. However, I have noticed something else that the `required` attribute extension function as well as some others don't stick to the documented way of passing boolean attributes, but pass an empty string to indicate the attribute is true instead:

https://github.com/JetBrains/compose-multiplatform/blob/e9597e71931dc38de8171605ce701efe8a6f4d8c/html/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/Attrs.kt#L197-L206

Is this the new recommended way that's not documented yet? Shall we update the KDocs and this added boolean `attr` function to not take an optional `value`parameter instead then?

eymar:
> Hi, I have finished the requested changes. However, I have noticed something else that the `required` attribute extension function as well as some others don't stick to the documented way of passing boolean attributes, but pass an empty string to indicate the attribute is true instead:
> 
> https://github.com/JetBrains/compose-multiplatform/blob/e9597e71931dc38de8171605ce701efe8a6f4d8c/html/core/src/jsMain/kotlin/org/jetbrains/compose/web/attributes/Attrs.kt#L197-L206
> 
> Is this the new recommended way that's not documented yet? Shall we update the KDocs and this added boolean `attr` function to not take an optional `value` parameter and pass an empty string instead then?

@ShreckYe , Good question! 
I didn't pay attention to that, but now I recall why it's made that way: https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML

> Note: The strings "true" and "false" are invalid values. To set the attribute to false, the attribute should be omitted altogether. **Though modern browsers treat any string value as true, you should not rely on that behavior.**

So passing "false" doesn't really make it "false". And the current implementation you made won't be correct. 
According to documentation, if we want the attribute value to be false we need to ommit the attribute itself. I guess adding an extra check in your new function would make it correct. Something like:  `if (value) attr(attr, "")`.

ShreckYe:
Thanks for replying and the references you provide. It explains everything.

> I guess adding an extra check in your new function would make it correct. Something like: `if (value) attr(attr, "")`.

For this, what I am thinking is just removing the `value` parameter in this function just like `required` does. For users, they can add a boolean variable themselves when needed and Compose will react to the changes. This makes things simpler. What do you think of this?

eymar:
> For this, what I am thinking is just removing the value parameter in this function just like required does. For users, they can add a boolean variable themselves when needed and Compose will react to the changes. This makes things simpler. What do you think of this?

It's already possible to achieve this by `if (flag) attr("readonly", "")`

But there is a problem in the KDoc - it contradicts the official doc about boolean attributes - https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes:
```
For boolean attributes cast boolean value to String and pass it as value.
```

I think that adding an `attrs` overload without a value could bring some risks where IDE auto-suggest a simpler method when someone expects a method with a value.

WDYT about this:

```kotlin
/**
 * Conditionally sets a boolean HTML attribute based on a condition.
 */
fun <TElement : Element> AttrsScope<TElement>.attr(
    name: String, 
    condition: Boolean
): AttrsScope<TElement> {
    if (condition) {
        attr(name, "")
    }
    return this
}
```

P.S. I apologise for no replies here.

ShreckYe:
Thanks for the detailed reply, and no worries about the delay.

I updated the PR to follow your sketch.

One related thing: `hidden()` still does `attr(HIDDEN, true.toString())`. Should we fix that in this PR? ([MDN ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden))

PR #5183, opened 2024-12-11
Add FontVariation.Settings support to the resources library

terrakok:
Sample:
```kotlin
Text(
    modifier = Modifier.padding(16.dp),
    fontFamily = FontFamily(
        Font(
            Res.font.RobotoFlex_VariableFont,
            variationSettings = FontVariation.Settings(FontVariation.weight(400))
        ),
    ),
    text = "The quick brown fox jumps over the lazy dog"
)
```

Fixes https://youtrack.jetbrains.com/issue/CMP-7088

## Testing
Demo app:
`./gradlew :resources:demo:desktopApp:run`
image

### Highlights - Resources
- Add FontVariation.Settings support to the resources library


terrakok:
New API introduced

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
```diff
@@ -45,4 +49,26 @@ actual fun Font(resource: FontResource, weight: FontWeight, style: FontStyle): F
...
+): Font {
+    val resourceReader = LocalResourceReader.currentOrPreview
+    val fontFile by rememberResourceState(resource, weight, style, variationSettings, { defaultEmptyFont }) { env ->
+        val path = resource.getResourceItemByEnvironment(env).path
+        val settingsKey = variationSettings.settings
+            .sortedBy { it::class.simpleName }
+            .joinToString(",") { it::class.simpleName + "_" + it.hashCode() }
+        val key = "$path:$weight:$style:$settingsKey"
+        fontCache.getOrLoad(key) {
+            val fontBytes = resourceReader.read(path)
```
eymar, line 69: Looks like this line creates a new ByteArray for every call. The fetch request will resolve from the cache, though. But still the new bytearray would require a new memory allocation (not only on web, but on all platforms).

A potential solution would be to add one more layer of caching - caching for raw bytes. 
But I'm not sure how often it would be needed. It would be very much useful when using the same variable font with many different variationSettings, or just different weight and style. 

 WDYT?

eymar (approved):

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
terrakok: It should be a rare operation in practice

components/gradle.properties
```diff
@@ -9,7 +9,7 @@ android.useAndroidX=true
 #Versions
 kotlin.version=1.9.24
 agp.version=8.2.2
-compose.version=1.7.0
+compose.version=1.8.0-alpha01
```
igordmn, line 12: Do we need this?

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
```diff
@@ -45,4 +49,26 @@ actual fun Font(resource: FontResource, weight: FontWeight, style: FontStyle): F
...
+    weight: FontWeight,
+    style: FontStyle,
+    variationSettings: FontVariation.Settings,
+): Font {
+    val resourceReader = LocalResourceReader.currentOrPreview
+    val fontFile by rememberResourceState(resource, weight, style, variationSettings, { defaultEmptyFont }) { env ->
+        val path = resource.getResourceItemByEnvironment(env).path
+        val settingsKey = variationSettings.settings
+            .sortedBy { it::class.simpleName }
+            .joinToString(",") { it::class.simpleName + "_" + it.hashCode() }
```
igordmn, line 66: -  `it.hashCode()` can cause collisions
- It seems that Setting.axisName is [unique](https://github.com/JetBrains/compose-multiplatform-core/blob/2cbd13c3c91e796df47bed6e21b7b225cf2c851e/compose/ui/ui-text/src/commonMain/kotlin/androidx/compose/ui/text/font/FontVariation.kt#L234) and can be used as the key

terrakok: settings can not contain same setting class, so I decided it is better then axisName. because different settings may contain the same value

igordmn: > settings can not contain same setting class

Hmm, I see that they can:
image

If I understand it correctly, we should replace `it::class.simpleName` by `axisName`

> because different settings may contain the same value

Indeed, I didn't take it into account. But `hashCode()` is unreliable. It seems we should use something like `toVariationValue(Density(1f))`:

```
variationSettings.settings
            .sortedBy { it.axisName }
            .joinToString(",") { it.axisName  + "_" + it.toVariationValue(Density(1f)) }
```

igordmn: Another option is to replace `key: String` by `key: Any` in `fontCache.getOrLoad`, and add equality checks

igordmn (approved):

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
terrakok: Done. with unit tests now

vividcodex:
Does it support mutable fonts like MiSans

adamglin0:
> Does it support mutable fonts like MiSans

Do u mean variable fonts?

terrakok:
https://youtrack.jetbrains.com/issues?q=project:CMP

PR #5189, opened 2024-12-17
CMP-7288 explicitly specify IJ platform compatibility range

kropp:
Explicitly specify compatible IJ platform builds: current version as a minimum and open range as maximum

Fixes https://youtrack.jetbrains.com/issue/CMP-7288

### DIFF
--- a/idea-plugin/build.gradle.kts
+++ b/idea-plugin/build.gradle.kts
@@ -33,7 +33,13 @@ dependencies {
 }
 
 intellijPlatform {
-    pluginConfiguration { name = "Compose Multiplatform IDE Support" }
+    pluginConfiguration {
+        name = "Compose Multiplatform IDE Support"
+        ideaVersion {
+            sinceBuild = "242.20224"
+            untilBuild = provider { null }
+        }
+    }
     buildSearchableOptions = false
     autoReload = false
 


idea-plugin/build.gradle.kts
```diff
@@ -33,7 +33,13 @@ dependencies {
 }
```
igordmn, line 33: We should cherry-pick it to `master` after merging.

(usually we merge to `master` first and cherry-pick to the release)

idea-plugin/build.gradle.kts
```diff
@@ -33,7 +33,13 @@ dependencies {
...
 
 intellijPlatform {
-    pluginConfiguration { name = "Compose Multiplatform IDE Support" }
+    pluginConfiguration {
+        name = "Compose Multiplatform IDE Support"
+        ideaVersion {
+            sinceBuild = "243.21565"
```
igordmn, line 39: It won't be compatible with Android Studio. The last stable in Toolbox seems `Build #AI-242.23339.11.2421.12700392`.

Can we use it?

kropp: Thanks, nice catch!
The latest stable Android Studio doesn't enable K2 mode by default, so it's less of an issue there.

PR #5190, opened 2024-12-18
Update GeneratedResClassSpec.kt

mgroth0:
Suppress "redundant visibility modifier" diagnostics so Res is compatible with `-Werror`, `-Wextra`, and IntelliJ inspections

Fixes [[CMP-7193]](https://youtrack.jetbrains.com/issue/CMP-7193/Generated-resource-accessors-contain-redundant-visibility-modifiers-causing-build-to-fail)

## Testing
A test may be added for this which confirms that the Res class compiles with `-Werror` and `-Wextra`.

## Release Notes

Section: - Fixes
Subsection: - Resources
Description: Generate `Suppress` annotation on generated Res class for redundant visibility modifier diagnostics


terrakok:
Test runs will be red because you are supposed to update tests as well

terrakok:
Download a zip archive with test reports here: https://github.com/JetBrains/compose-multiplatform/actions/runs/12399471669?pr=5190

terrakok:
https://github.com/JetBrains/compose-multiplatform/pull/5295

PR #5195, opened 2024-12-30
Support generated custom resources class name

Omico:
Assume we have two modules, as shown below.

### On app module

`app/build.gradle.kts`
```kotlin
dependencies {
    implementation(project(":core"))
}
```

### On core module

`core/build.gradle.kts`
```kotlin
compose {
    resources {
        publicResClass = true
    }
}
```

`core/src/kotlin/CoreRes.kt`
```kotlin
typealias CoreRes = Res
```

Now we cannot use `CoreRes.strings.xxx` like `Res.strings.xxx`.

So, we need to add the ability to let users define custom resource class name.

Usage:

```
compose {
    resources {
        resClassName = "CoreRes"
    }
}
```

### DIFF
--- a/gradle-plugins/.gitignore
+++ b/gradle-plugins/.gitignore
@@ -37,3 +37,5 @@ jacoco.exec
 # Layout Inspector creates it.
 captures/
 __pycache__
+
+.kotlin/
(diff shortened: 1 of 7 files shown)


dewantawsif:
@terrakok Could you please have this looked at? As for the failing test that seems to be ignored even in a recent merged PR #5260

adamglin0:
@terrakok @kropp

PR #5208, opened 2025-01-21
Lower mavenLocal priority in test projects of gradle-plugin

eymar:


MatkovIvan (changes requested):
It doesn't seem right. Locally built 0.0.0-SNAPSHOT should be used instead of one from dev repo

eymar:
@MatkovIvan I didn't change the repositories order for plugins. Only for libraries.

MatkovIvan (approved):
Not a big deal anyway. If it helps with CI cache issue - let's merge it

PR #5215, opened 2025-01-27
Benchmarks. Mimic the behavior of Skiko rendering

igordmn:
- Draw into an intermediate Picture. Without it, benchmarks showed an improvement by 8.61% between old/new layers, but in fact it was [a regression by 5.63%](https://docs.google.com/spreadsheets/d/1KQKDICvlmWDXtTqVoPOqIaqtXUPyQXA6RslTn79VW8o/edit?gid=406066323#gid=406066323):

image

- Clear Canvas each frame

- Add MavenLocal to becnhmark local versions

## Release Notes
N/A

### DIFF
--- a/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
+++ b/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
@@ -140,7 +140,7 @@ private val pictureRecorder = PictureRecorder()
  *
  * This is a simplified logic and it still can differ from the real cases:
  * - Rendering into an intermediate picture. Benchmarks can show incorrect results without it.
- *   For example, we had a case, when they showed an an improvement by 10%,
+ *   For example, we had a case when they showed an improvement by 10%,
  *   but there was a regression by 10%
  * - Clearing the canvas each frame
  *
--- a/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
+++ b/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
@@ -134,6 +134,19 @@ suspend fun measureComposable(
 
 private val pictureRecorder = PictureRecorder()
 
+/**
+ * Mimic Skiko render logic from https://github.com/JetBrains/skiko/blob/eb1f04ec99d50ff0bdb2f592fdf49711a9251aa7/skiko/src/awtMain/kotlin/org/jetbrains/skiko/SkiaLayer.awt.kt#L531
+ *
+ * This is very simplified logic, and it still can differ from the real cases.
+ *
+ * Though one main function - rendering into picture - was affecting performance.
+ *
+ * Benchmarks showed an improvement by 10%, but there was a regression by 10%.
+ *
+ * Beware that this logic can be changed in some new version of Skiko.
+ *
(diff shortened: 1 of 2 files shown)


benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
```diff
@@ -124,6 +127,20 @@ suspend fun measureComposable(
...
         scene.close()
+        surface.close()
         runGC() // cleanup for next benchmarks
     }
 }
+
+private val pictureRecorder = PictureRecorder()
+
+@OptIn(InternalComposeUiApi::class)
+fun ComposeScene.mimicSkikoRender(surface: Surface, time: Long, width: Int, height: Int) {
```
MatkovIvan, line 138: imo: requires BIG TODO to keep it in sync with skiko implementation, because I'm not sure that it won't be changed there

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
```diff
@@ -91,9 +94,9 @@ suspend fun measureComposable(
             val start = markNow()
 
             repeat(frameCount) {
-                val frameStart = start + nextVSync
+                val frameStart = markNow()
```
igordmn, line 97: From https://github.com/JetBrains/compose-multiplatform/pull/5216#discussion_r1930973719:

> We mark frameStart from nextVSync intentionally to catch possible GC. Consider revert this

igordmn: Is it by design that we also include the time of when we wait for vsync?

Originally I noticed it when I compared it to the average frame time, to understand how much frames deviates from it. Having additional time doesn't allow this comparison.

pjBooms: >Is it by design that we also include the time of when we wait for vsync?

Yes, It is by design of frames metrics that is greatly affected by GC: we need to count GC in fames measurements (because the benchmarks are also used for GC tuning) and if we count a frame from markNow we will loose GC that could happen after nextVSync and before actual work for a frame.

In general, this PR is very doubtable, we need to discuss it.

igordmn: Reverted to reduce the review scope

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
igordmn: Added a comment here

Without reusing the whole renderer from Skiko (which realistically we won't do without a good reason), we should keep it sync manually.

When we merge it, I will add a comment to Skiko:
```
Note, if you change the rendering logic a lot, notify about this in [this issue](new created issue), because [benchmarks project](link to this function) depends on it
```

pjBooms (approved):

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
igordmn: https://github.com/JetBrains/skiko/pull/1035

PR #5216, opened 2025-01-27
Benchmarks. Support GPU on Windows

igordmn:
Partially fixes https://youtrack.jetbrains.com/issue/CMP-7475/Benchmarks.-Add-ability-to-test-GPU-on-desktop-JVM (for Windows)

## Release notes
N/A

### DIFF
--- a/benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/GraphicsContext.desktop.kt
+++ b/benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/GraphicsContext.desktop.kt
@@ -1,0 +1,43 @@
+import org.jetbrains.skia.ColorSpace
+import org.jetbrains.skia.PixelGeometry
+import org.jetbrains.skia.Surface
+import org.jetbrains.skia.SurfaceColorFormat
+import org.jetbrains.skia.SurfaceOrigin
+import org.jetbrains.skia.SurfaceProps
+import org.jetbrains.skiko.ExperimentalSkikoApi
+import org.jetbrains.skiko.graphicapi.DirectXOffscreenContext
+import org.jetbrains.skiko.hostOs
+
+@OptIn(ExperimentalSkikoApi::class)
+fun graphicsContext(): GraphicsContext? = when {
+    hostOs.isWindows -> DirectXGraphicsContext()
+    else -> {
+        println("Unsupported desktop host OS: $hostOs. Using non-GPU graphic context")
+        null
+    }
+}
+
+@OptIn(ExperimentalSkikoApi::class)
+class DirectXGraphicsContext() : GraphicsContext {
+    // Note: we don't close `context` and `texture` after using,
+    // because it is created once in the main function
+    private val context = DirectXOffscreenContext()
+    private var texture: DirectXOffscreenContext.Texture? = null
+
+    override fun surface(width: Int, height: Int): Surface {
+        texture?.close()
+        texture = context.Texture(width, height)
+        return Surface.makeFromBackendRenderTarget(
+            context.directContext,
+            texture!!.backendRenderTarget,
+            SurfaceOrigin.TOP_LEFT,
+            SurfaceColorFormat.BGRA_8888,
+            ColorSpace.sRGB,
+            SurfaceProps(pixelGeometry = PixelGeometry.UNKNOWN)
+        ) ?: throw IllegalStateException("Can't create Surface")
+    }
+
+    override suspend fun awaitGPUCompletion() {
+        texture?.waitForCompletion()
+    }
+}
--- a/benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/main.desktop.kt
+++ b/benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/main.desktop.kt
@@ -13,6 +13,6 @@ fun main(args: Array<String>) {
         // Start the benchmark server to receive results from browsers
         BenchmarksSaveServer.start()
     } else {
-        runBlocking(Dispatchers.Main) { runBenchmarks() }
+        runBlocking(Dispatchers.Main) { runBenchmarks(graphicsContext = graphicsContext()) }
     }
 }


benchmarks/multiplatform/gradle.properties
```diff
@@ -1,4 +1,4 @@
-compose.version=1.7.1
```
igordmn, line 1: We shouldn't merge until we build compose with https://github.com/JetBrains/skiko/pull/1016.

The PR only works with custom Skiko right now:

```
// in skiko
checkout igor.demin/DirectXOffscreenContext
./gradlew publishToMavenLocal -deploy.version=722.0.0

// in benchmarks in build.gradle
    val desktopMain by getting {  
            dependencies {
                ...
                implementation("org.jetbrains.skiko:skiko:722.0.0-SNAPSHOT")  
                implementation("org.jetbrains.skiko:skiko-awt-runtime-macos-arm64:722.0.0-SNAPSHOT")  
```

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
```diff
@@ -91,7 +93,7 @@ suspend fun measureComposable(
             val start = markNow()
 
             repeat(frameCount) {
-                val frameStart = start + nextVSync
+                val frameStart = markNow()
```
pjBooms, line 96: We mark frameStart from nextVSync intentionally to catch possible GC. Consider revert this.

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
```diff
@@ -168,4 +168,14 @@ suspend fun runBenchmarks(
     runBenchmark("VisualEffects", width, height, targetFps, 1000, graphicsContext) { NYContent(width, height) }
     runBenchmark("LazyList", width, height, targetFps, 1000, graphicsContext) { MainUiNoImageUseModel()}
     runBenchmark("Example1", width, height, targetFps, 1000, graphicsContext) { Example1() }
-}
\ No newline at end of file
+}
+
+suspend fun benchmarksMain(
```
pjBooms, line 173: I would add graphicsContext?.close() directly to runBenchmarks

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
```diff
@@ -17,6 +17,8 @@ interface GraphicsContext {
     fun surface(width: Int, height: Int): Surface
 
     suspend fun awaitGPUCompletion()
+
+    fun close() = Unit
```
pjBooms, line 21: Don't we need to have `close` for macOS/iOS targets? 
BTW, what bad will happen if we don't close GraphicsContext?

pjBooms (changes requested):

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
igordmn: I accidentally added it in this PR. Let's move the discussion to https://github.com/JetBrains/compose-multiplatform/pull/5215/files#r1932225444

Will revert here

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
igordmn: > what bad will happen if we don't close GraphicsContext?

GPU memory won't be freed, even on GC.

Maybe it was also better to add an additional GC cleaning in Skiko, but generally GPU objects are better to clean when they are not needed.

igordmn: > Don't we need to have close for macOS/iOS targets?

We need actually. I will change the code.

I can free/close the Skia objects, but I can't free the ObjC objects, as Kotlin Native doesn't support that.

igordmn: > I can free/close the Skia objects

Though, there is no point in closing Skia objects if we have bigger issues with the graphic objects. Like `MTLTextureDescriptor` - it allocates a lot of bytes, and we don't free them. And we can't because we allocated them in Kotlin. But it will be freed on GC, and on low memory warning.

The fact that low memory warning takes graphic objects into account is a guess, as macOs/iOS uses a shared system/graphic memory. It wouldn't be the case with Windows/Linux, where memories are often physically separated, and so memory managers with own rules.

I'll keep the current code for simplicity. But for desktop we need to close because we have separated memories

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
igordmn: Reverted

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/MeasureComposable.kt
pjBooms: >GPU memory won't be freed, even on GC.

But we are quitting from the app just after close.

igordmn: Indeed, here we can skip closing. Then it is just my habit to assume it can be called anywhere, not necessarily in the end of the application.

I can remove it if you think it is better/cleaner without it.

igordmn: Removed to reduce the PR scope

benchmarks/multiplatform/gradle.properties
pjBooms: Maybe it is better to make the dependency  reflective for now?

However this way, we cannot compare with the old version anyway. I would add an option to choose between null/GPU/software rendering in addition, with backward compatible defaults. 

I would probably also wait GPU renderers for non-Windows desktop targets to have common code here.

igordmn:
@pjBooms, can we merge it now?

pjBooms (approved):

PR #5221, opened 2025-01-30
Update lwjgl-integration

Enaium:
It's working fine in Compose at `1.8.0-alpha02`

## Release Notes
N/A.


experimental/lwjgl-integration/src/main/kotlin/main.kt
```diff
@@ -1,6 +1,12 @@
-import androidx.compose.ui.ComposeScene
-import androidx.compose.ui.unit.Constraints
+/**
+ * @author Enaium
```
MatkovIvan, line 2: Not sure that it should be here

cc @igordmn to confirm

MatkovIvan (commented):
Usage of the reworked API of `ComposeScene` LGTM

douzooo:
Issue with this: Frame only updating when resizing window

System:
- MacOs 15
- M3 Pro

Description:
After clicking the button I only see the next frames when resizing the window

philip-segerfast:
On macos the composables only recompose when resizing the window, like douzooo said.
Also, the snapshot system doesn't work at all because it uses the AWT dispatcher which conflicts with GLFW's event queue (GLFW.glfwWaitEvents() never returns - is blocked by AWT).

I'm trying to make mod for a game that would allow you to use Compose to render UIs but this sadly doesn't work because the game uses GLFW.

If there are any workarounds I'd be very happy to hear about it.

System:
- MacOs 15
- M4 Pro

philip-segerfast:
Omg, I though I already tested that. It works!

kropp:

PR #5223, opened 2025-02-06
Changelog. Find PRs for not-squashed commits

igordmn:
Fixes https://youtrack.jetbrains.com/issue/CMP-7458/Changelog-script.-Change-PR-find-algorithm

It seems we can use `merge_commit_sha` field for it - it contains the top commit that was merged/rebased/squashed to jb-main

Example:
https://github.com/JetBrains/compose-multiplatform-core/pull/1808 was rebased on jb-main, and has `merge_commit_sha` of the top commit:

https://api.github.com/repos/JetBrains/compose-multiplatform-core/pulls?state=closed
```
    ...
    "html_url": "https://github.com/JetBrains/compose-multiplatform-core/pull/1808",
    "merge_commit_sha": "9fba10809a6816f778c526ea274c1f031b7d9cab",
```
image

# Testing
```
kotlin changelog.main.kts v1.8.0-alpha02..v1.8.0+dev2047
```
- Changelog for 1.8.0-alpha03 contains only more entries, not less
- it start to contain https://github.com/JetBrains/compose-multiplatform-core/pull/1771

# Release Notes
N/A

### DIFF
--- a/tools/changelog.main.kts
+++ b/tools/changelog.main.kts
@@ -40,6 +40,7 @@
 @file:DependsOn("com.google.code.gson:gson:2.10.1")
 
 import com.google.gson.Gson
+import com.google.gson.annotations.SerializedName
 import java.io.IOException
 import java.lang.ProcessBuilder.Redirect
 import java.net.URL
@@ -49,7 +50,6 @@ import java.time.LocalDate
 import java.time.format.DateTimeFormatter
 import java.util.*
 import kotlin.system.exitProcess
-import kotlin.text.substringAfterLast
 
 //region ========================================== CONSTANTS =========================================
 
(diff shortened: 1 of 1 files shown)


tools/changelog.main.kts
```diff
@@ -87,8 +87,10 @@ val argsKeyToValue = args
     .filter { it.contains("=") }
     .associate { it.substringBefore("=") to it.substringAfter("=") }
 
-val token = argsKeyToValue["token"] ?:
+val token = argsKeyToValue["token"]
+if (token != null) {
```
terrakok, line 91: is it what you want? here was a null check before

terrakok (approved):

tools/changelog.main.kts
igordmn: Yes.

If we write:
```
val token = null ?: println()
```

the token will be `Unit` and be passed as `Unit` in the http headers.

PR #5225, opened 2025-02-08
Fix will typos in `SplitPaneDSL`

michalsrutek:
Fix a couple of `wil` -> `will` typos.


MatkovIvan (approved):

kropp (changes requested):
I'd change these sentences from future to present tense.
There is also a typo at the end: `Heights` shouldn't be spelled like this.

PR #5228, opened 2025-02-10
Update changelog.main.kts, if (token == null) check

igordmn:
There was a mistake in the previous PR. The warning should be printed only if the token is omitted

# Release Notes
N/A


terrakok (approved):
I told you about that in the previous review)

PR #5232, opened 2025-02-12
Fix Maven Central publication tool

igordmn:
Fixes https://youtrack.jetbrains.com/issue/CMP-7603/Fix-Maven-Central-publication

- Update Space API to almost the latest
  - the most latest requires Ktor Eap lib
  - it is no longer depends on the broken `net.minidev:json-smart`
- Add artifact verification to avoid some security issues in the publication process

# Testing
The command from the issue passes configuration phase, and fails with:
```
java.lang.IllegalStateException: Provide value for 'maven.central.sign.key' Gradle property or 'MAVEN_CENTRAL_SIGN_KEY' environment variable
```

The actual test we can only do on CI, that artifacts are successfully uploaded to the stage.

# Release Notes
N/A


ci/build-helpers/buildSrc/build.gradle.kts
```diff
@@ -3,7 +3,7 @@ import org.gradle.kotlin.dsl.gradleKotlinDsl
 plugins {
     `java`
     `java-gradle-plugin`
-    kotlin("jvm") version "1.9.24"
+    kotlin("jvm") version "2.1.10"
```
igordmn, line 6: This is required by `org.jetbrains:space-sdk-jvm:2024.3-185883`

kropp (approved):

PR #5239, opened 2025-02-16
Expose APIs for working with resource environment

tamimattafi:
Expose APIs for working with resources environment for more control in the following cases:
1. Load localized strings inside iOS extension targets (e.g. Notification Extension):
`NSLocal` always returns default system language, even if we override `"AppLanguages"` in `NSUserDefaults.standardUserDefaults`. Passing a custom `ResourceEnvironment` to `getString` with the preferred locale fixes the issue.
2. Use a custom `ResourceEnvironment` to display a preview when customizing the App theme, without changing the configuration for the entire app. We can achieve this using `LocalComposeEnvironment`.
3. Use a custom `ResourceEnvironment` in dark-only screens, such as video players, or image editors. We can achieve this using `LocalComposeEnvironment`.
4. Use `copy` to change specific environment properties, and keep the rest as system default.

## Testing
- Changes were tested in the sample project (Code snippets are included in the PR).
- unit tests for this case already exist. They run successfully. Changes are unnecessary.

### DIFF
--- a/components/resources/demo/shared/src/commonMain/kotlin/org/jetbrains/compose/resources/demo/shared/ImagesRes.kt
+++ b/components/resources/demo/shared/src/commonMain/kotlin/org/jetbrains/compose/resources/demo/shared/ImagesRes.kt
@@ -1,23 +1,54 @@
 package org.jetbrains.compose.resources.demo.shared
 
 import androidx.compose.foundation.Image
-import androidx.compose.foundation.layout.*
+import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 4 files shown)


tamimattafi:
Hello, @terrakok! Could you please review this PR?

terrakok:
It is not enough to replace resources environment. We need to change global environment. So, it should be implemented in the runtime, but there is workaround for now:
https://youtrack.jetbrains.com/issue/CMP-4197#focus=Comments-27-11694865.0-0

PR #5240, opened 2025-02-17
Update Release Notes categories

igordmn:
- Rename `Breaking Changes` to `Migration Notes`
- Add Lifecycle/Navigation. We used them in [CHANGELOG.md](https://github.com/JetBrains/compose-multiplatform/blob/release/1.8.0-alpha03/CHANGELOG.md), and have it in the script, but didn't have in the template
- Describe each category
- Use platform categories as general ones. It contradicts with my comments [here](https://github.com/JetBrains/compose-multiplatform/pull/5231#discussion_r1953385844) and [here](https://github.com/JetBrains/compose-multiplatform-core/pull/1844#discussion_r1958445125), that it was only Compose related. But after thinking, we need some category for changes that don't fall under Compose or any other existing one. We can use platform categories (probably only Compose and some Compose components fall into this category right now)

## Release Notes
N/A

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,10 +20,10 @@ If we definitely shouldn't add Release Notes, add only N/A.
 Or enumerate sections, subsections and all changes.
 
 Possible sections:
-- Highlights             // new major features that need to be highlighted
-- Known Issues           // issues that has workaround and will be fixed
+- Highlights             // major features
+- Known Issues           // issues planned to be fixed, with possible workarounds
 - Migration Notes        // deprecations, experimental removals, minimal versions increases, behavior changes, compatibility breaking changes
-- Features               // new minor features. new API, behavior improvements
+- Features               // minor features
 - Fixes                  // bug fixes
 
 Possible subsections:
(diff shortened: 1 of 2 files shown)


MatkovIvan (approved):

terrakok (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -20,24 +20,27 @@ If we definitely shouldn't add Release Notes, add only N/A.
...
 
 Possible sections:
-- Highlights
-- Known Issues
-- Breaking Changes
-- Features
-- Fixes
+- Highlights             // major features
+- Known Issues           // issues planned to be fixed, with possible workarounds
+- Breaking Changes       // incompatible changes without deprecation cycle
```
kropp, line 25: `Breaking Changes` are still in the template, but removed from the script.
I guess we should prefer `Migration Notes`, but `Breaking Changes` should be supported by the script just in case

igordmn: Answered in https://jetbrains.slack.com/archives/G017NLN12D8/p1739985745739619?thread_ts=1739536033.270189&cid=G017NLN12D8

> should be supported by the script just in case

The script and template categories should match exactly — the script ensures everything aligns with the agreement.

kropp (approved):

PR #5246, opened 2025-02-26
Update shadow jar plugin, and code cleanup

Omico:
Please review each commit message for the details.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
@@ -29,8 +29,6 @@ import org.jetbrains.compose.web.WebExtension
 import org.jetbrains.compose.web.internal.configureWeb
 import org.jetbrains.kotlin.gradle.plugin.KotlinDependencyHandler
 
-internal val composeVersion get() = ComposeBuildConfig.composeVersion
-
 abstract class ComposePlugin : Plugin<Project> {
     override fun apply(project: Project) {
         val composeExtension = project.extensions.create("compose", ComposeExtension::class.java, project)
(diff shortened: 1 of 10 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -65,44 +63,44 @@ abstract class ComposePlugin : Plugin<Project> {
...
+        val html: HtmlDependencies = HtmlDependencies
+        val material = composeDependency("org.jetbrains.compose.material:material")
+        val material3 = composeDependency("org.jetbrains.compose.material3:material3")
+        val material3AdaptiveNavigationSuite = composeDependency("org.jetbrains.compose.material3:material3-adaptive-navigation-suite")
+        val materialIconsExtended = "org.jetbrains.compose.material:material-icons-extended:1.7.0-rc01"
+        val preview = composeDependency("org.jetbrains.compose.ui:ui-tooling-preview")
+        val runtime = composeDependency("org.jetbrains.compose.runtime:runtime")
+        val runtimeSaveable = composeDependency("org.jetbrains.compose.runtime:runtime-saveable")
+        val ui = composeDependency("org.jetbrains.compose.ui:ui")
+        val uiBackHandler = composeDependency("org.jetbrains.compose.ui:ui-backhandler")
```
MatkovIvan, line 81:
```suggestion
        @ExperimentalComposeLibrary
        val uiBackHandler = composeDependency("org.jetbrains.compose.ui:ui-backhandler")
```

Bigray2025 (commented):
I can I get the app coded here in my iOS device??

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -65,28 +63,34 @@ abstract class ComposePlugin : Plugin<Project> {
...
+        val runtimeSaveable = composeDependency("org.jetbrains.compose.runtime:runtime-saveable")
+        val ui = composeDependency("org.jetbrains.compose.ui:ui")
+
         @Deprecated("Use desktop.uiTestJUnit4", replaceWith = ReplaceWith("desktop.uiTestJUnit4"))
         @ExperimentalComposeLibrary
-        val uiTestJUnit4 get() = composeDependency("org.jetbrains.compose.ui:ui-test-junit4")
+        val uiTestJUnit4 = composeDependency("org.jetbrains.compose.ui:ui-test-junit4")
+
+        @ExperimentalComposeLibrary
+        val uiBackHandler = composeDependency("org.jetbrains.compose.ui:ui-backhandler")
```
MatkovIvan: I'd prefer not adding such new API at all.
1. We're not super happy of having such aliases at all. Probably most relevant task about it is [CMP-5861](https://youtrack.jetbrains.com/issue/CMP-5861)
2. We don't have plans to stabilize this module at all. It's opposite - it will be replaced to [navigationevent](https://developer.android.com/jetpack/androidx/releases/navigationevent) soon-ish

MatkovIvan (commented):
Combining cleanups and introducing new public API is not good idea.

PS I've assigned right people as reviewers

terrakok:
Could you explain your motivation for refactorings and version updates?
I don't see a reason for that at all. Yes, might be it is not perfect but there is no reasons to change it right now.

PR #5252, opened 2025-03-05
Changelog script. Remove incorrect Release Notes checks

igordmn:
Fixes https://youtrack.jetbrains.com/issue/CMP-7721/Changelog-script.-Remove-incorrect-Release-Notes-checks

Remove checks for missing and incorrect release notes.

These checks were applied on the script level, but no longer needed after we apply them before PR merge: https://github.com/JetBrains/compose-multiplatform/pull/5251

## Release Notes
N/A

### DIFF
--- a/tools/changelog.main.kts
+++ b/tools/changelog.main.kts
@@ -182,14 +182,14 @@ fun generateChangelog() {
 
             entries
                 .sortedBy { it.sectionOrder() }
-                .groupBy { it.sectionName() }
+                .groupBy { it.section }
                 .forEach { (section, sectionEntries) ->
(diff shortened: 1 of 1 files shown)


tools/changelog.main.kts
```diff
@@ -182,14 +182,14 @@ fun generateChangelog() {
 
             entries
```
igordmn, line 183: Should merge only after 1.8.0 Stable, as we still need to handle the previous PR without Release Notes

igordmn:
After create PR/Issue and thought, these check can still be valuable in a case we change the format, and need to check previous PR's

PR #5274, opened 2025-04-04
Lazy Grid Image Demo with three different implementations

pjBooms:
- CMP, native iOS, native Android with the same functionality

- supposed to be used to compare Compose Multiplatform performance metrics with native counter-parts.

Required for [CMP-7682](https://youtrack.jetbrains.com/issue/CMP-7682/Create-benchmarks-for-marketing-purposes).

## Testing
Manually

## Release Notes
N/A


benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setContent {
+            App()
+        }
+    }
+}
+
+@Composable
+fun App() {
```
ASalavei, line 46: Is it duplicate of the same `App` in common code? As far as I understand that does not affect performance.

ASalavei (approved):

benchmarks/showcases/LazyGridImageView/download_images.sh
```diff
@@ -0,0 +1,77 @@
...
+
+    echo "Downloading image $i/999: $filename"
+
+    # Use a specific image ID to ensure we get different images
+    # picsum.photos has images with IDs from 0 to about 1084
+    # We'll use modulo to cycle through available IDs if we need more than available
+    image_id=$((i % 1085))
+
+    # Download the image with specific ID, size 512x512, in JPG format
+    curl -s -L -o "$filepath" "https://picsum.photos/id/$image_id/512/512.jpg"
```
igordmn, line 27:
```suggestion
    curl -f -s -L -o "$filepath" "https://picsum.photos/id/$image_id/512/512.jpg"
```
(checked it on Windows)

Without it, it downloads https://picsum.photos/id/968/512/512.jpg (and 86 other images) on my machine, even if it returns 404

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+    modifier: Modifier = Modifier
+) {
+    Card(
+        modifier = modifier
+            .aspectRatio(1f) // Square aspect ratio
+            .fillMaxWidth(),
+        elevation = CardDefaults.cardElevation(defaultElevation = 4.dp)
+    ) {
+        val context = LocalContext.current
+        context.assets.open(imagePath).use { inputStream ->
```
igordmn, line 146: We should wrap `context.assets.open(imagePath).use ... bitmap.asImageBitmap()`  in `remember`. Otherwise Android benchmark can show incorrect results, as it can load the resource on every recomposition.

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+            val bitmap = BitmapFactory.decodeStream(inputStream)
```
igordmn, line 147: Did you consider using `painterResource(Int)` instead? It will be a closer analog to the CMP resources.

Ids can be found with:
```
context.resources.getIdentifier("name", "drawable", context.packageName)
```

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
```diff
@@ -0,0 +1,162 @@
...
+    val coroutineScope = rememberCoroutineScope()
+
+    // State for auto-scrolling
+    var autoScroll by remember { mutableStateOf(false) }
+    var scrollingDown by remember { mutableStateOf(true) }
+    var currentIndex by remember { mutableStateOf(0) }
+
+    val drawableResources = remember {
+        List(999) { index ->
+            val imageFilenames = context.assets.list("drawable") ?: emptyArray()
```
igordmn, line 58: Will you measure the startup time?

If so, this construction is rarely used, but it can affect performance, as it can read the list from the disk.

The same is true for `context.resources.getIdentifier` and `Res.allDrawableResources.values`.

I would move reading them out of `App`.

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
pjBooms: It is exactly the same app as in `composeApp` except resources usage. We just need to prove the statement that we are "native on Android" comparing native Android app using Jetpack Compose with  CMP App.

benchmarks/showcases/LazyGridImageView/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+* `/nativeiosApp` - Contains a native iOS implementation using SwiftUI
+  - Uses iOS asset catalog to load images
+
+## Downloading Images
+
+This project includes a script to download 999 different images 
+from [Picsum Photos](https://picsum.photos/) with 512x512 resolution:
+
+`./download_images.sh`
```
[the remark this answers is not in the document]
pjBooms: Do you mean the name of images from "image" to "downloadedImage"?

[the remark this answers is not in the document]
igordmn: Yes, so if we add to `.gitignore`, it will be a specific, not broad pattern

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
pjBooms: >Will you measure the startup time?

yes, of course

>I would move reading them out of App
hm, initially there was an idea to have a working app for non-complete dowloaded images.
I would think how to implement the same out of the app startup.

benchmarks/showcases/LazyGridImageView/download_images.sh
pjBooms: thanx, fixed

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
```diff
@@ -0,0 +1,162 @@
+package org.jetbrains.lazygridimage
+
+import android.graphics.BitmapFactory
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.layout.*
```
MatkovIvan, line 8: Should it be Compose for "Android Native"? Isn't it supposed to be xml views?

pjBooms: >Isn't it supposed to be xml views?

No, it is not. [See](https://docs.google.com/document/d/1GfyXCnKPX4eozPZ2IadArSZECgs6egKyp069Hm7wEHc/edit?tab=t.0)

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
pjBooms: But there is no way to get all drawable ids, right? CMP resources are stored in Android assets, so ideologically CMP resources are closer to Android ones, but in the sense of comparison assets may be better to use.

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
igordmn: My motivation was that app developers usually either:
1. download generic images from network using a networking API
2. accessing concrete image resources by known id (not by a string) using a resource API
3. open generic images from disk using file API (assets API on Android looks similar, but `context.filesDir` is used more for such case)

The first we can't do, because we can't measure it properly.

The second (the current way) probably we can do, if accessing by a string and listing don't impact performance. Seems they don't impact. But there can be questions from users if we don't mention about it (it looks suspicious). One of the options - explicitly exclude it, and mention that we didn't measure it. My first suggestion was for helping excluding this measurement from the main application.

The third will be the closest to real cases, but not sure if it is worth implementing. 

> sounds as overkill.

Agree

> if (UIImage(named: name) != nil)

It looks like we can just remove this check, because the images are always available?

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
igordmn: > CMP resources are stored in Android assets

It looks like implementation details. For an Android developer resources != assets

> But there is no way to get all drawable ids

We can hardcode "image$num"

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
pjBooms: Rewrote it to use painterResource

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
pjBooms: Ok, I rewrote it as you suggest

benchmarks/showcases/LazyGridImageView/nativeAndroidApp/app/src/main/java/org/jetbrains/lazygridimage/MainActivity.kt
pjBooms: Please check the latest version. Now only CMP version has a little performance overhead due to `allDrawableResources` usage (4ms on iOS, 30 ms on Android), I believe we can live with it.

igordmn: > are loaded from local assets

There is also an important detail - both [CMP resources](https://github.com/JetBrains/compose-multiplatform/blob/9c7841c601efd9eecf3af4fc6e8ab95f343cbdd4/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt#L158) and [Android resources](https://android.googlesource.com/platform/frameworks/base/+/fb302ccd8e0610a09691ea5503ff8111dc7a2e41/core/java/android/content/res/ResourcesImpl.java#493) cache decoded Bitmap in global memory. In CMP the cache is unlimited, on Android I am not sure, it can be limited by size. It behaves this way because it is designed for a case for predefined set of images, not for a case when we need to load arbitrary images from local assets.

Caching may skew the measurements.

If we want to avoid caching, contrary to my first comment, we shouldn't use a resource API everywhere.

benchmarks/showcases/LazyGridImageView/README.md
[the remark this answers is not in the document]
pjBooms: ok, done

PR #5275, opened 2025-04-09
Use png image instead of a vector image in AnimatedVisibility benchmark 

eymar:
It's necessary in preparation for D8 benchmarks, since D8 environment misses some APIs including a parser of xml.

Also, make sure the web targets process the asynchronous fetch resources operations by yielding the event loop. Otherwise, a benchmark would skip a part of the workload.

## Testing
N/A

## Release Notes
N/A


eymar:
I'm closing this PR in favor of https://github.com/JetBrains/compose-multiplatform/pull/5277 which contains the changes from this PR too.

PR #5279, opened 2025-04-11
Fix `./gradlew runRelease`, update ProGuard to 7.7.0

igordmn:
Fixes https://youtrack.jetbrains.com/issue/CMP-7577/Desktop-runRelease-crash-when-upgrade-to-CMP-1.8.0-alpha02
Fixes https://youtrack.jetbrains.com/issue/CMP-3818/Update-ProGuard-to-version-7.4-to-support-new-Java-versions
Fixes https://youtrack.jetbrains.com/issue/CMP-4883/Default-Proguard-rules-cause-release-task-failure-when-using-Material-3-instead-of-2-in-Compose-Desktop

Targeting 1.8.0

The rules was added by testing our current examples. The comprehensive rules are only possible if we run [all tests with ProGuard](https://youtrack.jetbrains.com/issue/CMP-3404/Run-all-Compose-tests-with-ProGuard-on).

Also, each rule wasn't investigated thoroughly, in some cases it might a valid rule, in some cases a ProGuard bug.

The auto integration checks on CI will be added in https://youtrack.jetbrains.com/issue/CMP-7970/Add-integration-checks-for-all-libraries.

## Testing
`./gradlew runRelease` works in:
- https://kmp.jetbrains.com
- all examples

### Features - Desktop
- The default ProGuard version is set to 7.7.0
   - If there are any new errors in the release build, update [the ProGuard rules](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-native-distribution.html#minification-and-obfuscation)
   - A usual workaround is to add `-keep class` for the associated with error class in "Location:"
   - If the error contains `androidx.` package, it might a Compose bug, please report in https://youtrack.jetbrains.com/issues/CMP. The `-keep class` workaround should also work in this case.
- _(prerelease fix)_ `./gradlew runRelease` doesn't crash with `java.lang.VerifyError`

### DIFF
--- a/gradle-plugins/compose/src/main/resources/default-compose-desktop-rules.pro
+++ b/gradle-plugins/compose/src/main/resources/default-compose-desktop-rules.pro
@@ -35,6 +35,9 @@
 -keep,allowshrinking,allowobfuscation class kotlinx.coroutines.Job { *; }
 -dontnote kotlinx.coroutines.**
 
+# org.jetbrains.kotlinx:kotlinx-coroutines-swing
+-keep class kotlinx.coroutines.swing.SwingDispatcherFactory
+
 # Kotlinx Datetime
 #   Material3 depends on it, and it references `kotlinx.serialization`, which is optional
 #   Copied from https://github.com/Kotlin/kotlinx-datetime/blob/v0.6.2/core/jvm/resources/META-INF/proguard/datetime.pro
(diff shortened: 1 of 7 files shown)


gradle-plugins/compose/src/main/resources/default-compose-desktop-rules.pro
```diff
@@ -56,4 +62,6 @@
 -dontwarn org.graalvm.compiler.core.aarch64.AArch64NodeMatchRules_MatchStatementSet*
 
 # Androidx
+-keep,allowshrinking,allowobfuscation class androidx.compose.runtime.SnapshotStateKt__DerivedStateKt { *; }
+-keep class androidx.compose.material3.SliderDefaults { *; }
```
igordmn, line 66: Failed in graphics-2d example

shishkin-pavel (approved):

PR #5281, opened 2025-04-14
Remove old Kotlin compatibility tests and related projects

terrakok:
## Release Notes
N/A


igordmn (approved):

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/KotlinCompatibilityTest.kt
```diff
@@ -1,100 +0,0 @@
...
-        "beforeKotlin2/custom-compiler", defaultTestEnvironment.copy(
-            kotlinVersion = "1.8.0",
-            composeCompilerPlugin = "\"androidx.compose.compiler:compiler:1.4.0\""
-        )
-    ).checkCustomComposeCompiler()
-
-    @Test
-    fun testSettingLatestCompiler() = testProject(
-        "beforeKotlin2/custom-compiler", defaultTestEnvironment.copy(
-            kotlinVersion = "1.8.20",
```
MatkovIvan, line 73: It seems weird that this particular one passed with the K2 update. Do you have any idea why?

MatkovIvan (approved):

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/KotlinCompatibilityTest.kt
terrakok: because it uses 1.8.20 and it works. It fails with 1.8.10

PR #5296, opened 2025-04-23
Gradle Plugin DSL to change the generated Res class name

terrakok:
Gradle Plugin DSL to change the generated Res class name:

```kotlin
compose.resources {
    nameOfResClass = "MyRes"
}
```

Fixes [CMP-7341](https://youtrack.jetbrains.com/issue/CMP-7341) Support generating custom resource class name
## Testing
Integration tests.

### Features - Resources
- Gradle Plugin DSL to change the generated Res class name

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResClassTask.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResClassTask.kt
@@ -9,12 +9,11 @@ import org.jetbrains.compose.internal.IdeaImportTask
 import java.io.File
 
 internal abstract class GenerateResClassTask : IdeaImportTask() {
-    companion object {
-        private const val RES_FILE_NAME = "Res"
-    }
-
     @get:Input
     abstract val packageName: Property<String>
+
+    @get:Input
+    abstract val resClassName: Property<String>
 
     @get:Input
     @get:Optional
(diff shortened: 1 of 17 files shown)


igordmn:
Requires 2 approvals, as this is new API in Gradle plugin

igordmn (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResClassTask.kt
```diff
@@ -9,13 +9,12 @@ import org.jetbrains.compose.internal.IdeaImportTask
...
 internal abstract class GenerateResClassTask : IdeaImportTask() {
-    companion object {
-        private const val RES_FILE_NAME = "Res"
-    }
-
     @get:Input
     abstract val packageName: Property<String>
 
+    @get:Input
+    abstract val resClassName: Property<String>
```
igordmn, line 16: Should we name it just `className`? To be consistent with `packageName` and to avoid double wording in `resources { res... = ``

igordmn:
Please, also fix the new failing check.

m-sasha (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResClassTask.kt
terrakok: I don't think so, because it changes the `Res` class name only.

terrakok: for the consistency I'll rename it to `nameOfResClass`

PR #5298, opened 2025-04-24
Refactor resource accessors generator to minify output source files.

terrakok:
Refactor resource accessors generator to minify output source files.

- Number of items per file was decreased to 100
- Common part of the path extracted to the private const
- Added imports
- Removed the private object with item's properties
- The file structure was reorganized

Fixes [CMP-7934](https://youtrack.jetbrains.com/issue/CMP-7934) Too large String0.commonMain.kt will break code insight feature in AS

## Testing
Integration tests.  
This should be tested by QA

### Fixes - Resources
- Fix IDE highlighting/resoulution when a generated file with resource accessors is too big


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
```diff
@@ -55,9 +55,8 @@ private fun ResourceType.requiresKeyName() =
     this in setOf(ResourceType.STRING, ResourceType.STRING_ARRAY, ResourceType.PLURAL_STRING)
```
igordmn, line 55: Please fix the failing CI before merging

igordmn (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
igordmn, line 55: Please add Release notes, because it fixes a bug

terrakok: the error is not related to changes

igordmn: It is better to always restart the checks, to be sure. These tests were failed for the first time in this PR:
image

igordmn: I was just talking about a good practice :).

The last build is the same stability now:
image

PR #5313, opened 2025-05-09
Update samples to Compose 1.8

kropp:
This PR only updates Compose and Kotlin versions.
I'm going to follow it up with further updates addressing various warnings.

## Testing
`./validateExamples.sh`, `./validateExamplesIos.sh`, and `./validateExamplesWithJs.sh`

## Release Notes
N/A


gradle-plugins/gradle.properties
```diff
@@ -8,7 +8,7 @@ kotlin.code.style=official
 dev.junit.parallel=false
 
 # Default version of Compose Libraries used by Gradle plugin
-compose.version=1.8.10+dev2347
+compose.version=1.8.0
```
eymar, line 11: It used to be `1.8.10+dev2347`. I'm wondering if the 1.8.10-dev was intentioally set here. 
This value affects the compose version used in the gradle plugin tests.

cc: @terrakok do we need 1.8.10-dev here?

eymar (approved):

gradle-plugins/gradle.properties
terrakok: it requires a new resources library for tests as I remember. it is available in 1.8.1

kropp: Reverted for now

PR #5314, opened 2025-05-09
Bundle kotlinx serialization ProGuard rules into the Compose plugin

igordmn:
When users use `androidx.navigation`, they see `@Serialization` annotation they can use, without explictly adding `kotlinx.serialization`. They see, because it is added as an `api` dependency.

The Compose Gradle plugin on the other hand provide `./gradlew runRelease` task that uses ProGuard to minify binaries. Because the plugin should support not only Compose, but also all support libraries (androidx, components), we should bundle serialization ProGuard rules into it.

Fixes https://youtrack.jetbrains.com/issue/CMP-8050

## Testing
```
import androidx.compose.ui.window.singleWindowApplication
import androidx.navigation.compose.NavHost
import androidx.navigation.compose.composable
import androidx.navigation.compose.rememberNavController

fun main() = singleWindowApplication {
    NavHost(
        navController = rememberNavController(),
        startDestination = LoginRoute()
    ) {
        composable<LoginRoute> {}
    }
}

sealed interface Route

@kotlinx.serialization.Serializable
data class LoginRoute(val id: Long? = null) : Route
```
Doesn't crash when run `./gradlew runRelease`

### Fixes - Desktop
- Fix "Serializer for class is not found" using `androidx.navigation` and running `./gradlew runRelease`
- `kotlinx.serialization` ProGuard rules are bundled in the Compose Gradle plugin

### DIFF
--- a/gradle-plugins/compose/src/main/resources/default-compose-desktop-rules.pro
+++ b/gradle-plugins/compose/src/main/resources/default-compose-desktop-rules.pro
@@ -13,7 +13,6 @@
 
 # Kotlinx Coroutines Rules
 # https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/resources/META-INF/proguard/coroutines.pro
-
 -keepnames class kotlinx.coroutines.internal.MainDispatcherFactory {}
 -keepnames class kotlinx.coroutines.CoroutineExceptionHandler {}
 -keepclassmembers class kotlinx.coroutines.** {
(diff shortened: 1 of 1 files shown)


terrakok (approved):

igordmn:
> Can we re-use ones from AOSP

I am not aware of androidx-specific rules besides small `# Androidx` section. It is not worth to implement just for it. The issues are with external dependencies (Kotlin/Kotlinx).

https://youtrack.jetbrains.com/issue/CMP-7777/Support-ProGuard-rules-bundled-in-libraries should allow to read the bundled rules.

PR #5315, opened 2025-05-10
Add ResourceConfiguration API.

TBSten:

(Optional) Fixes [link to the issue]

- [CMP-8134](https://youtrack.jetbrains.com/issue/CMP-8134/Switch-the-ClassLoader-used-by-ResourceReader)

## Testing
(Optional) Describe how you tested your changes (provide a snippet or/and steps)

- [x] Add Test to **desktopTest** SourceSet

(Optional) This should be tested by QA

### Features - Resources
- Add `ProvideResourceConfiguration` for change ClassLoader used to load resources on desktop and android


components/resources/library/src/desktopTest/resources/resource-sample-shared-desktop.jar
MatkovIvan, line 1: ~8MB of unknown jar doesn't seem good

MatkovIvan (changes requested):

PR #5319, opened 2025-05-14
KMT-879: Make `org.jetbrains.compose.ui.tooling.preview.PreviewParameterProvider` an actual implementation of `androidx.compose.ui.tooling.preview.PreviewParameterProvider` on Android

boiler23:
This is required for the preview parameters to be correctly picked up by the preview adapter in IDE, which expects the provider to be a subclass of AndroidX's interface, and fails with the `ClassCastException` otherwise.

### Fixes - Multiple Platforms
- Support Preview parameters for Previews in common source sets in IJ and AS. Note: IDEs also need to implement support on their end. Please check the respective IDE release notes to confirm this is supported.

Example usage:

```
import androidx.compose.runtime.Composable
import org.jetbrains.compose.ui.tooling.preview.Preview
import org.jetbrains.compose.ui.tooling.preview.PreviewParameter
import org.jetbrains.compose.ui.tooling.preview.PreviewParameterProvider

class MyPreviewParameterProvider : PreviewParameterProvider<String> {
    override val values = sequenceOf("Hello, Compose!", "Hello, World!")
}

/**
 * This function will generate two preview images with different texts.
 */
@Preview
@Composable
fun MyPreview(@PreviewParameter(MyPreviewParameterProvider::class) text: String) {
    Text(text)
}
```


components/ui-tooling-preview/library/src/commonMain/kotlin/org/jetbrains/compose/ui/tooling/preview/PreviewParameter.kt
```diff
@@ -22,7 +22,7 @@ import kotlin.reflect.KClass
  * Interface to be implemented by any provider of values that you want to be injected as @[Preview]
  * parameters. This allows providing sample information for previews.
  */
-interface PreviewParameterProvider<T> {
+expect interface PreviewParameterProvider<T> {
```
igordmn, line 25: In Compose Multiplatform itself we have a guideline to not make expect/actual API if we can, and encourage making truly common interface without expect/actual. It introduces some issues:

- it fragmentates the ecosystem. If the expect interface diverges between platforms, clients of this platform have to start thinking about every supported platform, instead of just thinking platform-independent way.

- it restricts extensibility. After making it as typealias on Android, we won't be able to extend this interface for more common purposes further, and will be forever bound to the Android implementation.

Considering these points, a more correct approach is to not have this as `androidx.compose.ui.tooling.preview.PreviewParameterProvider` on Android. It is to support `org.jetbrains.compose.ui.tooling.preview.PreviewParameterProvider` in the Android Preview.

Of course, we can make an exception. But we should be conscious about it, if we truly think it is difficult to go the suggested way.

components/gradle/libs.versions.toml
```diff
@@ -4,6 +4,7 @@ androidx-appcompat = "1.6.1"
 androidx-activity-compose = "1.8.2"
 androidx-test = "1.5.0"
 androidx-compose = "1.6.1"
+ui-tooling-preview-android = "1.8.1"
```
MatkovIvan, line 7: I guess `androidx-compose` should be re-used instead

components/ui-tooling-preview/library/src/commonMain/kotlin/org/jetbrains/compose/ui/tooling/preview/PreviewParameter.kt
boiler23: That was my initial intention, yes.

However, it turns out that the rendering engine is quite heavily coupled to receive exactly `androidx`'s preview provider.

For reference:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui-tooling/src/androidMain/kotlin/androidx/compose/ui/tooling/ComposeViewAdapter.android.kt;l=530
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui-tooling/src/androidMain/kotlin/androidx/compose/ui/tooling/PreviewUtils.android.kt;l=25

Modifying this code to support jb preview parameters would also need considerable effort and approval from Google, which I doubt we can obtain quickly (if possible at all).

So, I think an exception here is very reasonable.

components/gradle/libs.versions.toml
boiler23: But why is its version so outdated? 1.6.1 is quite an old one :)

components/ui-tooling-preview/library/src/androidMain/kotlin/org/jetbrains/compose/ui/tooling/preview/PreviewParameter.android.kt
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jetbrains.compose.ui.tooling.preview
+
+/**
+ * On Android, we just typealias to the androidx.compose.ui.tooling.preview.PreviewParameterProvider.
```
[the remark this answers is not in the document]
boiler23: done

components/ui-tooling-preview/library/src/desktopMain/kotlin/org/jetbrains/compose/ui/tooling/preview/PreviewParameter.desktop.kt
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Interface to be implemented by any provider of values that you want to be injected as @[Preview]
+ * parameters. This allows providing sample information for previews.
+ */
+actual interface PreviewParameterProvider<T> {
+  /**
+   * [Sequence] of values of type [T] to be passed as @[Preview] parameter.
+   */
+  actual val values: Sequence<T>
```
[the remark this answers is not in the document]
boiler23: done

components/gradle/libs.versions.toml
boiler23: I removed the new version reference, but I'm still interested to learn why 1.6.1 is used :)

MatkovIvan: There was no reason for update, feel free to bump if required

components/ui-tooling-preview/library/src/appleMain/kotlin/org/jetbrains/compose/ui/tooling/preview/PreviewParameter.apple.kt
```diff
@@ -0,0 +1,34 @@
+/*
```
pjBooms, line 1: If all non-Android implementations are the same, it is better to introduce nonAndroidMain source set an place the impl there.

components/gradle/libs.versions.toml
```diff
@@ -4,6 +4,7 @@ androidx-appcompat = "1.6.1"
 androidx-activity-compose = "1.8.2"
```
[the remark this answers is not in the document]
boiler23: Updated the release notes, please take a look if that works better now.

components/ui-tooling-preview/library/src/commonMain/kotlin/org/jetbrains/compose/ui/tooling/preview/PreviewParameter.kt
pjBooms: As far as I remember, in long term, we are going to commonize `androidx.compose.ui.tooling.preview.Preview` and deprecate `org.jetbrains.compose.ui.tooling.preview.Preview`. So it sounds reasonable not to bother Google with something we are going to deprecate in the future.

components/ui-tooling-preview/library/src/appleMain/kotlin/org/jetbrains/compose/ui/tooling/preview/PreviewParameter.apple.kt
boiler23: good point, changed this.

components/ui-tooling-preview/library/src/commonMain/kotlin/org/jetbrains/compose/ui/tooling/preview/PreviewParameter.kt
```diff
@@ -31,7 +31,7 @@ interface PreviewParameterProvider<T> {
     /**
      * Returns the number of elements in the [values] [Sequence].
      */
-    val count get() = values.count()
+    open val count: Int
```
MatkovIvan, line 34: Why explicit `open` is required here?

boiler23: Otherwise, it's not compatible with the androidx alias.
I assume this is due to the default implementation of `count` there.

MatkovIvan (approved):

components/ui-tooling-preview/library/build.gradle.kts
```diff
@@ -52,3 +52,7 @@ configureMavenPublication(
     name = "Experimental Compose Multiplatform tooling library API. This library provides the API required to declare " +
             "@Preview composables in user apps."
 )
+
+dependencies {
+    compileOnly(libs.androidx.ui.tooling.preview.android)
```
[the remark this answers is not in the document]
boiler23: Changed to implementation, as we discussed in DMs.

components/gradle/libs.versions.toml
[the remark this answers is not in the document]
igordmn: Thanks, I rewrote it more, please check

components/ui-tooling-preview/library/src/commonMain/kotlin/org/jetbrains/compose/ui/tooling/preview/PreviewParameter.kt
igordmn: Complications of making this expect seems not major:
- we can't extend `PreviewParameterProvider`. But the Android one didn't change from the very beginning
- [we have](https://github.com/JetBrains/compose-multiplatform/pull/5319#discussion_r2096776670) to expose 2 API in `androidMain`. But current users already see 2 APIs

If anything goes wrong, we still can revert this change without breaking compatibility with IDEs.

igordmn:
LGTM after the change `implementation -> api`

components/ui-tooling-preview/library/build.gradle.kts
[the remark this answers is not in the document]
igordmn: @MatkovIvan, as it is new "API", please review it as well

igordmn (approved):

components/ui-tooling-preview/library/build.gradle.kts
```diff
@@ -52,3 +73,7 @@ configureMavenPublication(
     name = "Experimental Compose Multiplatform tooling library API. This library provides the API required to declare " +
             "@Preview composables in user apps."
 )
+
+dependencies {
+    api(libs.androidx.ui.tooling.preview)
+}
```
MatkovIvan, line 79: Could you please move it into `sourceSets { androidMain {} }` block?

boiler23: done

MatkovIvan (approved):

PR #5327, opened 2025-05-30
Implement saving benchmark results to JSON for browser

pjBooms:
Fixes [CMP-8282](https://youtrack.jetbrains.com/issue/CMP-8282/Benchmarks-support-saveStatsToJSON-for-web)

## Release Notes
N/A

### DIFF
--- a/benchmarks/multiplatform/benchmarks/build.gradle.kts
+++ b/benchmarks/multiplatform/benchmarks/build.gradle.kts
@@ -70,6 +70,9 @@ kotlin {
                 implementation(libs.kotlinx.serialization.json)
                 implementation(libs.kotlinx.io)
                 implementation(libs.kotlinx.datetime)
+                implementation(libs.ktor.client.core)
+                implementation(libs.ktor.client.content.negotiation)
+                implementation(libs.ktor.serialization.kotlinx.json)
             }
         }
 
@@ -77,6 +80,19 @@ kotlin {
             dependencies {
                 implementation(compose.desktop.currentOs)
                 runtimeOnly(libs.kotlinx.coroutines.swing)
+                implementation(libs.ktor.server.core)
+                implementation(libs.ktor.server.netty)
+                implementation(libs.ktor.server.content.negotiation)
+                implementation(libs.ktor.client.java)
+                implementation(libs.ktor.server.cors)
+            }
+        }
+
+        val wasmJsMain by getting {
+            dependencies {
+                implementation(libs.ktor.client.js)
+                implementation(libs.ktor.client.content.negotiation)
+                implementation(libs.ktor.serialization.kotlinx.json)
             }
         }
     }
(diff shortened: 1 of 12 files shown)


benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/main.desktop.kt
```diff
@@ -8,5 +8,11 @@ import kotlinx.coroutines.runBlocking
 
 fun main(args: Array<String>) {
     Config.setGlobalFromArgs(args)
-    runBlocking(Dispatchers.Main) { runBenchmarks() }
+
+    if (Config.runServer) {
+        // Start the benchmark server to receive results from browsers
+        BenchmarksSaveServer.start()
```
eymar, line 14: what do you think about creating a separate module for a server? So the `benchmarks` module won't be responsible for anything else than running the benchmarks.

Then the `runBrowserAndSaveStats` would run the server task on a different module:

For example
```
serverProcess = ProcessBuilder("./gradlew", "startCollectorServer:run", "-PrunArguments=runServer=true saveStatsToJSON=true")
```

benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/BenchmarksSaveServer.kt
```diff
@@ -0,0 +1,93 @@
...
+                    withContext(Dispatchers.IO) {
+                        if (Config.saveStatsToJSON) {
+                            val jsonPath = Path("build/benchmarks/json-reports/${result.name}.json")
+                            SystemFileSystem.createDirectories(jsonPath.parent!!)
+                            SystemFileSystem.sink(jsonPath).writeText(result.stats)
+                            println("JSON results saved to ${SystemFileSystem.resolve(jsonPath)}")
+                        }
+
+                        if (Config.saveStatsToCSV) {
+                            // TODO: for CSV, we would need to convert JSON to the values
```
eymar, line 71: let's add a println here with explanation that csv is not supported yet?

benchmarks/multiplatform/benchmarks/build.gradle.kts
```diff
@@ -142,6 +158,41 @@ tasks.register("buildD8Distribution", Zip::class.java) {
...
+    doFirst {
+        var serverProcess: Process? = null
+        var clientProcess: Process? = null
+        try {
+            serverProcess = ProcessBuilder("./gradlew", "benchmarks:run", "-PrunArguments=runServer=true saveStatsToJSON=true")
+                .start()
+            printProcessOutput(serverProcess.inputStream)
+            printProcessOutput(serverProcess.errorStream)
+
+            clientProcess = ProcessBuilder("./gradlew", "benchmarks:wasmJsBrowserProductionRun")
```
eymar: looks like `saveStatsToJSON=true` is not passed here, the browser doesn't make `/benchmark` requests

benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/main.desktop.kt
pjBooms: I thought about this but in the end came into conclusion that adding it to current sources are easier:

1. Some save logic is shared in this case.
2. Communication port is shared

It seems that extracting a separate module is a bit overkill

benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/BenchmarksSaveServer.kt
pjBooms: fixed

benchmarks/multiplatform/benchmarks/build.gradle.kts
pjBooms: fixed

benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/BenchmarksSaveServer.kt
```diff
@@ -0,0 +1,95 @@
...
+
+        server = embeddedServer(Netty, port = port) {
+            install(ContentNegotiation) {
+                json()
+            }
+            install(CORS) {
+                allowMethod(HttpMethod.Get)
+                allowMethod(HttpMethod.Post)
+                allowHeader(HttpHeaders.ContentType)
+                allowHost("localhost:8080")
```
eymar: perhaps using `anyHost()` would help avoid the collisions when the web benchmarks run on a different port?

https://ktor.io/docs/server-cors.html#hosts

pjBooms: done

ApoloApps:
Are there any plans to integrate benchmarks module for use in Compose gradle plugin (Multiplatform). This would let us benchmark our own components directly inside the plugin

PavelPunegov (approved):

pjBooms:
> Are there any plans to integrate benchmarks module for use in Compose gradle plugin (Multiplatform). This would let us benchmark our own components directly inside the plugin

No, there are no such plans so far

PR #5330, opened 2025-06-06
Replace `KotlinPlatformType` check with `KotlinJvmTarget` in JVM target configuration.

terrakok:
Problem is the new AGP androidLibrary target has `platformType == KotlinPlatformType.jvm` as well.

Fixes https://youtrack.jetbrains.com/issue/CMP-8325 

## Testing
Unit tests

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/JvmApplicationContext.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/JvmApplicationContext.kt
@@ -13,9 +13,9 @@ import org.jetbrains.compose.desktop.application.dsl.JvmApplicationBuildType
 import org.jetbrains.compose.internal.KOTLIN_JVM_PLUGIN_ID
 import org.jetbrains.compose.internal.KOTLIN_MPP_PLUGIN_ID
 import org.jetbrains.compose.internal.javaSourceSets
-import org.jetbrains.compose.internal.utils.joinDashLowercaseNonEmpty
 import org.jetbrains.compose.internal.mppExt
-import org.jetbrains.kotlin.gradle.plugin.KotlinPlatformType
+import org.jetbrains.compose.internal.utils.joinDashLowercaseNonEmpty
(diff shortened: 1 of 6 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/JvmApplicationContext.kt
```diff
@@ -56,7 +56,7 @@ internal data class JvmApplicationContext(
         if (project.plugins.hasPlugin(KOTLIN_MPP_PLUGIN_ID)) {
             var isJvmTargetConfigured = false
             project.mppExt.targets.all { target ->
-                if (target.platformType == KotlinPlatformType.jvm) {
+                if (target is KotlinJvmTarget) {
```
Tapchicoma, line 59: Could compose Gradle plugin be used with in projects using Kotlin/JVM projects? If yes - better to also check for `KotlinWithJavaTarget` target type.

igordmn: Yes, it can. @terrakok, let's add it

terrakok: @Tapchicoma yes, that's why I ask to review. could you clarify how to use this target to add a test for that?

Tapchicoma: This target is only used by "org.jetbrains.kotlin.jvm" plugin. So any test with this plugin should work.

terrakok: well. in the code above:
```
if (project.plugins.hasPlugin(KOTLIN_MPP_PLUGIN_ID)) {
```
I guess it should work then

alorma:
Any ETA on when this fix will be available?

PR #5334, opened 2025-06-13
Make ResourceReader a public API

lamba92:
Fixes CMP-8134 Switch the ClassLoader used by ResourceReader

This change makes the `ResourceReader` interface and `DefaultResourceReader` public, allowing users to provide custom resource loading implementations.

The internal `getPlatformResourceReader()` function has been removed in favor of the public `DefaultResourceReader` property.


components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceReader.android.kt
```diff
@@ -8,7 +8,7 @@ import androidx.compose.runtime.ProvidableCompositionLocal
 import java.io.FileNotFoundException
 import java.io.InputStream
 
-internal actual fun getPlatformResourceReader(): ResourceReader = object : ResourceReader {
```
rock3r, line 11: I would avoid unnecessary breaking API changes here

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceReader.desktop.kt
```diff
@@ -30,18 +36,14 @@ internal actual fun getPlatformResourceReader(): ResourceReader = object : Resou
...
-        val classLoader = getClassLoader()
-        return classLoader.getResourceAsStream(path) ?: throw MissingResourceException(path)
-    }
+    private fun getResourceAsStream(path: String): InputStream =
+        classLoader.getResourceAsStream(path) ?: throw MissingResourceException(path)
 
-    private fun getClassLoader(): ClassLoader {
-        return this.javaClass.classLoader ?: error("Cannot find class loader")
-    }
 }
```
rock3r, line 52: Nit: add missing empty line at EOF while you're at it

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceReader.android.kt
lamba92: The function was used only by `DefaultResourceReader`. I believe it is not necessary to have a function just to be used a line after. `DefaultResourceReader` is indeed used everywhere.

rock3r: I'd still avoid changing a function that creates an object to a val that looks from the outside like an accessor (hiding expensive operations behind a val is an anti-pattern)

components/resources/library/build.gradle.kts
```diff
@@ -214,6 +214,18 @@ apiValidation {
...
+
+tasks {
+    val desktopTestProcessResources =
+        named<ProcessResources>("desktopTestProcessResources")
+
+    withType<Test> {
+        dependsOn(desktopTestProcessResources)
+        environment("RESOURCES_PATH", desktopTestProcessResources.map { it.destinationDir.absolutePath }.get())
+    }
+}
```
rock3r, line 229: Can you add back the empty line?

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceReader.android.kt
rock3r: I don’t like that this is a `get()` in the JS variant that does work. I'd either keep it as function everywhere, or change the JS variant to also be init only once

components/resources/library/src/desktopTest/kotlin/org/jetbrains/compose/resources/CustomClassloaderTest.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    val RESOURCES_PATH
+        get() = System.getenv("RESOURCES_PATH")
+            ?.let { Path(it) }
+            ?.takeIf { it.isDirectory() }
+            ?: error("RESOURCES_PATH environment variable is not set or is not a directory")
+
+    @Test
+    fun testCustomClassloader() = runTest {
+
```
rock3r: Nit:
```suggestion
```

lamba92:
> When you open the ResourceReader API to public it means you finalize a format of the internal strings serialization format:
> 
> ```
> suspend fun readPart(path: String, offset: Long, size: Long): ByteArray
> ```
> 
> that is why I haven't been ready to do that. The format was invented on the go and I haven't think to stabilize it yet.
> 
> One of the problem with these offset and size: [CMP-5039](https://youtrack.jetbrains.com/issue/CMP-5039) Overriding string resources causes crashes or inelligible strings

We could annotate `ResourceReader` as internal. Alternatively we could provide a `Configuration` in the local composition for resource loaders. Wdyt?

components/resources/library/build.gradle.kts
```diff
@@ -11,6 +14,7 @@ plugins {
 }
 
 kotlin {
+
```
rock3r: Nit:
```suggestion
```

components/resources/library/src/desktopTest/kotlin/org/jetbrains/compose/resources/CustomClassloaderTest.kt
```diff
@@ -0,0 +1,40 @@
...
+class CustomClassloaderTest {
+
+    val RESOURCES_PATH
+        get() = System.getenv("RESOURCES_PATH")
+            ?.let { Path(it) }
+            ?.takeIf { it.isDirectory() }
+            ?: error("RESOURCES_PATH environment variable is not set or is not a directory")
+
+    @Test
+    fun testCustomClassloader() = runTest {
```
rock3r, line 23: Do you need `runTest`? I don't see any coroutine APIs being used in the test

components/resources/library/src/iosMain/kotlin/org/jetbrains/compose/resources/ResourceReader.ios.kt
```diff
@@ -48,11 +50,13 @@ internal actual fun getPlatformResourceReader(): ResourceReader = object : Resou
     }
 
     private fun readData(path: String): NSData {
-        return NSFileManager.defaultManager().contentsAtPath(path) ?: throw MissingResourceException(path)
+        return NSFileManager.defaultManager().contentsAtPath(path)
```
rock3r: Nit: expression syntax

components/resources/library/src/jsMain/kotlin/org/jetbrains/compose/resources/ResourceReader.js.kt
```diff
@@ -9,12 +9,13 @@ import org.w3c.files.Blob
...
-    return DefaultJsResourceReader
-}
+actual fun getDefaultResourceReader(): ResourceReader =
+    when {
+        isInTestEnvironment() -> TestResourceReader
+        else -> DefaultJsResourceReader
+    }
 
-private val DefaultJsResourceReader = object : ResourceReader {
+object DefaultJsResourceReader : ResourceReader {
```
rock3r: Shouldn't this be private?

components/resources/library/src/macosMain/kotlin/org/jetbrains/compose/resources/ResourceReader.macos.kt
```diff
@@ -2,10 +2,19 @@ package org.jetbrains.compose.resources
...
+import platform.Foundation.NSURL
+import platform.Foundation.closeFile
+import platform.Foundation.fileHandleForReadingAtPath
+import platform.Foundation.readDataOfLength
 import platform.posix.memcpy
 
-internal actual fun getPlatformResourceReader(): ResourceReader = object : ResourceReader {
+actual fun getDefaultResourceReader(): ResourceReader = DefaultMacOsResourceReader
+
+object DefaultMacOsResourceReader : ResourceReader {
```
rock3r: Shouldn't this be private?

components/resources/library/src/wasmJsMain/kotlin/org/jetbrains/compose/resources/ResourceReader.wasmJs.kt
```diff
@@ -23,12 +23,13 @@ private external fun jsExportInt8ArrayToWasm(src: Int8Array, size: Int, dstAddr:
...
-    return DefaultWasmResourceReader
-}
+actual fun getDefaultResourceReader(): ResourceReader =
+    when {
+        isInTestEnvironment() -> TestWasmResourceReader
+        else -> DefaultWasmResourceReader
+    }
 
-private val DefaultWasmResourceReader = object : ResourceReader {
+object DefaultWasmResourceReader : ResourceReader {
```
rock3r: Shouldn't this be private?

components/resources/library/src/wasmJsMain/kotlin/org/jetbrains/compose/resources/ResourceReader.wasmJs.kt
```diff
@@ -71,60 +72,59 @@ private val DefaultWasmResourceReader = object : ResourceReader {
...
+            ByteArray(size) { i -> (memBuffer + i).loadByte() }
         }
     }
 }
 
 // For blocking XmlHttpRequest the response can be only in text form, so we convert it to bytes manually
 private fun requestResponseAsByteArray(req: XMLHttpRequest): Int8Array =
-    js(""" {
+    js(
+        """ {
```
rock3r: Nit:
```suggestion
        """
        {
```

components/resources/library/src/wasmJsMain/kotlin/org/jetbrains/compose/resources/ResourceReader.wasmJs.kt
```diff
@@ -71,60 +72,59 @@ private val DefaultWasmResourceReader = object : ResourceReader {
...
         var text = req.responseText;
         var int8Arr = new Int8Array(text.length);
         for (var i = 0; i < text.length; i++) {
             int8Arr[i] = text.charCodeAt(i) & 0xFF;
         }
         return int8Arr;
-    }""")
+    }"""
```
rock3r: Nit:

```suggestion
    }
    """
```

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceReader.android.kt
```diff
@@ -8,7 +8,9 @@ import androidx.compose.runtime.ProvidableCompositionLocal
...
+actual fun getDefaultResourceReader(): ResourceReader = DefaultAndroidResourceReader
+
+object DefaultAndroidResourceReader : ResourceReader {
```
rock3r: Shouldn't this be private?

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestResourceReader.kt
```diff
@@ -6,15 +6,15 @@ internal class TestResourceReader : ResourceReader {
...
 
     override suspend fun readPart(path: String, offset: Long, size: Long): ByteArray {
         readPathsList.add("$path/$offset-$size")
-        return DefaultResourceReader.readPart(path, offset, size)
+        return getDefaultResourceReader().readPart(path, offset, size)
     }
 
     override fun getUri(path: String): String {
-        return DefaultResourceReader.getUri(path)
+        return getDefaultResourceReader().getUri(path)
```
rock3r: Nit: expression syntax

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceReader.desktop.kt
```diff
@@ -1,15 +1,22 @@
...
 import java.io.InputStream
 
-internal actual fun getPlatformResourceReader(): ResourceReader = object : ResourceReader {
-    override suspend fun read(path: String): ByteArray {
-        val resource = getResourceAsStream(path)
-        return resource.use { input -> input.readBytes() }
+internal actual fun getDefaultResourceReader(): ResourceReader =
+    JvmResourceReader.DEFAULT
+
+internal class JvmResourceReader(
```
rock3r: Shouldn't this be private?

components/resources/library/src/iosMain/kotlin/org/jetbrains/compose/resources/ResourceReader.ios.kt
```diff
@@ -26,7 +26,9 @@ import platform.Foundation.seekToFileOffset
 import platform.posix.memcpy
 
 @OptIn(BetaInteropApi::class)
-internal actual fun getPlatformResourceReader(): ResourceReader = object : ResourceReader {
+actual fun getDefaultResourceReader(): ResourceReader = DefaultIOsResourceReader
+
+object DefaultIOsResourceReader : ResourceReader {
```
rock3r: Shouldn't this be private?

components/resources/library/src/desktopTest/kotlin/org/jetbrains/compose/resources/CustomClassloaderTest.kt
lamba92:
```kotlin
interface ResourceReader {
  ...
  suspend fun read(path: String): ByteArray
  ..
}
```

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceReader.desktop.kt
lamba92: Well no, the whole point is to be able to instanciate your own `JvmResourceReader`

rock3r: It is internal though; "instanciate your own" as a library user is still not possible, is it?

components/resources/library/src/desktopTest/kotlin/org/jetbrains/compose/resources/CustomClassloaderTest.kt
rock3r: I see, I forgot `read` is suspend!

rock3r:
I can see why you'd want the JVM one to be public, but making others public is a bad idea for two reasons:
 1. You're adding public APIs you do not need and won't be able to easily change in the future (in a library you should keep a tight control of what API you publish, not just dump everything into the public API)
 2. Most (all?) other implementations are `object`s that can't be neither extended nor instantiated, and it is thus pointless to expose them to begin with

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceReader.desktop.kt
```diff
@@ -1,15 +1,22 @@
...
-        return resource.use { input -> input.readBytes() }
+actual fun getDefaultResourceReader(): ResourceReader =
+    JvmResourceReader.DEFAULT
+
+class JvmResourceReader(
+    private val classLoader: ClassLoader = JvmResourceReader::class.java.classLoader
+) : ResourceReader {
+
+    companion object {
+        val DEFAULT = JvmResourceReader()
```
igordmn:
```suggestion
        val Default = JvmResourceReader()
```
the Compose style guidelines are different from Kotlin guidelines

(cc @terrakok, we should take this into account for all future API, even if we already used another style)

igordmn: Consider making a top-level `val DefaultJvmResourceReader` to be consistent with the other targets

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceReader.kt
```diff
@@ -22,20 +22,19 @@ suspend fun readResourceBytes(path: String): ByteArray = DefaultResourceReader.r
...
+interface ResourceReader {
     suspend fun read(path: String): ByteArray
     suspend fun readPart(path: String, offset: Long, size: Long): ByteArray
     fun getUri(path: String): String
 }
 
-internal expect fun getPlatformResourceReader(): ResourceReader
-
-internal val DefaultResourceReader = getPlatformResourceReader()
+expect fun getDefaultResourceReader(): ResourceReader
```
igordmn: Consider making `val DefaultResourceReader` to be consistent with the platform implementations

igordmn (commented):
Besides the comments, LGTM. @terrakok should perform the main review

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceReader.kt
lamba92: I also changed other platform implementations to be `object`s. What do you mean exactly?

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceReader.desktop.kt
lamba92: I also changed other platform implementations to be `object`s. What do you mean exactly?

lamba92:
I added `@ExperimentalResourceApi` to the APIs I made public in this PR

components/resources/library/build.gradle.kts
```diff
@@ -25,25 +28,24 @@ kotlin {
     iosSimulatorArm64()
     js {
         browser {
-            testTask(Action {
+            testTask {
                 enabled = false
-            })
+            }
```
terrakok: please, revert back all unrelated to the PR changes

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceReader.kt
```diff
@@ -22,20 +22,22 @@ suspend fun readResourceBytes(path: String): ByteArray = DefaultResourceReader.r
...
     suspend fun read(path: String): ByteArray
     suspend fun readPart(path: String, offset: Long, size: Long): ByteArray
     fun getUri(path: String): String
 }
 
-internal expect fun getPlatformResourceReader(): ResourceReader
-
-internal val DefaultResourceReader = getPlatformResourceReader()
+@ExperimentalResourceApi
+expect fun getDefaultResourceReader(): ResourceReader
```
terrakok: Why do we need these changes?
I mean why it should be public?
Isn't it enough to have a public `LocalResourceReader` composition local?

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestResourceReader.kt
```diff
@@ -6,15 +6,14 @@ internal class TestResourceReader : ResourceReader {
...
         readPathsList.add("$path/$offset-$size")
-        return DefaultResourceReader.readPart(path, offset, size)
+        return getDefaultResourceReader().readPart(path, offset, size)
     }
 
-    override fun getUri(path: String): String {
-        return DefaultResourceReader.getUri(path)
-    }
+    override fun getUri(path: String): String =
+        getDefaultResourceReader().getUri(path)
```
terrakok: Please, don't change a code style

terrakok: if you don't agree with the current, it is OK. but should be discussed and applied by separate commit for the whole file

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceReader.kt
```diff
@@ -22,20 +22,22 @@ suspend fun readResourceBytes(path: String): ByteArray = DefaultResourceReader.r
...
+@ExperimentalResourceApi
+interface ResourceReader {
     suspend fun read(path: String): ByteArray
     suspend fun readPart(path: String, offset: Long, size: Long): ByteArray
     fun getUri(path: String): String
 }
 
-internal expect fun getPlatformResourceReader(): ResourceReader
-
-internal val DefaultResourceReader = getPlatformResourceReader()
```
terrakok, line 35: since `JvmResourceReader.Default` is object, why did you refactored this code?

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceReader.desktop.kt
```diff
@@ -30,18 +39,14 @@ internal actual fun getPlatformResourceReader(): ResourceReader = object : Resou
...
+            ?.toURI()
+            ?.toString()
+            ?: throw MissingResourceException(path)
 
-    private fun getResourceAsStream(path: String): InputStream {
-        val classLoader = getClassLoader()
-        return classLoader.getResourceAsStream(path) ?: throw MissingResourceException(path)
-    }
+    private fun getResourceAsStream(path: String): InputStream =
+        classLoader.getResourceAsStream(path) ?: throw MissingResourceException(path)
```
terrakok: avoid not required changes

components/resources/library/src/iosMain/kotlin/org/jetbrains/compose/resources/ResourceReader.ios.kt
terrakok, line 1: revert all unrelated changes here and in other files, please

terrakok (changes requested):

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceReader.kt
lamba92: This would allow the user to restore the default one if needed be. If you believe it should not be the case, I will revert it to private.

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestResourceReader.kt
lamba92: Sure, no problem, reverting

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceReader.desktop.kt
lamba92: Sure, no problem, reverting

components/resources/library/src/iosMain/kotlin/org/jetbrains/compose/resources/ResourceReader.ios.kt
lamba92: Sure, no problem, reverting

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceReader.kt
lamba92: `getPlatformResourceReader` was used only by `DefaultResourceReader`. But no I think it is better to revert this.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceReader.kt
terrakok: Users are able to get it from the Local provider, aren't they?

components/resources/library/src/wasmJsMain/kotlin/org/jetbrains/compose/resources/ResourceReader.wasmJs.kt
```diff
@@ -71,46 +73,43 @@ private val DefaultWasmResourceReader = object : ResourceReader {
     }
 }
 
-// It uses a synchronous XmlHttpRequest (blocking!!!)
```
terrakok, line 74: why did you delete the comment?

terrakok: it was lazy because it is needed for tests only

terrakok (changes requested):

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceReader.kt
terrakok: If so, then make it private again, please. And the PR will be very small and clear 😉

components/resources/library/src/wasmJsMain/kotlin/org/jetbrains/compose/resources/ResourceReader.wasmJs.kt
lamba92: This wasn't intended. Reverting

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceReader.android.kt
terrakok: why is it public?

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceReader.desktop.kt
```diff
@@ -1,10 +1,20 @@
...
+
+@ExperimentalResourceApi
+class JvmResourceReader(
+    private val classLoader: ClassLoader = JvmResourceReader::class.java.classLoader
+) : ResourceReader {
+
+    companion object {
+        val Default = JvmResourceReader()
+    }
+
```
terrakok: I guess it is better
```kotlin
@ExperimentalResourceApi
internal actual fun getPlatformResourceReader(): ResourceReader = JvmResourceReader()

@ExperimentalResourceApi 
class JvmResourceReader(
    private val classLoader: ClassLoader = JvmResourceReader::class.java.classLoader
) : ResourceReader {
  //no companion object
}
```

terrakok: oh. I see the difference now. then make
```
    companion object {
        internal val Default = JvmResourceReader()
    }
```

lamba92: In Kotlin all `object`s are initialized on their first symbol invokation. As such, a top level object is lazy by default. The `lazy` has been designed for inlining lazyness in a function or a class property. 

> Please, make only the JVM resource reader class public.

On it

lamba92:
❤️

PR #5339, opened 2025-06-19
KMT-599: Expose some of the preview parameters from androidx's Preview annotation

boiler23:
Exposed the following Preview annotation parameters:
- name
- group
- widthDp
- heightDp
- locale
- showBackground
- backgroundColor

## Testing
Tested the local artifact in IJ: new parameters are exposed and picked up. No additional changes on the IDE end are required.

But this should be tested by QA, of course.

### Features - Multiple Platforms
- Extended the `@Preview` (`org.jetbrains.compose.ui.tooling.preview`) annotation with the following parameters: name, group, widthDp, heightDp, locale, showBackground, backgroundColor. IDE (IJ or AS) will pick up these parameters in the same way it works for `androidx` Preview annotations.

### DIFF
--- a/components/ui-tooling-preview/library/src/commonMain/kotlin/org/jetbrains/compose/ui/tooling/preview/Preview.kt
+++ b/components/ui-tooling-preview/library/src/commonMain/kotlin/org/jetbrains/compose/ui/tooling/preview/Preview.kt
@@ -18,16 +18,32 @@ package org.jetbrains.compose.ui.tooling.preview
 
 /**
  * [Preview] can be applied to either of the following:
- * - @[Composable] methods with no parameters to show them in the IDE preview.
+ * - @[Composable] methods with no parameters to show them in the Android Studio preview.
  * - Annotation classes, that could then be used to annotate @[Composable] methods or other
- * annotation classes, which will then be considered as indirectly annotated with that [Preview].
+ *   annotation classes, which will then be considered as indirectly annotated with that [Preview].
+ *
+ * The annotation contains a number of parameters that allow to define the way the @[Composable]
+ * will be rendered within the preview.
+ *
+ * The passed parameters are only read by Studio when rendering the preview.
+ *
+ * @param name Display name of this preview allowing to identify it in the panel.
+ * @param group Group name for this @[Preview]. This allows grouping them in the UI and display only
+ *   one or more of them.
+ * @param widthDp Max width in DP the annotated @[Composable] will be rendered in. Use this to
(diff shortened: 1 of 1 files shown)


components/ui-tooling-preview/library/src/commonMain/kotlin/org/jetbrains/compose/ui/tooling/preview/Preview.kt
```diff
@@ -29,4 +29,12 @@ package org.jetbrains.compose.ui.tooling.preview
     AnnotationTarget.FUNCTION
 )
 @Repeatable
-annotation class Preview
+annotation class Preview(
```
igordmn, line 48: Let's copy Kdoc from https://github.com/androidx/androidx/blob/6a81aa377427fe02e51f305ecfa1550d1016731e/compose/ui/ui-tooling-preview/src/androidMain/kotlin/androidx/compose/ui/tooling/preview/Preview.android.kt#L62 (and verify that it is multiplatform friendly)

boiler23: Copied and left only the documentation for the fields we have.

igordmn (approved):

components/ui-tooling-preview/library/src/commonMain/kotlin/org/jetbrains/compose/ui/tooling/preview/Preview.kt
igordmn: Noticed only after merge:
```
- @[Composable] methods with no parameters to show them in the Android Studio preview.
```

this should be
```
- @[Composable] methods with no parameters to show them in the IDE preview.
```
right?

The meaning of "Let's copy Kdoc" was actually about parameters in my head :D

boiler23: oops, yes. Didn't notice he reference to AS too :)

igordmn: https://github.com/JetBrains/compose-multiplatform/pull/5346/files, please review

boiler23: let me fix it in a separate MR.

PR #5341, opened 2025-06-24
Introduce public API for clearing internal resource caches

pjBooms:
To support [hot reloading of resources](https://github.com/JetBrains/compose-hot-reload/issues/151) in Compose Hot Reload, it is required to clear internal resource caches when resources are changed.

Fixes [CMP-8439](https://youtrack.jetbrains.com/issue/CMP-8439)

## Testing
Unit tests are added.

## Release Notes
N/A


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
```diff
@@ -0,0 +1,22 @@
+package org.jetbrains.compose.resources
+
+private val caches = mutableListOf<AsyncCache<Any, Any>>()
+
+internal fun registerCache(cache: AsyncCache<Any, Any>) {
+    caches.add(cache)
+}
```
terrakok:
```kotlin
private val caches = mutableListOf<AsyncCache<*, *>>()

internal fun registerCache(cache: AsyncCache<*, *>) {
    caches.add(cache)
}
```

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/AsyncCache.kt
```diff
@@ -11,6 +11,11 @@ internal class AsyncCache<K, V> {
     private val mutex = Mutex()
     private val cache = mutableMapOf<K, Deferred<V>>()
 
+    init {
+        @Suppress("UNCHECKED_CAST")
+        registerCache(this as AsyncCache<Any, Any>)
```
terrakok: then you don't need the casting here

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
```diff
@@ -46,7 +49,7 @@ actual fun Font(resource: FontResource, weight: FontWeight, style: FontStyle): F
         val key = "$path:$weight:$style"
         fontCache.getOrLoad(key) {
             val fontBytes = resourceReader.read(path)
-            Font(path, fontBytes, weight, style)
+            Font("$path@${checksum(fontBytes)}", fontBytes, weight, style)
```
terrakok: let's use here a bytes size in common code, since it is already so for every platform except jvm. or even `size + last byte value` for correctness

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Clears any cached resources maintained internally by the system.
+ *
+ * This method is intended for use from Compose Hot Reload,
```
terrakok: I wouldn't mention the hot reload here.

components/resources/library/src/skikoTest/kotlin/org/jetbrains/compose/resources/FontCacheTest.kt
```diff
@@ -0,0 +1,125 @@
...
+
+@OptIn(ExperimentalTestApi::class)
+class FontCacheTest {
+
+    init {
+        dropResourceCaches()
+    }
+
+    @Test
+    fun testFontResourceCache() = runComposeUiTest {
```
terrakok: could you add a test to check font file switch with the same name?

pjBooms: It is in the commit named `testFontReplacement`

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
pjBooms: fixed. Thanx

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/AsyncCache.kt
pjBooms: done

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
pjBooms: Ok, some simple common function added

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
pjBooms: fixed

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
```diff
@@ -34,6 +34,9 @@ private val fontCache = AsyncCache<String, Font>()
 internal val Font.isEmptyPlaceholder: Boolean
     get() = this == defaultEmptyFont
 
+
+private fun checksum(data: ByteArray) = data.fold(0) { acc, byte -> acc * 31 + byte }
```
terrakok: I'm afraid it is too slow.
what do you think about:
```kotlin
private fun ByteArray.footprint() = "[$size:${lastOrNull()?.toInt()}]"
```

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * It can be useful to release memory or reset cached resources that
+ * may be changed or no longer be required.
+ *
+ * Note that frequent or unnecessary calls to this function may impact
+ * performance by removing resources that might otherwise benefit from being cached.
+ */
+fun dropResourceCaches() {
+   caches.forEach { it.clear() }
+}
```
terrakok: Let's create `ResourceCaches` object with this, and move the code into AsyncCache file. (the top level function is too wide visible)

```kotlin
internal object ResourceCaches {
  private val caches = mutableListOf<AsyncCache<*, *>>()

/**
 * Clears any cached resources maintained internally by the system.
 *
 * It can be useful to release memory or reset cached resources that
 * may be changed or no longer be required.
 *
 * Note that frequent or unnecessary calls to this function may impact
 * performance by removing resources that might otherwise benefit from being cached.
 */
  fun drop() {
     caches.forEach { it.clear() }
  }
}
```

And the `registerCache` function will not be necessary then.

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
pjBooms: ok, done

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
pjBooms: This way, you make the API internal because of the internal object. Moreover, it seems it is better then to move AsyncCache to ResourceCaches.kt, wdyt?

pjBooms: >And the registerCache function will not be necessary then.

I am not sure why it is not necessary

terrakok: yes, let's do the public object and move AsyncCache to ResourceCaches.kt 👍

components/resources/library/src/skikoMain/kotlin/org/jetbrains/compose/resources/FontResources.skiko.kt
```diff
@@ -46,7 +49,7 @@ actual fun Font(resource: FontResource, weight: FontWeight, style: FontStyle): F
         val key = "$path:$weight:$style"
         fontCache.getOrLoad(key) {
             val fontBytes = resourceReader.read(path)
-            Font(path, fontBytes, weight, style)
+            Font("$path${fontBytes.footprint()}", fontBytes, weight, style)
```
eymar, line 52: Why is it necessary to add the `footprint` here and below?

eymar: I guess I see: it's for a case when the file content changed, but its name/path is the same?

The "footprint" collision is still possible, right?

pjBooms: Because else the font will be cached by Skiko and won't be reloaded when we change it phisically

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
pjBooms: I have made a new commit, please look at, if it is what you meant

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
```diff
@@ -24,8 +28,26 @@ internal class AsyncCache<K, V> {
...
 
-    //@TestOnly
     fun clear() {
         cache.clear()
     }
-}
\ No newline at end of file
+}
+
+object ResourceCaches {
```
eymar, line 38: Do we want to mark it with `@ExperimentalResourceApi`?

Also about `fun drop()`.
I think "clear" or "reset" are more classic names in this context. Wdyt?

terrakok (approved):

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
eymar: Adding some KDoc on the object itself would be good.
We can share some info, that the caches are populated and managed internally and there is no direct access to the caches, but we provide an API to reset the cache if/when needed.

terrakok:
Update release notes: it is not a public feature. I'd say it is a fix to support the hot reload

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
pjBooms: I do not think we need to mark it with @ExperimentalResourceApi -- the usage from Compose Hot Reload will be reflective anyway. The idea is to provide stable API that has guarantee  to not change and not to break Compose Hot Reload in the future.

As to naming, I have no strong opinion here. Let's agree on `clear`? @terrakok?

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
```diff
@@ -24,8 +28,26 @@ internal class AsyncCache<K, V> {
...
+    /**
+     * Clears any cached resources maintained internally by the system.
+     *
+     * It can be useful to release memory or reset cached resources that
+     * may be changed or no longer be required.
+     *
+     * Note that frequent or unnecessary calls to this function may impact
+     * performance by removing resources that might otherwise benefit from being cached.
+     */
+    fun clear() {
```
sellmair: Is it expected that this API will ever be called on the UI thread?

eymar (approved):

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
terrakok: `@ExperimentalResourceApi` I think we need it. it is a new API and we don't plan to change it but anyway

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
terrakok: yes. all the resources processing works on UI thread only

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
pjBooms: Ok, I moved public API to desktop target only and marked as `@ExperimentalResourceApi`.

eymar: The `object ResourceCaches {` is public though. Did you want to make it internal too?

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestUtils.kt
```diff
@@ -66,4 +74,49 @@ internal fun parsePluralSamples(samples: String): List<Int> {
...
+}
+
+/**
+ * Executes a test block within a Compose UI testing environment while ensuring
+ * that any cached resources are cleared before the test begins.
+ */
+@OptIn(ExperimentalTestApi::class)
+fun runComposeResourceTest(block: ComposeUiTest.() -> Unit) = runComposeUiTest {
+    var areCachesCleared = false
+    setContent {
```
eymar: is this `setContent { LaunchedEffect() ... }` required?

Is it needed to run the coroutine?
____

Do such tests pass on web? The `waitUntil` is blocking the event loop. I'd expect that on web the tests will just timeout.

pjBooms: They pass

eymar: I see. The mutex is not locked, so it proceeds without suspending there. 

Could you please add a TODO to refactor this code after updating to CMP 1.9.0?
The refactoring would get rid of `setContent { LaunchedEffect() ... }` and the coroutine can be launched right in the test block - it will be suspend in 1.9.0.

eymar: maybe even now it makes sense to add:

```
MainScope().launch {
    ResourceCaches.asyncClear()
    areCachesCleared = true
}
```

instead of `setContent { LaunchedEffect() ... }`

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestUtils.kt
```diff
@@ -66,4 +74,49 @@ internal fun parsePluralSamples(samples: String): List<Int> {
...
+                ResourceCaches.asyncClear()
+                areCachesCleared = true
+            }
+        }
+    }
+
+    return {
+        clearCaches = true
+        waitUntil { areCachesCleared }
+    }
```
eymar: as we briefly discussed during the call, this indeed can be changed too.

we can have a function:

```
    // TODO: after updating to CMP 1.9.0, this function won't be needed - the tests will be able to call `ResourceCaches.asyncClear()` directly in the runComposeUiTest scope

@OptIn(ExperimentalTestApi::class)
fun ComposeUiTest.cleanResourceCache() {
    var areCachesCleared = false
    
    MainScope().launch(start = CoroutineStart.UNDISPATCHED) {
        ResourceCaches.asyncClear()
        areCachesCleared = true
    }
    
    waitUntil { areCachesCleared }
}
```

A test can call it whener it needs. And there is no need for an additional `setContentWithResourceCacheCleaning` - `setContent` will be enough.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
pjBooms: No. We want for `ResourceCaches.clear()` to be public, so `object ResourceCaches {` should be public

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestUtils.kt
pjBooms: Thank you! Done   :)

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestUtils.kt
pjBooms: done

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
```diff
@@ -24,8 +28,25 @@ internal class AsyncCache<K, V> {
...
+ *
+ * While the caches are managed automatically, there might be scenarios where manual cache clearing
+ * is desired. For such cases, we provide an API to reset the caches.
+*/
+object ResourceCaches {
+    private val caches = mutableListOf<AsyncCache<*, *>>()
+
+    internal fun registerCache(cache: AsyncCache<*, *>) = caches.add(cache)
+
+    internal suspend fun asyncClear() {
```
terrakok: let's rename it to `internal suspend fun clear()` and move the doc to the function

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestUtils.kt
```diff
@@ -66,4 +69,16 @@ internal fun parsePluralSamples(samples: String): List<Int> {
...
-}
\ No newline at end of file
+}
+
+/**
+ * Executes a test block within a Compose UI testing environment while ensuring
+ * that any cached resources are cleared before the test begins.
+ */
+@OptIn(ExperimentalTestApi::class)
+fun runComposeResourceTest(block: suspend ComposeUiTest.() -> Unit) = runComposeUiTest {
```
terrakok: I'd name it `clearResourceCachesAndRunUiTest`. this is much easier for understanding

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.desktop.kt
```diff
@@ -0,0 +1,14 @@
...
+import kotlinx.coroutines.runBlocking
+
+/**
+ * Clears any cached resources maintained internally by the system.
+ *
+ * It can be useful to release memory or reset cached resources that
+ * may be changed or no longer be required.
+ */
+@ExperimentalResourceApi
+fun ResourceCaches.clear() {
```
terrakok: name it `clearBlocking`

terrakok (changes requested):

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.desktop.kt
pjBooms: done

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/TestUtils.kt
pjBooms: done

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
pjBooms: done

terrakok (approved):

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.desktop.kt
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * Clears any cached resources maintained internally by the system.
+ *
+ * It can be useful to release memory or reset cached resources that
+ * may be changed or no longer be required.
+ */
+@ExperimentalResourceApi
+fun ResourceCaches.clearBlocking() {
+    runBlocking { clear() }
+}
```
sellmair, line 14: I want to raise a concern here: Clearing resources asynchronously might not work well and might lead to several issues. Wrapping it in 'runBlocking' might not work as the intended calling point from CHR would be on the main thread already.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
```diff
@@ -24,8 +28,25 @@ internal class AsyncCache<K, V> {
...
-        cache.clear()
+    suspend fun clear() {
+        mutex.withLock {
+            cache.clear()
+        }
+    }
+}
+
+object ResourceCaches {
+    private val caches = mutableListOf<AsyncCache<*, *>>()
```
sellmair, line 39: When reading AsyncCache and this ResourcesCache, then I wonder how it is defending against memory leaks?
It seems. now that every cache is strongly held statically?

components/resources/library/src/desktopMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.desktop.kt
pjBooms: Discussed privately. The API is async because of the web, we can't make it magically synchronous.  On the other hand, on non-web targets it is de-facto synchronous as resources are loaded synchronously using `runBlocking` too. So I don't expect problems here. Moreover I tested it in CHR environment before the merge.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceCaches.kt
pjBooms: There is a limited amount of caches created in the library (4 for now), so this global list does not bring much to memory consumption. As to AsyncCache itself, we are waiting for a motivating example to implement some evicting policy. In fact, it is limited too with a number of resources in the program though theoretically non-evicting cache can be a problem.

PR #5345, opened 2025-06-30
use maven local artifacts for teamcity test runs

shishkin-pavel:
instead of relying on maven publications for compose compiler test runs, rely on kotlin builds via "artifact" dependency
artifacts from those kotlin builds should be placed in maven local (done on teamcity config side)

## Release Notes
N/A

### DIFF
--- a/compose/integrations/composable-test-cases/build.gradle.kts
+++ b/compose/integrations/composable-test-cases/build.gradle.kts
@@ -20,8 +20,8 @@ allprojects {
 
         maven("https://packages.jetbrains.team/maven/p/kt/dev")
         maven("https://redirector.kotlinlang.org/maven/dev")
-        if (providers.gradleProperty("kotlin_version").orNull == "true") {
-            mavenLocal()
+        maven {
+            url = uri("${rootDir}/build/maven-project")
         }
     }
 
--- a/compose/integrations/composable-test-cases/build.gradle.kts
+++ b/compose/integrations/composable-test-cases/build.gradle.kts
@@ -1,6 +1,4 @@
 import org.jetbrains.kotlin.gradle.ExperimentalWasmDsl
-import org.jetbrains.kotlin.gradle.dsl.KotlinJsCompile
-import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
 
 //group "com.example"
 //version "1.0-SNAPSHOT"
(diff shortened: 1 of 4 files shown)


compose/integrations/composable-test-cases/build.gradle.kts
```diff
@@ -22,7 +20,9 @@ allprojects {
 
         maven("https://packages.jetbrains.team/maven/p/kt/dev")
         maven("https://redirector.kotlinlang.org/maven/dev")
-//        mavenLocal()
+        if (System.getenv("TEAMCITY_VERSION") != null) {
+            mavenLocal()
```
igordmn: What is the issue with current configuration?

Not sure why it helps, because the current configuration doesn't publish anything to Maven Local: https://teamcity.jetbrains.com/admin/editRunType.html?id=buildType:JetBrainsPublicProjects_Compose_IntegrationComposableTests&runnerId=RUNNER_1&cameFromUrl=%2Fadmin%2FeditBuildRunners.html%3Fid%3DbuildType%253AJetBrainsPublicProjects_Compose_IntegrationComposableTests%26init%3D1&cameFromTitle=

compose/integrations/composable-test-cases/build.gradle.kts
```diff
@@ -22,7 +20,9 @@ allprojects {
 
         maven("https://packages.jetbrains.team/maven/p/kt/dev")
         maven("https://redirector.kotlinlang.org/maven/dev")
-//        mavenLocal()
+        if (System.getenv("TEAMCITY_VERSION") != null) {
```
igordmn: Instead of `TEAMCITY_VERSION` it is better to use a custom Gradle property `project.property("use.maven.local") == "true"` and pass it from CI.

This way we don't bound the code to TeamCity (and can run in GitHub actions if needed)

compose/integrations/composable-test-cases/build.gradle.kts
shishkin-pavel: it uses artifact dependency (dependencies tab in configuration) and artifacts from that dependency should be placed in mavenlocal folder

igordmn: Using Maven Local is usually not advised, it can contain a lot of extra libraries that are downloaded by some other task and cached there, including broken ones.

Alternatively, we can copy the library into `build/maven-project`, and use it this way:
```
flatDir {       
    dirs("build/maven-project")
}
```

compose/integrations/composable-test-cases/build.gradle.kts
igordmn: It is still better to not the use existing parameters which purpose is different. `kotlin.version` is currently only for changing the Kotlin version. Adding a new repo is a more broad responsibility.

But if we add `dirs("build/maven-project")`, we can add it just without condition.

PR #5347, opened 2025-06-30
Enhanced `String.replaceWithArgs` to support non-positional format 

henryqwe:
Enhanced `String.replaceWithArgs` to support non-positional format specifiers (%s, %d) when only one argument is provided

[CMP-8385](https://youtrack.jetbrains.com/issue/CMP-8385/Resources-Support-for-d-and-s-in-StringResourcesUtils-for-single-argument-string-resource)

## Testing
Unit tests added

### DIFF
--- a/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/StringResourcesUtils.kt
+++ b/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/StringResourcesUtils.kt
@@ -19,7 +19,11 @@ internal fun String.replaceWithArgs(args: List<String>): String {
                 )
             }
         }
-        args[index]
+        args.getOrElse(index) {
+            throw IllegalArgumentException(
+                "Formatting failed: Placeholder '${match.value}' at position ${index + 1} is out of bounds. Only ${args.size} argument(s) provided for format string \"$this\""
+            )
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/StringResourcesUtils.kt
```diff
@@ -4,9 +4,23 @@ import org.jetbrains.compose.resources.plural.PluralCategory
...
+        val index = when {
+            placeholderNumber != null -> placeholderNumber - 1
+            args.size == 1 -> 0
+            else -> {
+                throw IllegalArgumentException(
+                    "Formatting failed: Non-positional placeholder '${match.value}' is ambiguous when multiple arguments are provided in \"$this\""
+                )
+            }
+        }
+        args[index]
```
harrytmthy: If a placeholder index exceeds the number of provided args (e.g. %3$s with only 2 args), this will throw IndexOutOfBoundsException. Would it be worth catching that and throwing a clearer error (similar to MissingFormatArgumentException) so devs can quickly see which placeholder is invalid?

kropp (changes requested):
Suggested changes and tests do not cover all cases supported by formatter.
See [Argument index](https://developer.android.com/reference/java/util/Formatter#argument-index_1) section for details.

terrakok:
Answered in the issue. Let's continue the discussion there

PR #5356, opened 2025-07-04
Fix macOs "could not find certificate"

TheDome0:
surround commandline arguments with quotes
see https://youtrack.jetbrains.com/issue/CMP-4272/createDistributable-signing-failing-on-macOs-with-Could-not-find-certificate#focus=Comments-27-12309226.0-0

Fixes https://youtrack.jetbrains.com/issue/CMP-4272/createDistributable-signing-failing-on-macOs-with-Could-not-find-certificate

### Fixes - Desktop
- Surround commandline arguments of macOs' "find-certificate" with double quotes

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/MacSigner.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/MacSigner.kt
@@ -75,8 +75,10 @@ internal class MacSignerImpl(
                     "find-certificate",
                     "-a",
                     "-c",
-                    settings.fullDeveloperID,
-                    settings.keychain?.absolutePath
+                    "\"${settings.fullDeveloperID}\"",
+                    settings.keychain?.absolutePath?.let { 
+                        "\"$it\""
+                    }
                 ),
                 processStdout = { signKeyValue = matchCertificates(it) }
             )


m-sasha:
Admittedly I don't understand anything about mac certificates or the signing process, but I'm not convinced that not having quotes is the problem here. I have two reasons for that:

1. It seems that the argument you're trying to quote, `settings.fullDeveloperID`, typically contains spaces. So if it was the problem, it would *never* work.
2. We have a (disabled; but I enabled it to try) test: `DesktopApplicationTest.testMacSign`. When run without quoting, the `security` command succeeds, and in the log I see:
```
Starting process 'command '/usr/bin/security''. Working directory: /private/var/folders/9f/j1xgg23s6d7_5rds74bwchnh0000gn/T/junit9606694438399542848 Command: /usr/bin/security find-certificate -a -c Developer ID Application: Compose Test /private/var/folders/9f/j1xgg23s6d7_5rds74bwchnh0000gn/T/junit9606694438399542848/compose.test.keychain
Successfully started process 'command '/usr/bin/security''
keychain: "/private/var/folders/9f/j1xgg23s6d7_5rds74bwchnh0000gn/T/junit9606694438399542848/compose.test.keychain"
version: 256
class: 0x80001000 
attributes:
    "alis"<blob>="Developer ID Application: Compose Test"
    "cenc"<uint32>=0x00000003 
    "ctyp"<uint32>=0x00000001 
    "hpky"<blob>=0x022C6E070EF5D9731FD5E14ECB6E55C55EDC3521  "\002,n\007\016\365\331s\037\325\341N\313nU\305^\3345!"
    "issu"<blob>=0x303E312F302D06035504030C26446576656C6F706572204944204170706C69636174696F6E3A20436F6D706F73652054657374310B3009060355040613025255  "0>1/0-\006\003U\004\003\014&Developer ID Application: Compose Test1\0130\011\006\003U\004\006\023\002RU"
    "labl"<blob>="Developer ID Application: Compose Test"
    "skid"<blob>=<NULL>
    "snbr"<blob>=0x01 
    "subj"<blob>=0x303E312F302D06035504030C26446576656C6F706572204944204170706C69636174696F6E3A20436F6D706F73652054657374310B3009060355040613025255  "0>1/0-\006\003U\004\003\014&Developer ID Application: Compose Test1\0130\011\006\003U\004\006\023\002RU"
```
So it appears that the `security` command parses the developer id from the commandline correctly even without quotes.

On the other hand, if I quote the `fullDeveloperID` argument, I get
```
Starting process 'command '/usr/bin/security''. Working directory: /private/var/folders/9f/j1xgg23s6d7_5rds74bwchnh0000gn/T/junit5874071336260233762 Command: /usr/bin/security find-certificate -a -c "Developer ID Application: Compose Test" "/private/var/folders/9f/j1xgg23s6d7_5rds74bwchnh0000gn/T/junit5874071336260233762/compose.test.keychain"
```
and the output of `security` appears to be empty, so `matchCertificates` fails with:
```
Could not find certificate for 'Compose Test' in keychain [/private/var/folders/9f/j1xgg23s6d7_5rds74bwchnh0000gn/T/junit5874071336260233762/compose.test.keychain]
```

m-sasha:
The problem here appears to be an unusual output of `security find-certificate` when the developer id contains non-ascii charactera.
Normally, the `alis` line looks like so:
```
"alis"<blob>="Developer ID Application: Compose Test"
```
but if the developer id has non-ascii characters, it looks like so:
```
"alis"<blob>=0x446576656C6F706572204944204170706C69636174696F6E3A20444F4D494E494B205354C39642592028423941464E483235325129  "Developer ID Application: DOMINIK ST\303\226BY (B9AFNH252Q)"
```

So our regex (`"alis"<blob>="([^"]+)"`) doesn't match it.
Looks like `jpackage` has (had?) a similar issue: https://bugs.openjdk.org/browse/JDK-8308042
Maybe we can adapt the code from their fix.

PR #5357, opened 2025-07-04
Add ResourceContentHash annotation to generated resource accessors

pjBooms:
Used by Compose Hot Reload to mark a related accessor as dirty if the hash is changed after reloading of new accessors.

Fixes [CMP-8513](https://youtrack.jetbrains.com/issue/CMP-8513)

## Release Notes
N/A


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceContentHash.kt
```diff
@@ -0,0 +1,5 @@
+package org.jetbrains.compose.resources
+
+@Suppress("unused")
+@Retention(AnnotationRetention.BINARY)
+annotation class ResourceContentHash(val hash: Int)
```
sellmair, line 5: I wonder if we might want to use Long instead?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
```diff
@@ -283,11 +286,17 @@ private fun getChunkFileSpec(
                 .endControlFlow()
                 .build()
 
-            val accessor = PropertySpec.builder(resName, type.getClassName(), resModifier)
+            val accessorBuilder = PropertySpec.builder(resName, type.getClassName(), resModifier)
                 .receiver(ClassName(packageName, resClassName, type.accessorName))
                 .delegate(initializer)
-                .build()
-            chunkFile.addProperty(accessor)
+            if (System.getProperty("compose.resources.generate.ResourceContentHash.annotation") == "true") {
```
sellmair, line 292: I would be interested to learn more about it, having to be enabled? Especially if the content hash is just 32-bit, the memory concern is a really lightweight argument? Am I missing something?

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceContentHash.kt
sellmair: UPD: I see that this uses String.hashCode therefore Int is used here, as well. 
Probably good enough to get started, but we should be clear about the fact that it _can_ happen that hashes collide.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
pjBooms: It is just unnecessary overhead that we would like to avoid for final distributions. We can strip the annotations in our tooling but there are chances that in a custom build they would be present. So it seems it is better just not generate the annotations but for Hot Reload.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceContentHash.kt
pjBooms: I wanted to keep the things simple. I don't know collisions probability of String.hashCode(), but I guess it is good enough for our case.  Though, no problem to make it 64-bit and more collision resistant

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
```diff
@@ -283,11 +286,17 @@ private fun getChunkFileSpec(
...
+                accessorBuilder.addAnnotation(
+                    AnnotationSpec.builder(resourceContentHashAnnotationClass)
+                    .useSiteTarget(AnnotationSpec.UseSiteTarget.DELEGATE)
+                    .addMember("%L", items.fold(0){acc, item -> ((acc * 31) + item.contentHash)}).build()
```
terrakok: fix indentions, please

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -110,7 +110,8 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask() {
         }
 
         val type = ResourceType.fromString(typeString) ?: error("Unknown resource type: '$typeString'.")
-        return listOf(ResourceItem(type, qualifiers, file.nameWithoutExtension.asUnderscoredIdentifier(), path))
+        return listOf(ResourceItem(type, qualifiers, file.nameWithoutExtension.asUnderscoredIdentifier(), path,
+            file.readBytes().contentHashCode()))
```
terrakok: let's put every parameter on a new line?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -141,7 +142,8 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask() {
         path: Path
     ): ResourceItem {
         val record = ValueResourceRecord.createFromString(recordString)
-        return ResourceItem(record.type, qualifiers, record.key.asUnderscoredIdentifier(), path, offset, size)
+        return ResourceItem(record.type, qualifiers, record.key.asUnderscoredIdentifier(), path,
+            record.content.hashCode(), offset, size)
```
terrakok: here as well

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -578,6 +578,17 @@ class ResourcesTest : GradlePluginTestBase() {
         }
     }
 
+    @Test
+    fun testGeneratedAccessorsAnnotatedWithResourceContentHash(): Unit = with(testProject("misc/commonResourcesAnnotatedWithResourceContentHash")) {
```
terrakok: Instead of the new test project, we discussed using an existed one (commonResources). You just need to add new expected directory: https://github.com/JetBrains/compose-multiplatform/tree/master/gradle-plugins/compose/src/test/test-projects/misc/commonResources

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
pjBooms: fixed

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
pjBooms: fixed

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
pjBooms: fixed

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
pjBooms: fixed

gradle-plugins/compose/src/test/test-projects/misc/commonResources/expected-with-hash/commonMainResourceAccessors/app/group/resources_test/generated/resources/Drawable0.commonMain.kt
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+@delegate:ResourceContentHash(1_620_038_668)
+internal val Res.drawable.camelCaseName: DrawableResource by lazy {
+      DrawableResource("drawable:camelCaseName", setOf(
+        ResourceItem(setOf(), "${MD}drawable/camelCaseName.xml", -1, -1),
+      ))
+    }
+
+@delegate:ResourceContentHash(1_620_038_668)
```
terrakok, line 32: Is it expected that hashes are equal?

pjBooms: As underlying files seem to be the same by content then yes, the hashes will be equal.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -110,7 +112,25 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask() {
...
+            )
+        )
+    }
+
+    private fun File.contentHashCode(): Int {
+        return if (extension.lowercase() == "xml") {
+            // Text files have different line endings on Windows/*nix,
+            // so let's ignore them in the hashcode calculation
+            Files.readAllLines(toPath()).joinToString().hashCode()
+        } else {
```
terrakok: sorry, I don't understand it. how does it affect our usecase?

pjBooms: Tests on Windows fail without this because when xml files are checked out they are not equal by content with the same files on Linux/Mac (because of different line endings).

terrakok: the same should happen for any text files, right? I think a correct way is to fix tests instead of hacks here

pjBooms: Ok, done

pjBooms:
> Could you check how the build time is being changed?

No differences noted: 1.9.0-beta03 and this version both build around 37s on my machine after `./gradlew clean`

PR #5358, opened 2025-07-07
Fix signing when the developer ID contains non-ASCII characters.

m-sasha:
When the developer id contains non-ASCII characters, the output of `security find-certificates` is different, and we need different code to parse it.

## Testing
Tested manually and added a unit test.
Note that the test is disabled, and it actually fails at the end, but it fails after the code in question, and the previously existing `macSign` test fails at the same place.

### Fixes - Gradle Plugin
- Fix codesigning on macOS when the developer id contains non-ASCII characters. Note that this requires JDK 21 or later.


gradle-plugins/compose/src/test/test-projects/application/macSign/settings.gradle
```diff
@@ -3,6 +3,7 @@ pluginManagement {
         id 'org.jetbrains.kotlin.jvm' version 'KOTLIN_VERSION_PLACEHOLDER'
         id 'org.jetbrains.kotlin.plugin.compose' version 'KOTLIN_VERSION_PLACEHOLDER'
         id 'org.jetbrains.compose' version 'COMPOSE_GRADLE_PLUGIN_VERSION_PLACEHOLDER'
+        id 'org.gradle.toolchains.foojay-resolver-convention' version '1.0.0'
```
terrakok, line 6: Why do we need this?

terrakok (commented):
LGTM

gradle-plugins/compose/src/test/test-projects/application/macSign/settings.gradle
m-sasha: It's a plugin that automatically downloads and "installs" JDKs as needed. We need it because the tests now include testing with a JDK not installed by default (JDK 21, in `testMacSignWithNonAsciiDeveloperId`)

PR #5359, opened 2025-07-08
Use new exception handling proposal for k/wasm targets

eymar:
Just adding a compilation flag to use the new proposal.

## Testing
N/A

## Release Notes
N/A


eymar:
cc: @danleh merged it

PR #5360, opened 2025-07-09
Hardcode Material3 to stable

igordmn:
Fixes https://youtrack.jetbrains.com/issue/CMP-8527/Hardcode-material3-alias-to-the-stable-version

### Migration Notes - Multiple Platforms
- Material3 versioning is decoupled for the Compose Multiplatform 1.9.* release due the upstream Jetpack Compose Material3 1.4 has not been released as stable yet
- `compose.material3` now points to the latest stable Material3 version, 1.8.2. If the latest Material3 features are needed, please include it this way:
```
implementation("org.jetbrains.compose.material3:material3:1.9.0-beta06")
```

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
@@ -69,8 +69,8 @@ abstract class ComposePlugin : Plugin<Project> {
         val animationGraphics get() = composeDependency("org.jetbrains.compose.animation:animation-graphics")
         val foundation get() = composeDependency("org.jetbrains.compose.foundation:foundation")
         val material get() = composeDependency("org.jetbrains.compose.material:material")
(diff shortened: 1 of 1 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -69,8 +69,8 @@ abstract class ComposePlugin : Plugin<Project> {
         val animationGraphics get() = composeDependency("org.jetbrains.compose.animation:animation-graphics")
```
igordmn, line 69: Should be merged either to `master` or `release/1.9` depending on the discussion in https://youtrack.jetbrains.com/issue/CMP-8527/Hardcode-material3-alias-to-the-stable-version#focus=Comments-27-12340180.0-0

igordmn: Decided to merge it right now

PR #5364, opened 2025-07-14
Fix Maven Project test template (add -Dcompose.material3.version=)

igordmn:
After decoupling Material3 from Compose, we need to pass their versions separately

[CI fix](https://jetbrains.team/p/ui/repositories/compose-teamcity-config/revision/df83b9cf9acff67a45f799e0de7559bc04d7c516)

## Testing
```
./check.sh
./check.sh -Dcompose.version=1.9.0-alpha02
./check.sh -Dcompose.version=1.9.0-alpha01 -Dcompose.material3.version=1.9.0-alpha03
```

## Release Notes
N/A

### DIFF
--- a/ci/templates/maven-test-project/pom.xml
+++ b/ci/templates/maven-test-project/pom.xml
@@ -14,6 +14,7 @@
         <kotlin.compiler.jvmTarget>11</kotlin.compiler.jvmTarget>
         <kotlin.version>2.1.20</kotlin.version>
         <compose.version>1.8.2</compose.version>
+        <compose.material3.version>1.8.2</compose.material3.version>
     </properties>
 
     <repositories>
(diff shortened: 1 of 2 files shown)


ci/templates/maven-test-project/pom.xml
```diff
@@ -99,10 +100,15 @@
             <version>${kotlin.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.jetbrains.compose.material3</groupId>
-            <artifactId>material3-desktop</artifactId>
+            <groupId>org.jetbrains.compose.foundation</groupId>
```
igordmn, line 103: After including foundation, we have another issue with 1.9.0-alpha03, which we missed because material3 included only 1.8.2. I am fixing in a separate PR.
```
[ERROR] Failed to execute goal on project maven-test-project: Could not resolve dependencies for project org.example:maven-test-project:jar:1.0-SNAPSHOT: Failure to find androidx.core:core-viewtree:jar:1.0.0 in https://repo1.maven.org/maven2/ was cached in the local repository, resolution will not be reattempted until the update interval of mavenCentral has elapsed or updates are forced -> [Help 1]
```

MatkovIvan (approved):

PR #5375, opened 2025-07-24
Introduce wasmJsBrowserDistribution task

Schahen:
## Testing
Clone https://github.com/Schahen/ComposeWebApp and run 
```
 ./gradlew  :composeApp: composeCompatibilityBrowserDistribution
```

### Features - Web
- Introduce composeCompatibilityBrowserDistribution  task. This task combines two prod distributions - for js and for wasm in such way so that if modern require features are not supported by the consumer browser, application switch to js mode.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
@@ -21,9 +21,7 @@ import org.jetbrains.compose.internal.utils.file
 import org.jetbrains.compose.internal.utils.registerTask
 import org.jetbrains.compose.web.WebExtension
 import org.jetbrains.compose.web.tasks.UnpackSkikoWasmRuntimeTask
-import org.jetbrains.kotlin.gradle.targets.js.KotlinWasmTargetType
 import org.jetbrains.kotlin.gradle.targets.js.ir.KotlinJsIrTarget
-import org.jetbrains.kotlin.gradle.tasks.IncrementalSyncTask
 
 internal fun Project.configureWeb(
     composeExt: ComposeExtension,
@@ -84,6 +82,98 @@ internal fun configureWebApplication(
 
         skikoRuntimeFiles = skikoJsWasmRuntimeConfiguration
         outputDir.set(unpackedRuntimeDir)
+    }
+
+    val webProductionDist = project.layout.buildDirectory.dir("dist/web/productionExecutable")
+    val jsAppRenamed = "__jsApp.js"
+    val wasmAppRenamed = "__wasmApp.js"
+
+    val copyJsDistSpec = project.copySpec { spec ->
+        val jsTarget = targets.firstOrNull { it.name == "js" }?.outputModuleName
+        spec.apply {
+            from( project.tasks.named("jsBrowserDistribution")) {
+                rename { name ->
+                    val moduleName = jsTarget?.get()
+                    when (name) {
+                        "${moduleName}.js" -> jsAppRenamed
+                        "${moduleName}.js.map" -> "${jsAppRenamed}.map"
+                        else -> name
+                    }
+                }
+            }
+        }
+    }
+
+    val copyWasmDistSpec = project.copySpec { spec ->
+        val wasmTarget = targets.firstOrNull { it.name == "wasmJs" }?.outputModuleName
+        spec.apply {
+            from( project.tasks.named("wasmJsBrowserDistribution")) {
+                rename { name ->
+                    val moduleName = wasmTarget?.get()
+                    when (name) {
+                        "${moduleName}.js" -> wasmAppRenamed
+                        "${moduleName}.js.map" -> "${wasmAppRenamed}.map"
+                        else -> name
+                    }
+                }
+            }
+        }
+    }
+
+
+    val composeWebCompatibilityDistTask = project.registerTask<Copy>("composeWebCompatibilityDist") {
+        val jsTarget = targets.firstOrNull { it.name == "js" }?.outputModuleName
+        val wasmTarget = targets.firstOrNull { it.name == "wasmJs" }?.outputModuleName
+
+        duplicatesStrategy = DuplicatesStrategy.WARN
+        onlyIf {
+            jsTarget != null && wasmTarget != null
+        }
+
+        into(webProductionDist)
+        with(copyJsDistSpec, copyWasmDistSpec)
+
+        doLast {
+            val fallbackResolverCode = """
+                 const simpleWasmModule = new Uint8Array([
+                0,  97, 115, 109,   1,   0,   0,  0,   1,   8,   2,  95,
+                1, 120,   0,  96,   0,   0,   3,  3,   2,   1,   1,  10,
+               14,   2,   6,   0,   6,  64,  25, 11,  11,   5,   0, 208,
+              112,  26,  11,   0,  45,   4, 110, 97, 109, 101,   1,  15,
+                2,   0,   5, 102, 117, 110,  99, 48,   1,   5, 102, 117,
+              110,  99,  49,   4,   8,   1,   0,  5, 116, 121, 112, 101,
+               48,  10,  11,   1,   0,   1,   0,  6, 102, 105, 101, 108,
+              100,  48
+                ]);
+
+            const hasSupportOfAllRequiredWasmFeatures = () =>
+                typeof WebAssembly !== "undefined" &&
+                typeof WebAssembly?.validate === "function" &&
+                WebAssembly.validate(simpleWasmModule);
+
+            const createScript = (src) => {
+                const script = document.createElement("script");
+                script.src = src;
+                script.type = "application/javascript";
+                return script;
+            }
+            
+            if (hasSupportOfAllRequiredWasmFeatures()) {
+                document.body.appendChild(createScript('$wasmAppRenamed'));
+            } else {
+                document.body.appendChild(createScript('$jsAppRenamed'));
+            }
+            
+            """.trimIndent()
+
+            webProductionDist.file("${jsTarget!!.get()}.js").get().asFile.writeText(
+                fallbackResolverCode
+            )
+
+            webProductionDist.file("${wasmTarget!!.get()}.js").get().asFile.writeText(
+                fallbackResolverCode
+            )
+        }
     }
 
     val processSkikoRuntimeForKWasm = project.registerTask<Copy>("processSkikoRuntimeForKWasm") {
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
@@ -58,70 +58,13 @@ internal fun Project.configureWeb(
     }
 }
 
-internal fun configureWebApplication(
-    targets: Collection<KotlinJsIrTarget>,
-    project: Project,
-    shouldRunUnpackSkiko: Provider<Boolean>
-) {
-    val skikoJsWasmRuntimeConfiguration = project.configurations.create("COMPOSE_SKIKO_JS_WASM_RUNTIME")
-    val skikoJsWasmRuntimeDependency = skikoVersionProvider(project).map { skikoVersion ->
-        project.dependencies.create("org.jetbrains.skiko:skiko-js-wasm-runtime:$skikoVersion")
-    }
-    skikoJsWasmRuntimeConfiguration.defaultDependencies {
-        it.addLater(skikoJsWasmRuntimeDependency)
-    }
-
-    val unpackedRuntimeDir = project.layout.buildDirectory.dir("compose/skiko-for-web-runtime")
-    val processedRuntimeDir = project.layout.buildDirectory.dir("compose/skiko-runtime-processed-wasmjs")
-    val taskName = "unpackSkikoWasmRuntime"
-
-    val unpackRuntime = project.registerTask<UnpackSkikoWasmRuntimeTask>(taskName) {
-        onlyIf {
-            shouldRunUnpackSkiko.get()
-        }
-
-        skikoRuntimeFiles = skikoJsWasmRuntimeConfiguration
-        outputDir.set(unpackedRuntimeDir)
-    }
-
+private fun Project.introduceComposeWebCompatibilityTask(targets: Collection<KotlinJsIrTarget>) {
     val webProductionDist = project.layout.buildDirectory.dir("dist/web/productionExecutable")
+
     val jsAppRenamed = "__jsApp.js"
     val wasmAppRenamed = "__wasmApp.js"
 
-    val copyJsDistSpec = project.copySpec { spec ->
-        val jsTarget = targets.firstOrNull { it.name == "js" }?.outputModuleName
-        spec.apply {
-            from( project.tasks.named("jsBrowserDistribution")) {
-                rename { name ->
-                    val moduleName = jsTarget?.get()
-                    when (name) {
-                        "${moduleName}.js" -> jsAppRenamed
-                        "${moduleName}.js.map" -> "${jsAppRenamed}.map"
-                        else -> name
-                    }
-                }
-            }
-        }
-    }
-
-    val copyWasmDistSpec = project.copySpec { spec ->
-        val wasmTarget = targets.firstOrNull { it.name == "wasmJs" }?.outputModuleName
-        spec.apply {
-            from( project.tasks.named("wasmJsBrowserDistribution")) {
-                rename { name ->
-                    val moduleName = wasmTarget?.get()
-                    when (name) {
-                        "${moduleName}.js" -> wasmAppRenamed
-                        "${moduleName}.js.map" -> "${wasmAppRenamed}.map"
-                        else -> name
-                    }
-                }
-            }
-        }
-    }
-
-
-    val composeWebCompatibilityDistTask = project.registerTask<Copy>("composeWebCompatibilityDist") {
+    project.registerTask<Copy>("composeWebCompatibilityDist") {
         val jsTarget = targets.firstOrNull { it.name == "js" }?.outputModuleName
         val wasmTarget = targets.firstOrNull { it.name == "wasmJs" }?.outputModuleName
 
@@ -131,7 +74,35 @@ internal fun configureWebApplication(
         }
 
         into(webProductionDist)
-        with(copyJsDistSpec, copyWasmDistSpec)
+        with(project.copySpec { spec ->
+            val jsTarget = targets.firstOrNull { it.name == "js" }?.outputModuleName
+            spec.apply {
+                from( project.tasks.named("jsBrowserDistribution")) {
+                    rename { name ->
+                        val moduleName = jsTarget?.get()
+                        when (name) {
+                            "${moduleName}.js" -> jsAppRenamed
+                            "${moduleName}.js.map" -> "${jsAppRenamed}.map"
+                            else -> name
+                        }
+                    }
+                }
+            }
+        }, project.copySpec { spec ->
+            val wasmTarget = targets.firstOrNull { it.name == "wasmJs" }?.outputModuleName
+            spec.apply {
+                from( project.tasks.named("wasmJsBrowserDistribution")) {
+                    rename { name ->
+                        val moduleName = wasmTarget?.get()
+                        when (name) {
+                            "${moduleName}.js" -> wasmAppRenamed
+                            "${moduleName}.js.map" -> "${wasmAppRenamed}.map"
+                            else -> name
+                        }
+                    }
+                }
+            }
+        })
 
         doLast {
             val fallbackResolverCode = """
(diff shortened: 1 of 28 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -86,6 +84,98 @@ internal fun configureWebApplication(
...
+                        "${moduleName}.js.map" -> "${wasmAppRenamed}.map"
+                        else -> name
+                    }
+                }
+            }
+        }
+    }
+
+
+    val composeWebCompatibilityDistTask = project.registerTask<Copy>("composeWebCompatibilityDist") {
```
eymar: this task is registered unconditionally. 

This task only makes sense if there are both k/js and k/wasm targets.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -86,6 +84,98 @@ internal fun configureWebApplication(
         outputDir.set(unpackedRuntimeDir)
     }
 
+    val webProductionDist = project.layout.buildDirectory.dir("dist/web/productionExecutable")
```
eymar: Could you please extract this new logic into a separate function?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
eymar: I see there is 

```
onlyIf {
            jsTarget != null && wasmTarget != null
        }
```

But for a user of this gradle task it's not clear why it's available but doesn't run.

eymar: Maybe it's okay to have this task unconditionally, if at least we print some logs explaining why it didn't run.

Schahen: @eymar yep, that's the idea - this way it's more obvious that this task exists but not running

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
Schahen: @eymar pushed

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -60,6 +58,99 @@ internal fun Project.configureWeb(
...
+
+    val jsAppRenamed = "__jsApp.js"
+    val wasmAppRenamed = "__wasmApp.js"
+
+    val jsDistTask = targets.firstOrNull { it.name == "js" } ?: return
+    val wasmDistTask = targets.firstOrNull { it.name == "wasmJs" } ?: return
+
+    registerTask<Copy>("composeWebCompatibilityDist") {
+        val jsTargetModuleName = jsDistTask.outputModuleName
+        val wasmTargetModuleName = wasmDistTask.outputModuleName
```
eymar: Do both `outputModuleName` have some default value when a user doesn't explicitly specify it in the DSL?

When would this be `false`?

```
onlyIf {
            jsTargetModuleName != null && wasmTargetModuleName != null
        }
```

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -60,6 +58,99 @@ internal fun Project.configureWeb(
     }
 }
 
+private fun Project.introduceComposeWebCompatibilityTask(targets: Collection<KotlinJsIrTarget>) {
```
eymar: There are some gradle plugin tests in the project.

Would you like to add some tests?
For example:
- a task is created when the condition match and it's not created otherwise
- Run the task and check the output directory exists. And the output directory contains the must-have files (index.html, .js file, wasm file, etc)

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
```diff
@@ -86,7 +87,7 @@ class GradlePluginTest : GradlePluginTestBase() {
             }
         }
 
-    // Note: we can't test non-jvm targets with Kotlin older than 2.1.0, because of klib abi version bump in 2.1.0
+    // Note: we can"t test non-jvm targets with Kotlin older than 2.1.0, because of klib abi version bump in 2.1.0
```
eymar: accidental change?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
```diff
@@ -114,6 +115,32 @@ class GradlePluginTest : GradlePluginTestBase() {
...
+            check.taskSuccessful(":composeApp:composeWebCompatibilityDist")
+
+            file("./composeApp/build/dist/web/productionExecutable").apply {
+                checkExists()
+                assertTrue(isDirectory)
+                val distributionFiles = listFiles()!!.map { it.name }.toSet()
+
+                assertTrue(distributionFiles.any { it.endsWith(".wasm") })
+
+                assertContentEquals(distributionFiles.filter { !it.endsWith(".wasm") }, setOf(
```
eymar: `.wasm` is expected too. Can it be included in the check? 
I remember its name is not persistent. Maybe checking that .wasm files are there (1 for skiko and 1 for the compose app).

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
terrakok: >targets

in the gradle we must avoid to use final collections.
proper way is:
 1) register a new task with empty inputs
 2) subscribe on the live targets collection and add them in the callback

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
terrakok: this should be lazy providers to be read in the execution

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -60,6 +58,99 @@ internal fun Project.configureWeb(
...
+    val webProductionDist = layout.buildDirectory.dir("dist/web/productionExecutable")
+
+    val jsAppRenamed = "__jsApp.js"
+    val wasmAppRenamed = "__wasmApp.js"
+
+    val jsDistTask = targets.firstOrNull { it.name == "js" } ?: return
+    val wasmDistTask = targets.firstOrNull { it.name == "wasmJs" } ?: return
+
+    registerTask<Copy>("composeWebCompatibilityDist") {
```
terrakok: let's introduce new task extended from Copy and extract it to a new file

terrakok (changes requested):

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
Schahen: @eymar we are checking that wasm is present but unfortunately we can not know exact names - well, we can everything but it's tricky

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,152 @@
...
+            fallbackResolverCode
+        )
+
+        outputDir.file("${wasmOutputName.get()}.js").get().asFile.writeText(
+            fallbackResolverCode
+        )
+    }
+}
+
+private fun Project.registerWebCompatibilityTask(mppPlugin: KotlinMultiplatformExtension)  =  registerTask<WebCompatibilityTask>("composeWebCompatibilityDist") {
```
terrakok: we have `jsBrowserDistribution` and `wasmJsBrowserDistribution`
let's call it `compatibilityBrowserDistribution`

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,152 @@
...
+    onlyIf {
+        jsOutputName.orNull != null && wasmOutputName.orNull != null
+    }
+
+    outputDir.set(webProductionDist)
+    jsDistFiles = tasks.named("jsBrowserDistribution").get().outputs.files
+    wasmDistFiles = tasks.named("wasmJsBrowserDistribution").get().outputs.files
+}
+
+internal fun Project.configureWebCompatibility() {
```
terrakok, line 169: let's add a description what's going here

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
terrakok: 🧐

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,152 @@
...
+        )
+
+        outputDir.file("${wasmOutputName.get()}.js").get().asFile.writeText(
+            fallbackResolverCode
+        )
+    }
+}
+
+private fun Project.registerWebCompatibilityTask(mppPlugin: KotlinMultiplatformExtension)  =  registerTask<WebCompatibilityTask>("composeWebCompatibilityDist") {
+    val webProductionDist = layout.buildDirectory.dir("dist/web/productionExecutable")
```
eymar: `dist/web/productionExecutable` - potentially it can conflict with some Kotlin tasks.

Can we have a compose specific path? For example `dist/composeWebCompatibility/productionExecutable` ?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
eymar: Can we have a group and description for the task?

for example:
```
this.group = "compose"
this.description = "This task combines both js and wasm distributions into one"
```

Schahen: pushed a fix

gradle-plugins/compose/src/test/test-projects/application/webJsWasm/composeApp/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+import org.jetbrains.kotlin.gradle.targets.js.webpack.KotlinWebpackConfig
+
+plugins {
+    id("org.jetbrains.kotlin.multiplatform")
+    id("org.jetbrains.compose")
+    id("org.jetbrains.kotlin.plugin.compose")
+}
+
+kotlin {
+    js(IR) {
```
terrakok, line 11: As I know, we already don't need `IR` here

gradle-plugins/compose/src/test/test-projects/application/webJsWasm/composeApp/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+        outputModuleName.set("composeApp")
```
terrakok, line 12: If we want to test a custom name here, then you need a different name, because "composeApp" is a name of the module already

gradle-plugins/compose/src/test/test-projects/application/webJsWasm/composeApp/build.gradle.kts
```diff
@@ -0,0 +1,56 @@
...
+            commonWebpackConfig {
+                outputFileName = "composeApp.js"
+            }
+        }
+        binaries.executable()
+    }
+
+    @OptIn(ExperimentalWasmDsl::class)
+    wasmJs {
+        outputModuleName.set("composeApp")
```
terrakok, line 23: what will happen if there is a different name?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
```diff
@@ -114,6 +115,32 @@ class GradlePluginTest : GradlePluginTestBase() {
...
+            check.taskSuccessful(":composeApp:wasmJsBrowserDistribution")
+
+            file("./composeApp/build/dist/web/productionExecutable").apply {
+                checkExists()
+                assertTrue(isDirectory)
+                val distributionFiles = listFiles()!!.map { it.name }.toList().sorted()
+
+                assertTrue(distributionFiles.any { it.endsWith(".wasm") })
+
+                assertContentEquals(distributionFiles.filter { !it.endsWith(".wasm") }, listOf(
```
terrakok: can we check a full list of files?

terrakok (changes requested):
let's add a test with a custom target name `wasmJs("myStrangeTarget")`

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
Schahen: typo  - pushed correction

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
eymar: @terrakok, we wanted to make it obvious that the task is for Compose.  And avoid potential name collisions. 
So I think that the word "compose" must be in the name.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,170 @@
...
+        val wasmAppRenamed = "__wasmApp.js"
+
+        fileOperations.clearDirs(outputDir)
+
+        fileOperations.copy { copySpec ->
+            copySpec.duplicatesStrategy = DuplicatesStrategy.WARN
+
+            copySpec.from(jsDistFiles) {
+                copySpec.rename { name ->
+                    val moduleName = jsOutputName.get()
```
harrytmthy: NIT: Since `jsOutputName` and `wasmOutputName` are annotated with `@Optional`, calling `.get()` directly could throw an `IllegalStateException` if they're unset.

Would it make sense to either:
- Use `orNull ?: return@rename name` here, or
- Remove `@Optional` if these properties are guaranteed to be set by the time the task runs?

This keeps the annotation and implementation consistent 🙂

eymar: the task is not expected to run when either of them is not set. There is `onlyIf { ... }` checking it. 

So yes, they seem to be not Optional

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
terrakok: then `composeCompatibilityBrowserDistribution`

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
Schahen: @terrakok you mean recursively?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
harrytmthy: Exactly, that was my point. Since the properties aren't actually optional, removing `@Optional` would make it clearer 👍

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,170 @@
...
+                        "${moduleName}.js.map" -> "${jsAppRenamed}.map"
+                        else -> name
+                    }
+                }
+            }
+
+            copySpec.into(outputDir)
+        }
+
+        fileOperations.copy { copySpec ->
```
terrakok: let's combine two `fileOperations.copy` into a single one

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,170 @@
...
+    abstract val jsOutputName: Property<String>
+
+    @get:Input
+    @get:Optional
+    abstract val wasmOutputName: Property<String>
+
+    @TaskAction
+    fun run() {
+        val jsAppRenamed = "__jsApp.js"
+        val wasmAppRenamed = "__wasmApp.js"
```
terrakok:
```kotlin
val prefix = "origin"
val jsAppFileName = jsOutputName.get() + ".js"
val jsAppRenamed = joinLowerCamelCase(prefix, jsAppFileName)
val wasmAppFileName = wasmOutputName.get() + ".js"
val wasmAppRenamed = joinLowerCamelCase(prefix, wasmAppFileName)
```

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,170 @@
...
+            
+            """.trimIndent()
+
+        outputDir.file("${jsOutputName.get()}.js").get().asFile.writeText(
+            fallbackResolverCode
+        )
+
+        outputDir.file("${wasmOutputName.get()}.js").get().asFile.writeText(
+            fallbackResolverCode
+        )
```
terrakok:
```kotlin
val outputDir = outputDir.get().asFile
File(outputDir, jsAppFileName).writeText(fallbackResolverCode)
File(outputDir, wasmAppFileName).writeText(fallbackResolverCode)
```

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
terrakok: no.i mean with no the wasm filter.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    onlyIf {
+        val hasBothDistributions = !jsDistFiles.isEmpty && !wasmDistFiles.isEmpty
+        val hasBothOutputs = jsOutputName.orNull != null && wasmOutputName.orNull != null
+
+        if (!hasBothDistributions) {
+            logger.lifecycle("Task ${this.name} skipped: no js and wasm distributions found, both are required for compatibility")
+        } else if (!hasBothOutputs) {
+            logger.lifecycle("Task ${this.name} skipped: no js and wasm output names specified")
```
terrakok: add tests for this logic too, please

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,170 @@
...
+        val jsAppRenamed = "__jsApp.js"
+        val wasmAppRenamed = "__wasmApp.js"
+
+        fileOperations.clearDirs(outputDir)
+
+        fileOperations.copy { copySpec ->
+            copySpec.duplicatesStrategy = DuplicatesStrategy.WARN
+
+            copySpec.from(jsDistFiles) {
+                copySpec.rename { name ->
```
terrakok: you have to use `it` here. you nedd to configure a local copyspec instead of global

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,170 @@
...
+            }
+
+            copySpec.into(outputDir)
+        }
+
+        fileOperations.copy { copySpec ->
+            copySpec.duplicatesStrategy = DuplicatesStrategy.WARN
+
+            copySpec.from(wasmDistFiles) {
+                copySpec.rename { name ->
```
terrakok: you have to use `it` here. you nedd to configure a local copyspec instead of global

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
```diff
@@ -0,0 +1,170 @@
...
+    val webProductionDist = layout.buildDirectory.dir("dist/composeWebCompatibility/productionExecutable")
+    outputDir.set(webProductionDist)
+
+    mppPlugin.targets.matching { it is KotlinJsIrTarget }.all { target ->
+        val outputName = when {
+            (target as KotlinJsIrTarget).wasmTargetType != null -> wasmOutputName
+            else -> jsOutputName
+        }
+
+        outputName.set(target.outputModuleName)
```
terrakok: outputModuleName = root project name + module name

that is not what we expected here

terrakok (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/tasks/WebCompatibilityTask.kt
terrakok: you can add a log check into one of existed tests for other cases with only one of the web targets

PR #5379, opened 2025-08-07
Use Web Cache API for extra layer of resources caching

eymar:
Use Web Cache API for all resources 

The Cache is reset on every app launch (page refresh).
The initial idea was to reset the Cache only when a new session starts, but we risk to have an outdated resources state (it can be incompatible with the app logic expectations and lead to crashes).

Fixes https://youtrack.jetbrains.com/issue/CMP-7996 

## Testing
This should be tested by QA

### Fixes - Resources
- Use Web Cache API for all resources to avoid repeated and redundant HTTP requests

### DIFF
--- a/components/resources/demo/shared/build.gradle.kts
+++ b/components/resources/demo/shared/build.gradle.kts
@@ -69,10 +69,6 @@ kotlin {
             nativeMain.get().dependsOn(this)
             jsMain.get().dependsOn(this)
         }
-
-        webMain.dependencies {
-            implementation(libs.kotlinx.browser)
-        }
     }
 }
 


eymar:
cc: @terrakok when you are back :) 

I'd like to include this into 1.10

components/resources/library/src/webMain/kotlin/org/jetbrains/compose/resources/ResourceWebCache.web.kt
```diff
@@ -0,0 +1,70 @@
...
+
+import kotlinx.browser.window
+import kotlinx.coroutines.await
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import org.w3c.fetch.Response
+import org.w3c.workers.Cache
+
+/**
+ * We use [Cache] and [org.w3c.dom.WindowSessionStorage] APIs to cache the successful strings.cvr responses.
```
terrakok: If we use it specifically for Strings.CVR then let's rename it to StringsWebCache. Otherwise, we mustn't mention cvr in the kdoc

eymar: Now it's called `ResourceWebCache`. It's applied for all requests which go through web resource reader (in `readAsBlob`).

components/resources/demo/shared/build.gradle.kts
```diff
@@ -69,6 +69,10 @@ kotlin {
             nativeMain.get().dependsOn(this)
             jsMain.get().dependsOn(this)
         }
+
+        webMain.dependencies {
+            implementation(libs.kotlinx.browser)
+        }
```
terrakok: is it required in the demo?

eymar: removed it

PR #5380, opened 2025-08-08
Introduce test for invoking composeCompatibilityBrowserDistribution in heavily reconfigured project

Schahen:
CompatibilityDistributionTest::testWebJsWasmReconfigured makes sure that composeCompatibilityBrowserDistribution task can be reconfigured to be used with arbitrary set of input tasks

## Testing
Pipeline

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/CompatibilityDistributionTest.kt
+++ b/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/CompatibilityDistributionTest.kt
@@ -1,66 +1,110 @@
 package org.jetbrains.compose.test.tests.integration
 
-import org.gradle.internal.impldep.junit.framework.TestCase.assertTrue
 import org.jetbrains.compose.test.utils.GradlePluginTestBase
+import org.jetbrains.compose.test.utils.TestProject
 import org.jetbrains.compose.test.utils.checkExists
 import org.jetbrains.compose.test.utils.checks
 import org.junit.jupiter.api.Test
-import kotlin.test.assertContentEquals
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
 
 class CompatibilityDistributionTest : GradlePluginTestBase() {
-    @Test
-    fun testWebJsWasm() = with(
+    private val defaultDistDir = "./composeApp/build/dist/composeWebCompatibility/productionExecutable"
+
+    private fun TestProject.assertCompatibilityDistribution(
+        dirPath: String,
+        expectedFileNames: Set<String>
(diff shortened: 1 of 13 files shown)


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/CompatibilityDistributionTest.kt
```diff
@@ -1,66 +1,110 @@
...
 import org.jetbrains.compose.test.utils.GradlePluginTestBase
+import org.jetbrains.compose.test.utils.TestProject
 import org.jetbrains.compose.test.utils.checkExists
 import org.jetbrains.compose.test.utils.checks
 import org.junit.jupiter.api.Test
-import kotlin.test.assertContentEquals
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
 
 class CompatibilityDistributionTest : GradlePluginTestBase() {
```
igordmn:
```suggestion
class WebCompatibilityDistributionTest : GradlePluginTestBase() {
```
(and the file name)

igordmn (approved):

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/WebCompatibilityDistributionTest.kt
```diff
@@ -0,0 +1,110 @@
...
+                ?: error("Expected to list files in $dirPath but got null")
+            val distributionNames = listed.map { it.name }
+
+            assertTrue(
+                distributionNames.any { it.endsWith(".wasm") },
+                "Expected at least one .wasm file in $dirPath, found: $distributionNames"
+            )
+
+            val actualFiles = distributionNames.filterNot { it.endsWith(".wasm") }.toSet()
+            assertEquals(expectedFileNames, actualFiles, "files mismatch in $dirPath")
```
harrytmthy, line 32: NIT: The `.wasm` check uses `any { ... }`, which is lenient, while the `actualFiles` check uses strict `assertEquals(...)`, failing even if there's just one unexpected file (e.g. `Untitled.txt`).  

Should we make both checks consistent in strictness/leniency, depending on whether extra non-`.wasm` files should be allowed? 🤔

Schahen: Hey @harrytmthy ! You're raising valid questions, it's just that we are putting main effort into checking for actual name of the generated artefacts, so this check-for-wasm approach is a temporary trade-off.

PR #5385, opened 2025-08-17
Fix AGP 9.0.0-alpha01 compatibility for compose resources

simonlebras:
AGP 9.0.0-alpha01 removed onVariant{} API from KMP Variant extension. (https://cs.android.com/android-studio/platform/tools/base/+/1d00038ec037fb04b4c83e0e0612ba6f0d35cc8b)

Replacement is the plural onVariants{} API

### DIFF
--- a/gradle-plugins/gradle/libs.versions.toml
+++ b/gradle-plugins/gradle/libs.versions.toml
@@ -2,7 +2,7 @@
 kotlin = "2.2.0-RC2"
 gradle-download-plugin = "5.5.0"
 kotlin-poet = "2.1.0"
-plugin-android = "8.9.1"
+plugin-android = "8.10.1"
 shadow-jar = "8.1.1"
 publish-plugin = "1.2.1"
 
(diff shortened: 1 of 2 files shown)


gradle-plugins/gradle/libs.versions.toml
```diff
@@ -2,7 +2,7 @@
 kotlin = "2.2.0-RC2"
 gradle-download-plugin = "5.5.0"
 kotlin-poet = "2.1.0"
-plugin-android = "8.9.1"
+plugin-android = "8.10.1"
```
simonlebras, line 5: onVariants is only available with AGP 8.10+

terrakok: so, how does it work if user uses AGP 8.9?

simonlebras: Does not compile unfortunately. It either does not work on 8.9 (with the fix) or on 9.0 (without the fix).

terrakok:
https://youtrack.jetbrains.com/issue/CMP-8771

PR #5396, opened 2025-08-25
Declare generateResourceContentHashAnnotation property for GenerateResourceAccessorsTask

pjBooms:
- to allow configuring it from other tasks

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/ComposeResourcesGeneration.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/ComposeResourcesGeneration.kt
@@ -153,6 +153,8 @@ private fun Project.configureResourceAccessorsGeneration(
             task.packagingDir.set(packagingDir)
         }
         task.onlyIf { shouldGenerateCode.get() }
+        task.generateResourceContentHashAnnotation.set(
+            System.getProperty("compose.resources.generate.ResourceContentHash.annotation") == "true")
     }
 
     //register generated source set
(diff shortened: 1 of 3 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/ComposeResourcesGeneration.kt
```diff
@@ -153,6 +153,8 @@ private fun Project.configureResourceAccessorsGeneration(
             task.packagingDir.set(packagingDir)
         }
         task.onlyIf { shouldGenerateCode.get() }
+        task.generateResourceContentHashAnnotation.set(
+            System.getProperty("compose.resources.generate.ResourceContentHash.annotation") == "true")
```
terrakok:
```kotlin
project.providers
  .systemProperty("compose.resources.generate.ResourceContentHash.annotation")
  .map { it.toBoolean() }
  .orElse(false)
```

terrakok (changes requested):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/ComposeResourcesGeneration.kt
terrakok: it works correctly with a configuration cache

pjBooms: done

PR #5399, opened 2025-08-28
Bump versions (Kotlin, AGP, Gradle)

MatkovIvan:
## Release Notes
N/A


examples/codeviewer/gradle.properties
```diff
@@ -10,6 +10,6 @@ kotlin.native.useEmbeddableCompilerJar=true
 kotlin.mpp.androidSourceSetLayoutVersion=2
 # Enable kotlin/native experimental memory model
 kotlin.native.binary.memoryModel=experimental
-kotlin.version=2.1.20
-agp.version=8.5.0
-compose.version=1.8.2
+kotlin.version=2.2.10
+agp.version=8.9.0
+compose.version=1.9.0-rc01
```
igordmn: We shouldn't use non-stable version in the examples

igordmn: We should keep 1.8.2, and it should work locally and on the CI with the building version

igordmn (changes requested):

examples/codeviewer/gradle.properties
MatkovIvan: Done

PR #5403, opened 2025-08-28
Revert compose in gradle tests

MatkovIvan:
After combination of #5399 and #5400

Build fails with
```
org.gradle.internal.resolve.ModuleVersionNotFoundException: Could not find org.jetbrains.compose.material3:material3:1.9.0-rc01.
```

It happens because material3 is uncoupled in that version, but if gradle plugin is built from sources with passing compose version as 1.9.0-rc01, and using it as material3 alias - it's just invalid state. Revert to 1.8.2 as hotfix.

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/gradle.properties
+++ b/gradle-plugins/gradle.properties
@@ -8,7 +8,7 @@ kotlin.code.style=official
 dev.junit.parallel=false
 
 # Default version of Compose Libraries used by Gradle plugin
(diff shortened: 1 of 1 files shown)


igordmn:
Fixed in 
https://github.com/JetBrains/compose-multiplatform/pull/5404
https://github.com/JetBrains/compose-multiplatform/pull/5405

PR #5414, opened 2025-09-01
Remove non-publishing libraries from Changeling generating script

ASalavei:
## Release Notes
N/A


tools/changelog/changelog.main.kts
```diff
@@ -221,9 +221,7 @@ fun generateChangelog() {
 
                     - Compose Material3 libraries `org.jetbrains.compose.material3:material3*:$versionComposeMaterial3`. Based on [Jetpack Compose Material3 $versionRedirectingComposeMaterial3](https://developer.android.com/jetpack/androidx/releases/compose-material3#$versionRedirectingComposeMaterial3)
```
igordmn, line 222: This can be also reverted (make the same as the other libraries) for the time being, until we decide to hardcode it again

tools/changelog/changelog.main.kts
```diff
@@ -221,9 +221,7 @@ fun generateChangelog() {
```
igordmn, line 215: This can be also reverted (make the same as the other libraries) for the time being, until we decide to hardcode it again

ASalavei: Done this as well as small cleanup

PR #5425, opened 2025-09-11
Make ResourceEnvironment public

lamba92:
# Problem
Currently, `StringResources.getString(...)` and `ResourceEnvironment` are only convenient to use inside a composition. For **server-side rendering (SSR)** with Kotlin/WasmJS, we need to serve html pages that are SEO-friendly and localized. This requires access to string resources **outside composition** and the ability to configure a `ResourceEnvironment` manually.

# Proposed Changes
1. **ResourceEnvironment**
   - Made public (`language`, `region`, `theme`, `density` now public `val`s).
   - Added a `copy(...)` helper to allow modification without exposing a `data class`.

2. **StringResources**
   - Added overloads of `getString(...)` that accept a `ResourceReader` parameter (default: `DefaultResourceReader`).
   - This enables reading localized strings outside of composition while keeping composition-aware behavior unchanged.

# Motivation
- SSR requires rendering UI to HTML on the server, not just in the browser.
- Without these changes, we cannot reliably access localized strings during SSR, leading to incomplete or incorrect rendering.
- The ability to override `ResourceEnvironment` is crucial for generating pages in multiple locales programmatically.

# Open Questions
- The current PR only adds support for **strings**. Should we extend a similar pattern to other resource types (images, fonts, etc.) for consistency?
- Binary compatibility: do we need to preserve exact overloads of the old `getString` signatures to avoid breaking precompiled binaries?This change allows to access resources outside the composition while also allowing configuring ResourceEnvironment.

The reason behind this change is to simplify the generation of SSR pages for Kotlin/WasmJs while also respecting localization.

### DIFF
--- a/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
+++ b/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
@@ -5,11 +5,12 @@ import androidx.compose.runtime.*
 import androidx.compose.ui.platform.LocalDensity
 import androidx.compose.ui.text.intl.Locale
 
-class ResourceEnvironment internal constructor(
-    internal val language: LanguageQualifier,
-    internal val region: RegionQualifier,
-    internal val theme: ThemeQualifier,
-    internal val density: DensityQualifier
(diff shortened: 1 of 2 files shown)


lamba92:
I updated the PR details. What do you think about it?

PR #5428, opened 2025-09-15
Changelog script. Add protection to check that it was cloned successfully

igordmn:
I had a situation when the folder was empty, and the git log collected commits from the parent repo compose-multiplatform instead of compose-multiplatform-core

## Release Notes
N/A


ASalavei (approved):
It's a bit strange that it accidentally got files from a different repo - maybe you're putting compose-multiplatform and compose-multiplatform-core in the same directory?

PR #5433, opened 2025-09-17
Don't throw an error if a context is null in the DefaultAndroidResourceReader

terrakok:
It fixes Robolectric tests where a context is available through InstrumentationRegistry

Fixes https://youtrack.jetbrains.com/issue/CMP-6612

### Fixes - Resources
- Fix resource access in a Robolectric test environment.

### DIFF
--- a/components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceReader.android.kt
+++ b/components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceReader.android.kt
@@ -13,13 +13,7 @@ internal actual fun getPlatformResourceReader(): ResourceReader = DefaultAndroid
 
 @ExperimentalResourceApi
 internal object DefaultAndroidResourceReader : ResourceReader {
-    private val assets: AssetManager by lazy {
-        val context = androidContext ?: error(
-            "Android context is not initialized. " +
-                    "If it happens in the Preview mode then call PreviewContextConfigurationEffect() function."
-        )
-        context.assets
-    }
+    private val assets: AssetManager? get() = androidContext?.assets
 
     private val instrumentedAssets: AssetManager?
         get() = try {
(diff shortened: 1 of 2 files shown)


components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceReader.android.kt
```diff
@@ -13,13 +13,7 @@ internal actual fun getPlatformResourceReader(): ResourceReader = DefaultAndroid
...
 @ExperimentalResourceApi
 internal object DefaultAndroidResourceReader : ResourceReader {
-    private val assets: AssetManager by lazy {
-        val context = androidContext ?: error(
-            "Android context is not initialized. " +
-                    "If it happens in the Preview mode then call PreviewContextConfigurationEffect() function."
-        )
-        context.assets
-    }
+    private val assets: AssetManager? get() = androidContext?.assets
```
igordmn, line 16: It looks like, it shouldn't compile after you made it nullable?
```
override fun getUri(path: String): String {
        val uri = if (assets.hasFile(path) || instrumentedAssets.hasFile(path)) {
```

terrakok: Why? This is extension on a nullable type

terrakok: https://github.com/JetBrains/compose-multiplatform/blob/7bb744474f1c8a2fdfc822dc4ec2a04e50e50ce0/components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceReader.android.kt#L101

igordmn: Okay. Didn't look at it

PR #5436, opened 2025-09-22
Add ui-backhandler dependency shortcut

MaximPestryakov:
In [1.8.0 release](https://github.com/JetBrains/compose-multiplatform/releases/tag/v1.8.0) multiplatform `BackHandler` and `PredictiveBackHandler` were implemented. But shortcut for `ui-backhandler` dependency wasn't. To be able to use 
`BackHandler` you have to manually add `org.jetbrains.compose.ui:ui-backhandler` dependency.

Now you add the dependency like this:
```kotlin
dependencies {
    implementation(compose.uiBackHandler)
}
```


kropp:
We are deprecating dependency aliases in #5462, so I'm closing this PR.
These aliases added unnecessary magic to the users' build files and unnecessary complexity to our builds.

PR #5439, opened 2025-09-29
Add support of important in compose html css dsl

InsanusMokrassar:
Fixes https://youtrack.jetbrains.com/issue/CMP-8992

## Testing

Add test in `org.jetbrains.compose.web.core.tests.css.StyleSheetTests`

### Features - HTML
- Add support of `important` keyword for CSS properties

### DIFF
--- a/html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
+++ b/html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
@@ -64,10 +64,10 @@ interface StyleScope {
     fun property(propertyName: String, value: StylePropertyValue, important: Boolean): Unit = error("!important is not supported by this implementation")
     fun variable(variableName: String, value: StylePropertyValue)
 
+    fun property(propertyName: String, value: String) = property(propertyName, StylePropertyValue(value))
     fun property(propertyName: String, value: String, important: Boolean) = property(propertyName, StylePropertyValue(value), important)
-    fun property(propertyName: String, value: String) = property(propertyName, value, false)
+    fun property(propertyName: String, value: Number) = property(propertyName, StylePropertyValue(value))
     fun property(propertyName: String, value: Number, important: Boolean) = property(propertyName, StylePropertyValue(value), important)
-    fun property(propertyName: String, value: Number) = property(propertyName, value, false)
     fun variable(variableName: String, value: String) = variable(variableName, StylePropertyValue(value))
     fun variable(variableName: String, value: Number) = variable(variableName, StylePropertyValue(value))
 


InsanusMokrassar:
Hello :) mr. @eymar is there any news when this PR will be reviewed? Can't wait to fix some problems here and see this PR in `master` :)

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
```diff
@@ -190,6 +194,7 @@ internal fun StylePropertyList.nativeEquals(properties: StylePropertyList): Bool
     return all { prop ->
         val otherProp = properties[index++]
         prop.name == otherProp.name &&
-            prop.value.toString() == otherProp.value.toString()
+            prop.value.toString() == otherProp.value.toString() &&
+                prop.important == otherProp.important
```
eymar:
```suggestion
            prop.important == otherProp.important && prop.value.toString() == otherProp.value.toString()
                
```

changing the order of the checks can help a bit - it's better to compare the booleans first.

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
```diff
@@ -175,10 +178,11 @@ open class StyleScopeBuilder : StyleScope, StyleHolder {
 
 data class StylePropertyDeclaration(
     val name: String,
-    val value: StylePropertyValue
+    val value: StylePropertyValue,
+    val important: Boolean = false,
```
eymar: Adding a new parameter here makes the constructors not binary compatible, because Kotlin doesn't create the overloads - it adds an extra bitmask parameter to indicate the presence of default values.

I guess that the only way is to add the overloads for every constructor and keep the new property separately + override the equals() and hashCode(). 

___

I think ideally this class would be a regular class not a `data` class, but we can't change it now because we want to keep it compatible.

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/dom/Base.kt
```diff
@@ -67,8 +67,9 @@ private class DomElementWrapper(override val node: Element): DomNodeWrapper(node
 
                 val style = node.unsafeCast<ElementCSSInlineStyle>().style
 
-                styleApplier.properties.forEach { (name, value) ->
-                    style.setProperty(name, value.toString())
+                styleApplier.properties.forEach { (name, value, important) ->
+                    if (important) error("important")
```
eymar: why does it throw an error when `important` is true?

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
```diff
@@ -42,11 +42,14 @@ interface StyleScope {
      * })
      * ```
      */
-    fun property(propertyName: String, value: StylePropertyValue)
+    fun property(propertyName: String, value: StylePropertyValue, important: Boolean)
```
eymar: The new methods in an interface should have a default implementaiton to remain source compatible:

The existing implementations (I mean potential implementations not in this project) implement the methods with 2 parameters.

I see two options for the default implementations of new 3arg methods:
-  just throw an error - "!important is not supported by this implementation"
- call the 2arg overloads dropping the `important` parameter

Usually, when an error is thrown then the problem will be noticed much earlier. So I'd choose the first option. WDYT?

Then we just make sure that our implementation of the interface overrides the new 3arg methods rather than uses the default (throwing an error).

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
InsanusMokrassar: Added overrides of constructors. I am not sure that `hashCode` and `equals` require manual realization, because old calls of `StylePropertyDeclaration` constructors will call constructor with `important` value `false` and they all will work as they have worked before. Correct me if I am wrong

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/dom/Base.kt
InsanusMokrassar: Fixed

InsanusMokrassar:
@eymar thanks for your review :) I have fixed code according to it, could you please check it again?

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
InsanusMokrassar: I agree with you, implemented first variant

html/core/src/jsMain/kotlin/org/jetbrains/compose/web/css/StyleScope.kt
```diff
@@ -43,10 +43,31 @@ interface StyleScope {
...
+     * })
+     * ```
+     */
+    fun property(propertyName: String, value: StylePropertyValue, important: Boolean): Unit = error("!important is not supported by this implementation")
     fun variable(variableName: String, value: StylePropertyValue)
 
-    fun property(propertyName: String, value: String) = property(propertyName, StylePropertyValue(value))
-    fun property(propertyName: String, value: Number) = property(propertyName, StylePropertyValue(value))
+    fun property(propertyName: String, value: String, important: Boolean) = property(propertyName, StylePropertyValue(value), important)
+    fun property(propertyName: String, value: String) = property(propertyName, value, false)
```
eymar: The "old" methods must not delegate to the new 3-arg overload. 

Imagine an implementation of old `interface StyleScope`, before this PR it had to implement only `fun property(propertyName: String, value: StylePropertyValue)`.  The overloads with `value: String` and `value: Number` called the `value: StylePropertyValue` overload and it worked.

But now  the 2-args overloads call a 3-arg overload. If my implementaiton doesn't implement the 3-arg overload, then my existing 2-args calls would stop working too (throw an error). 

Let me know if it's not clear what I mean.

InsanusMokrassar: Got it, I will fix it asap in the way I did it for base property realization

InsanusMokrassar: Fixed, please, check it

PR #5442, opened 2025-10-01
Deprecate plugin

kropp:
Removes all functionality from the plugin and adds a feature to suggest new Kotlin Multiplatform plugin as replacement.

Fixes https://youtrack.jetbrains.com/issue/CMP-9002

## Testing
Tested manually

## Release Notes
N/A


idea-plugin/build.gradle.kts
```diff
@@ -34,9 +34,9 @@ dependencies {
 
 intellijPlatform {
     pluginConfiguration {
-        name = "Compose Multiplatform IDE Support"
+        name = "Compose Multiplatform for Desktop IDE Support"
```
igordmn: > After a consideration, it is probably not worth the effort releasing the plugin again just to disable the functionality.

Do we not plan to release it? Can we then just remove this folder entirely?

kropp: Decided to postpone until Android Studio is supported

igordmn: Do you mean removing the published plugin?

Do we need to release a new version of it with changes in this PR? 

If not, maybe we can keep the code as is? And remove when we decide to remove the published plugin?

kropp: We don't need to release any new version now. However, at the same time we should not completely remove the plugin just yet for a (very unlikely) case we need to have an urgent hotfix for Windows/Linux users of Android Studio.
We will get back to the discussion early next year.

PR #5444, opened 2025-10-03
Bundle Compose Hot Reload Gradle plugin into the Compose Gradle plugin

pjBooms:
Compose Hot Reload Gradle plugin are bundled into the Compose Gradle plugin.

Fixes [CMP-8885](https://youtrack.jetbrains.com/issue/CMP-8885/Bundle-Compose-Hot-Reload-with-Compose-Multiplatform)

## Testing

Integration tests added

This should be tested by QA

### Highlights - Multiple Platforms
- Compose Hot Reload Gradle plugin is bundled with Compose Gradle plugin (non need to configure it separately)

### DIFF
--- a/gradle-plugins/compose/build.gradle.kts
+++ b/gradle-plugins/compose/build.gradle.kts
@@ -1,5 +1,9 @@
+import com.github.jengelman.gradle.plugins.shadow.relocation.SimpleRelocator
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
+import com.github.jengelman.gradle.plugins.shadow.transformers.TransformerContext
 import de.undercouch.gradle.tasks.download.Download
+import org.apache.tools.zip.ZipEntry
+import org.apache.tools.zip.ZipOutputStream
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask
 
 plugins {
@@ -25,12 +29,16 @@ mavenPublicationConfig {
 
 val buildConfigDir
     get() = project.layout.buildDirectory.dir("generated/buildconfig")
+
+val hotReloadVersion = BuildProperties.hotReloadVersion(project)
+
 val buildConfig = tasks.register("buildConfig", GenerateBuildConfig::class.java) {
     classFqName.set("org.jetbrains.compose.ComposeBuildConfig")
     generatedOutputDir.set(buildConfigDir)
     fieldsToGenerate.put("composeVersion", BuildProperties.composeVersion(project))
     fieldsToGenerate.put("composeMaterial3Version", BuildProperties.composeMaterial3Version(project))
     fieldsToGenerate.put("composeGradlePluginVersion", BuildProperties.deployVersion(project))
+    fieldsToGenerate.put("composeHotReloadVersion", hotReloadVersion)
 }
 tasks.named("compileKotlin", KotlinCompilationTask::class) {
     dependsOn(buildConfig)
@@ -56,6 +64,10 @@ dependencies {
         embeddedDependencies(dep)
     }
 
+    fun hotReloadDep(dep: String) = embedded(
+        "org.jetbrains.compose.hot-reload:$dep:$hotReloadVersion"
+    )
+
     compileOnly(gradleApi())
     compileOnly(localGroovy())
     compileOnly(kotlin("gradle-plugin"))
@@ -69,21 +81,75 @@ dependencies {
 
     embedded(libs.download.task)
     embedded(libs.kotlin.poet)
+    hotReloadDep("hot-reload-gradle-plugin")
+    hotReloadDep("hot-reload-gradle-core")
+    hotReloadDep("hot-reload-gradle-idea")
+    hotReloadDep("hot-reload-core")
+    hotReloadDep("hot-reload-orchestration")
+    hotReloadDep("hot-reload-annotations-jvm")
     embedded(project(":preview-rpc"))
     embedded(project(":jdk-version-probe"))
 }
 
+
 val packagesToRelocate = listOf("de.undercouch", "com.squareup.kotlinpoet")
+
+val relocationPackage = "org.jetbrains.compose.internal"
+
+val hotReloadPackage = "org.jetbrains.compose.reload"
+
+val hotReloadPackageRelocated = "$relocationPackage.$hotReloadPackage"
+
+val hotReloadPropertiesPath = "META-INF/gradle-plugins/org.jetbrains.compose.hot-reload.properties"
+
+private class HotReloadPropertiesTransformer() : com.github.jengelman.gradle.plugins.shadow.transformers.Transformer {
+    private val targetPath = "META-INF/gradle-plugins/org.jetbrains.compose.embedded.hot-reload.properties"
+    private val content = """
+        implementation-class=$hotReloadPackageRelocated.gradle.ComposeHotReloadPlugin
+    """.trimIndent()
+
+    override fun canTransformResource(element: FileTreeElement?): Boolean = false
+    override fun transform(context: TransformerContext?) = Unit
+    override fun hasTransformedResource(): Boolean = true
+
+    override fun modifyOutputStream(os: ZipOutputStream?, preserveFileTimestamps: Boolean) {
+        val entry = ZipEntry(targetPath)
+        entry.time = TransformerContext.getEntryTimestamp(preserveFileTimestamps, entry.time)
+        os?.run {
+            putNextEntry(entry)
+            write(content.toByteArray())
+            closeEntry()
+        }
+    }
+
+    override fun getName(): String = "$hotReloadPropertiesPath transformer"
+}
+
+fun ShadowJar.relocateHotReload() {
+    val relocator = object : SimpleRelocator(hotReloadPackage, hotReloadPackageRelocated, ArrayList(), ArrayList()) {
+        override fun canRelocatePath(path: String?): Boolean {
+            return super.canRelocatePath(path) &&
+                    // do not relocate orchestration as its objects are used in serialization.
+                    path?.startsWith("org/jetbrains/compose/reload/orchestration/") == false
+        }
+    }
+    relocate(relocator)
+}
 
 val shadow = tasks.named<ShadowJar>("shadowJar") {
     for (packageToRelocate in packagesToRelocate) {
-        relocate(packageToRelocate, "org.jetbrains.compose.internal.$packageToRelocate")
+        relocate(packageToRelocate, "$relocationPackage.$packageToRelocate")
     }
+    relocateHotReload()
+
+    transform(HotReloadPropertiesTransformer())
+
     archiveBaseName.set("shadow")
     archiveClassifier.set("")
     archiveVersion.set("")
     configurations = listOf(embeddedDependencies)
     exclude("META-INF/gradle-plugins/de.undercouch.download.properties")
+    exclude(hotReloadPropertiesPath)
     exclude("META-INF/versions/**")
 }
 
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
@@ -64,6 +64,14 @@ abstract class ComposePlugin : Plugin<Project> {
                 project.configureExperimentalTargetsFlagsCheck(mppExt)
             }
         }
+
+        try {
+            project.pluginManager.apply("org.jetbrains.compose.hot-reload")
+        } catch (_: Exception) {
+            // If a user does not set up the hot-reload plugin explicitly, set up the embedded one.
+            // TODO: issue a warning/error if the embedded version is higher than explicit one
+            project.pluginManager.apply("org.jetbrains.compose.embedded.hot-reload")
+        }
     }
 
     @Suppress("DEPRECATION")
--- a/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
+++ b/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
@@ -1,0 +1,83 @@
+package org.jetbrains.compose.test.tests.integration
+
+import org.gradle.testkit.runner.BuildResult
+import org.jetbrains.compose.ComposeBuildConfig
+import org.jetbrains.compose.test.utils.GradlePluginTestBase
+import org.jetbrains.compose.test.utils.checks
+import org.jetbrains.compose.test.utils.modify
+import org.junit.jupiter.api.Test
+import kotlin.concurrent.thread
+
+class HotReloadTest : GradlePluginTestBase() {
+    @Test
+    fun smokeTestHotRunTask() = with(testProject("application/jvm")) {
+        file("build.gradle").modify {
+            it + """
+                afterEvaluate {
+                    tasks.getByName("hotRun").doFirst {
+                        throw new StopExecutionException("Skip hotRun task")
+                    }
+                }
+            """.trimIndent()
+        }
+        gradle("hotRun").checks {
+            check.taskSuccessful(":hotRun")
+        }
+    }
+
+    @Test
+    fun testHotReload() = with(testProject("application/hotReload")) {
+        var result: BuildResult? = null
+        val hotRunThread = thread {
+            result = gradle("hotRunJvm")
+        }
+
+        while (!file("started").exists()) {
+            Thread.sleep(200)
+        }
+
+        modifyText("src/jvmMain/kotlin/main.kt") {
+            it.replace("Kotlin MPP", "KMP")
+        }
+
+        gradle("reload").checks {
+            check.taskSuccessful(":reload")
+            check.logContains("MainKt.class: modified")
+        }
+
+        hotRunThread.join()
+        check(result != null)
+        result.checks {
+            check.taskSuccessful(":hotRunJvm")
+            check.logContains("Kotlin MPP app is running!")
+            check.logContains("KMP app is running!")
+            check.logContains("Compose Hot Reload (${ComposeBuildConfig.composeHotReloadVersion})")
+        }
+    }
+
+    @Test
+    fun testExternalHotReload() = with(testProject("application/mpp")) {
+        val externalHotReloadVersion = "1.0.0-beta04"
+        modifyText("settings.gradle") {
+            it.replace(
+                "plugins {", "plugins {\n" +
+                        """
+                        id 'org.jetbrains.compose.hot-reload' version '$externalHotReloadVersion'
+                """.trimIndent()
+            )
+        }
+        modifyText("build.gradle") {
+            it.replace(
+                "plugins {", "plugins {\n" +
+                        """
+                        id "org.jetbrains.compose.hot-reload"
+                """.trimIndent()
+            )
+        }
+        gradle("hotRunJvm").checks {
+            check.taskSuccessful(":hotRunJvm")
+            check.logContains("Compose Hot Reload ($externalHotReloadVersion)")
+            check.logContains("Kotlin MPP app is running!")
+        }
+    }
+}
--- a/gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
+++ b/gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
@@ -1,0 +1,69 @@
+import org.jetbrains.compose.desktop.application.dsl.TargetFormat
+
+plugins {
+    id "com.android.application"
+    id "org.jetbrains.kotlin.multiplatform"
+    id "org.jetbrains.kotlin.plugin.compose"
+    id "org.jetbrains.compose"
+}
+
+kotlin {
+    // empty stub (no actual android app) to detect configuration conflicts
+    // like https://github.com/JetBrains/compose-jb/issues/2345
+    androidTarget()
+
+    jvm()
+    sourceSets {
+        jvmMain {
+            dependsOn(commonMain)
+
+            dependencies {
+                implementation(compose.desktop.currentOs)
+            }
+        }
+    }
+    jvmToolchain {
+        languageVersion.set(JavaLanguageVersion.of(11))
+    }
+}
+
+android {
+    namespace = "org.jetbrains.compose.testapp"
+    compileSdk = 35
+
+    defaultConfig {
+        minSdk = 23
+        targetSdk = 35
+    }
+}
+
+compose.desktop {
+    application {
+        mainClass = "MainKt"
+        nativeDistributions {
+            targetFormats(TargetFormat.Dmg, TargetFormat.Msi, TargetFormat.Deb)
+
+            packageVersion = "1.0.0"
+            packageName = "TestPackage"
+            description = "Test description"
+            copyright = "Test Copyright Holder"
+            vendor = "Test Vendor"
+
+            linux {
+                shortcut = true
+                packageName = "test-package"
+                debMaintainer = "example@example.com"
+                menuGroup = "menu-group"
+            }
+            windows {
+                console = true
+                dirChooser = true
+                perUserInstall = true
+                shortcut = true
+                menu = true
+                menuGroup = "compose"
+                upgradeUuid = "2d6ff464-75be-40ad-a256-56420b9cc374"
+            }
+        }
+    }
+}
--- a/gradle-plugins/gradle.properties
+++ b/gradle-plugins/gradle.properties
@@ -25,3 +25,4 @@ compose.tests.gradle-agp.exclude=8.7/8.9.0, 8.7/9.0.0-alpha01
 # A version of Gradle plugin, that will be published,
 # unless overridden by COMPOSE_GRADLE_PLUGIN_VERSION env var.
 deploy.version=9999.0.0-SNAPSHOT
+hotreload.version=1.0.0-beta08
--- a/gradle-plugins/buildSrc/src/main/kotlin/BuildProperties.kt
+++ b/gradle-plugins/buildSrc/src/main/kotlin/BuildProperties.kt
@@ -23,4 +23,6 @@ object BuildProperties {
     fun deployVersion(project: Project): String =
         System.getenv("COMPOSE_GRADLE_PLUGIN_VERSION")
             ?: project.findProperty("deploy.version") as String
+    fun hotReloadVersion(project: Project): String =
+        project.findProperty("hotreload.version") as String
 }
--- a/gradle-plugins/compose/src/test/test-projects/application/hotReload/gradle.properties
+++ b/gradle-plugins/compose/src/test/test-projects/application/hotReload/gradle.properties
@@ -1,0 +1,1 @@
+android.useAndroidX=true
--- a/gradle-plugins/compose/src/test/test-projects/application/hotReload/settings.gradle
+++ b/gradle-plugins/compose/src/test/test-projects/application/hotReload/settings.gradle
@@ -1,0 +1,34 @@
+pluginManagement {
+    plugins {
+        id 'org.jetbrains.kotlin.multiplatform' version 'KOTLIN_VERSION_PLACEHOLDER'
+        id 'org.jetbrains.kotlin.plugin.compose' version 'KOTLIN_VERSION_PLACEHOLDER'
+        id 'org.jetbrains.compose' version 'COMPOSE_GRADLE_PLUGIN_VERSION_PLACEHOLDER'
+        id 'com.android.application' version 'AGP_VERSION_PLACEHOLDER'
+    }
+    repositories {
+        mavenLocal()
+        gradlePluginPortal()
+        mavenCentral()
+        google()
+        maven {
+            url 'https://maven.pkg.jetbrains.space/public/p/compose/dev'
+        }
+        maven {
+            url 'https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev/'
+        }
+    }
+}
+dependencyResolutionManagement {
+    repositories {
+        mavenCentral()
+        google()
+        maven {
+            url 'https://maven.pkg.jetbrains.space/public/p/compose/dev'
+        }
+        maven {
+            url 'https://maven.pkg.jetbrains.space/kotlin/p/kotlin/dev/'
+        }
+        mavenLocal()
+    }
+}
+rootProject.name = "mpp"
--- a/gradle-plugins/compose/src/test/test-projects/application/hotReload/src/jvmMain/kotlin/main.kt
+++ b/gradle-plugins/compose/src/test/test-projects/application/hotReload/src/jvmMain/kotlin/main.kt
@@ -1,0 +1,13 @@
+import java.io.File
+
+fun message() = "Kotlin MPP app is running!"
+
+fun main() {
+    println(message())
+    File("started").createNewFile()
+    //wait for reload
+    while(!message().startsWith("KMP")){
+        Thread.sleep(200)
+    }
+    println(message())
+}


gradle-plugins/gradle.properties
```diff
@@ -25,3 +25,4 @@ compose.tests.gradle-agp.exclude=8.7/8.9.0, 8.7/9.0.0-alpha01
 # A version of Gradle plugin, that will be published,
 # unless overridden by COMPOSE_GRADLE_PLUGIN_VERSION env var.
 deploy.version=9999.0.0-SNAPSHOT
+hotreload.version=1.0.0-beta08
```
MatkovIvan: I guess Ii's supposed to be configured on CI, here is the place only for some placeholder

cc @igordmn

gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+    sourceSets {
+        jvmMain {
+            dependsOn(commonMain)
+
+            dependencies {
+                implementation(compose.desktop.currentOs)
+            }
+        }
+    }
+    jvmToolchain {
```
MatkovIvan: It seems it will require JDK 11 to be installed that won't be a case on CI soon.
I'd suggest approach like here - https://github.com/JetBrains/compose-multiplatform/commit/51a87ca55cd5bc7e8b9cc78b6c7c1531bda2d96a

gradle-plugins/gradle.properties
igordmn: It is better to configure it only here, not on CI.

Because it is a dependency version, not a version of the building component.

igordmn: It is usually better to define such versions in `lib.versions.toml`, not in `gradle.properties`

pjBooms: I have to have a place for version declaration that I can use in build script as well as in tests. I can't use `lib.versions.toml`  from tests.  I see that versions are exported to sources via `ComposeBuildConfig` generation. And `ComposeBuildConfig` is generated based on `gradle.properties`. Though I can use `lib.versions.toml` from `build.gradle.kts` and export it via `ComposeBuildConfig` but, I am not sure here because it won't be aligned with other declarations.

gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
pjBooms: It is copy&pasted from other test projects. I may change it just for this one, or we may change for all the tests in a separate PR.

gradle-plugins/gradle.properties
igordmn: > Though I can use lib.versions.toml from build.gradle.kts and export it via ComposeBuildConfig but, I am not sure here because it won't be aligned with other declarations.

It is not so important, but I suggested this because of the different nature from the build perspective - HotReload is hardcoded dependency that is built separately, Compose/Material3 on the other hand are non-hardcoded components passed from CI. That is why, it is probably better to have them configured differently.

From user (not build) perspective, they probably have the same nature - they are all versions of something provided by the plugin. That is why, it is probably better to have them in one place in `ComposeBuildConfig`.

The edge between component/dependency and hardcoded/passed from CI is thin, so any solution in the end should work, choose one you think is better.

gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
MatkovIvan: If we have it accross this tests - it's ok to keep it as-is for now

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -64,6 +64,14 @@ abstract class ComposePlugin : Plugin<Project> {
                 project.configureExperimentalTargetsFlagsCheck(mppExt)
             }
         }
+
+        try {
+            project.pluginManager.apply("org.jetbrains.compose.hot-reload")
+        } catch (_: Exception) {
```
sellmair: Catching _any_ exception might swallow issues? How about catching only the exception we get when the plugin is missing?

gradle-plugins/compose/build.gradle.kts
```diff
@@ -69,21 +81,75 @@ dependencies {
...
+
+    override fun getName(): String = "Hot reload properties transformer"
+}
+
+fun ShadowJar.relocateHotReload() {
+    val relocator = object : SimpleRelocator(hotReloadPackage, hotReloadPackageRelocated, ArrayList(), ArrayList()) {
+        override fun canRelocatePath(path: String?): Boolean {
+            return super.canRelocatePath(path) &&
+                    // do not relocate orchestration as its objects are used in serialization.
+                    path?.startsWith("org/jetbrains/compose/reload/orchestration/") == false
```
sellmair: As of right now, we are locked out with this attempt: 
We cannot shade the orchestration as this breaks java.io.Serializable, we cannot embedd it either because of classpath conflicts

sellmair (changes requested):
Waiting for the approach with the 'prefrerred' dependency

gradle-plugins/gradle.properties
pjBooms: Now version catalog is used

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
pjBooms: We don't use embedded version anymore so `try` is not needed now

gradle-plugins/compose/build.gradle.kts
pjBooms: So yes, we use explicit dependency now

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -64,6 +64,8 @@ abstract class ComposePlugin : Plugin<Project> {
                 project.configureExperimentalTargetsFlagsCheck(mppExt)
             }
         }
+
+        project.pluginManager.apply("org.jetbrains.compose.hot-reload")
```
terrakok: I know we need to fake apply all plugins in the root project to avoid problems with a class loading. How does it work in that case? When we do
```
plugins {
  id("org.jetbrains.compose.multiplatform").apply(false)
}
```
I guess this code is not being executed.

gradle-plugins/compose/build.gradle.kts
```diff
@@ -56,6 +57,16 @@ dependencies {
         embeddedDependencies(dep)
     }
 
+    fun hotReloadDep(dep: Any) {
+        // We want to explicitly depend on the Hot Reload Gradle plugin
```
terrakok: why do we want it?

gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+            windows {
+                console = true
+                dirChooser = true
+                perUserInstall = true
+                shortcut = true
+                menu = true
+                menuGroup = "compose"
+                upgradeUuid = "2d6ff464-75be-40ad-a256-56420b9cc374"
+            }
+        }
```
terrakok: is it required for the test?

gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
```diff
@@ -0,0 +1,69 @@
...
+
+kotlin {
+    // empty stub (no actual android app) to detect configuration conflicts
+    // like https://github.com/JetBrains/compose-jb/issues/2345
+    androidTarget()
+
+    jvm()
+    sourceSets {
+        jvmMain {
+            dependsOn(commonMain)
```
terrakok: why is manual configuration here?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -0,0 +1,101 @@
...
+import org.junit.jupiter.api.fail
+import org.junit.jupiter.api.Test
+import kotlin.concurrent.thread
+
+class HotReloadTest : GradlePluginTestBase() {
+    @Test
+    fun smokeTestHotRunTask() = with(testProject("application/jvm")) {
+        file("build.gradle").modify {
+            it + """
+                afterEvaluate {
```
terrakok: why is "afterEvaluate" here?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -0,0 +1,101 @@
...
+                    tasks.getByName("hotRun").doFirst {
+                        throw new StopExecutionException("Skip hotRun task")
+                    }
+                }
+            """.trimIndent()
+        }
+        gradle("hotRun").checks {
+            check.taskSuccessful(":hotRun")
```
terrakok: The task should run with an error, shouldn't it?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -0,0 +1,101 @@
...
+    @Test
+    fun testExternalHotReload() = with(testProject("application/mpp")) {
+        val externalHotReloadVersion = "1.0.0-beta04"
+        modifyText("settings.gradle") {
+            it.replace(
+                "plugins {", "plugins {\n" +
+                        """
+                        id 'org.jetbrains.compose.hot-reload' version '$externalHotReloadVersion'
+                """.trimIndent()
+            )
```
terrakok, line 105: could you format it?
```
                                "plugins {\n" +
                        """
                        id 'org.jetbrains.compose.hot-reload' version '$externalHotReloadVersion'
                """.trimIndent()
```

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -0,0 +1,101 @@
...
+            check.logContains("Kotlin MPP app is running!")
+            check.logContains("KMP app is running!")
+            check.logContains("Compose Hot Reload (${ComposeBuildConfig.composeHotReloadVersion})")
+        }
+    }
+
+    @Test
+    fun testExternalHotReload() = with(testProject("application/mpp")) {
+        val externalHotReloadVersion = "1.0.0-beta04"
+        modifyText("settings.gradle") {
```
terrakok, line 97: why do we need to apply CHR plugin to the settings here? if it is required then why the compose plugin doesn't do that?

gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
pjBooms: No, I will remove it.

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: I did it the same as "DesktopApplicationTest:smokeTestRunTask". 
I guess the idea behind it is to check that there is the task available for run, can be run, but we do not actually perform the run.

gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
pjBooms: Removed

gradle-plugins/compose/src/test/test-projects/application/hotReload/build.gradle
pjBooms: Removed

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: Rewrote the test to use `--dry-run`

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: Rewrote the test to use `--dry-run`

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: done

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: It is `pluginManagement {` block. Added a comment about it

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureHotReload.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.gradle.api.Project
+import org.jetbrains.compose.internal.KOTLIN_JVM_PLUGIN_ID
+import org.jetbrains.compose.internal.KOTLIN_MPP_PLUGIN_ID
+import org.jetbrains.compose.internal.mppExt
+import org.jetbrains.kotlin.gradle.targets.jvm.KotlinJvmTarget
+
+private fun Project.isJVMTargetAvailable(): Boolean {
+    return plugins.findPlugin(KOTLIN_JVM_PLUGIN_ID) != null ||
+           plugins.findPlugin(KOTLIN_MPP_PLUGIN_ID)
+               ?.let { mppExt.targets.withType(KotlinJvmTarget::class.java).isNotEmpty() } == true
```
terrakok: You are supposed to listen targets collection and to react when there is jvm target. At the moment you do a copy of current targets. If a target is added later, you will not handle it

pjBooms: fixed

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
pjBooms: The code will be executed at the time of applying the plugin in subprojects.

gradle-plugins/compose/build.gradle.kts
pjBooms: Refined the comment

.github/workflows/gradle-plugin.yml
```diff
@@ -24,8 +24,10 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-java@v4
         with:
-          distribution: 'corretto'
-          java-version: '17'
+          distribution: 'jetbrains'
```
terrakok, line 27: Have you tried to use two distributions here? `distribution: 'jetbrains', 'corretto'`

I see they support few versions simultaneously. Might be it works with distribution as well

https://github.com/actions/setup-java?tab=readme-ov-file#install-multiple-jdks

terrakok: It would be good to ask them about the feature too

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureHotReload.kt
```diff
@@ -0,0 +1,28 @@
...
+
+import org.gradle.api.Project
+import org.jetbrains.compose.internal.KOTLIN_JVM_PLUGIN_ID
+import org.jetbrains.compose.internal.KOTLIN_MPP_PLUGIN_ID
+import org.jetbrains.compose.internal.javaSourceSets
+import org.jetbrains.compose.internal.mppExt
+import org.jetbrains.kotlin.gradle.targets.jvm.KotlinJvmTarget
+
+private fun Project.applyHotReload() {
+    project.pluginManager.apply("org.jetbrains.compose.hot-reload")
```
terrakok, line 15: Please, add here an comment about how the version is getting from the `pom.xml`

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -0,0 +1,111 @@
...
+import org.junit.jupiter.api.Disabled
+import org.junit.jupiter.api.fail
+import org.junit.jupiter.api.Test
+import kotlin.concurrent.thread
+
+class HotReloadTest : GradlePluginTestBase() {
+    @Test
+    fun smokeTestHotRunTask() = with(testProject("application/jvm")) {
+        gradle("hotRun", "--dry-run").checks {
+            check.logContains(":hotRun")
```
terrakok: `check.taskSkipped(":hotRun")`

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -0,0 +1,111 @@
...
+            //  Apply hot reload plugin explicitly
+            it.replace(
+                "plugins {",
+                """
+                        plugins {
+                            id "org.jetbrains.compose.hot-reload"
+                """.trimIndent()
+            )
+        }
+        gradle("hotRunJvm").checks {
```
terrakok: Why are there two different task names?
`hotRun` and `hotRunJvm`

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -0,0 +1,111 @@
...
+            check.taskSuccessful(":hotRunJvm")
+            check.logContains("Compose Hot Reload ($externalHotReloadVersion)")
+            check.logContains("Kotlin MPP app is running!")
+        }
+        // wait for java process to complete
+        Thread.sleep(1000)
```
terrakok: why is it required?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -0,0 +1,111 @@
...
+        }
+        modifyText("build.gradle") {
+            //  Apply hot reload plugin explicitly
+            it.replace(
+                "plugins {",
+                """
+                        plugins {
+                            id "org.jetbrains.compose.hot-reload"
+                """.trimIndent()
+            )
```
terrakok, line 115: I suspect this is not necessary. Could you check?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureHotReload.kt
```diff
@@ -0,0 +1,28 @@
...
+    if (!ComposeProperties.disableHotReload(providers).get()) {
+        plugins.withId(KOTLIN_MPP_PLUGIN_ID) {
+            mppExt.targets.withType(KotlinJvmTarget::class.java) { target ->
+                val runtimeFilesProvider = JvmApplicationRuntimeFilesProvider.FromKotlinMppTarget(target)
+                applyHotReload()
+            }
+        }
+        plugins.withId(KOTLIN_JVM_PLUGIN_ID) {
+            val sourceSet = project.javaSourceSets.getByName("main")
+            val runtimeFilesProvider = JvmApplicationRuntimeFilesProvider.FromGradleSourceSet(sourceSet)
```
terrakok: I haven't seen it as a requirement in the hot reload docs 🤔 
Why do we need it?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -0,0 +1,111 @@
...
+import org.jetbrains.compose.test.utils.GradlePluginTestBase
+import org.jetbrains.compose.test.utils.checks
+import org.junit.jupiter.api.Disabled
+import org.junit.jupiter.api.fail
+import org.junit.jupiter.api.Test
+import kotlin.concurrent.thread
+
+class HotReloadTest : GradlePluginTestBase() {
+    @Test
+    fun smokeTestHotRunTask() = with(testProject("application/jvm")) {
```
terrakok: rename to "checkHotReloadTaskRegisteredInJvmProject"

terrakok: add one more test: "checkHotReloadTaskRegisteredInKmpProject"

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -587,18 +587,9 @@ class ResourcesTest : GradlePluginTestBase() {
...
-                }    
-        """.trimIndent()
-        )
-
+        val disabaleProperty = ComposeProperties.DISABLE_RESOURCE_CONTENT_HASH_GENERATION
         //check generated resource's accessors
-        gradle("prepareKotlinIdeaImport").checks {
+        gradle("prepareKotlinIdeaImport", "-P$disabaleProperty=false").checks {
             val expected = if (System.getProperty("os.name").lowercase().contains("windows")) {
                 // Windows has different line endings in comparison with Unixes,
```
terrakok: ⚠️ ⚠️ ⚠️  BTW! Does the CHR plugin break reproducible builds between Win and Unix machines?

sellmair: > ⚠️ ⚠️ ⚠️ BTW! Does the CHR plugin break reproducible builds between Win and Unix machines?

Can you please elaborate?

sellmair: The only thing I could imagine is that different OS create different hashes for resources (which would be a bug?)

pjBooms: >The only thing I could imagine is that different OS create different hashes for resources (which would be a bug?)

Windows has different line endings in comparison with Unixes. So `vector.xml` vector drawables become binary different when are they checked out from Git and thus produce different hashes. 

Last time when we discussed it, we came into conclusion that the hashes must be different, because the content is different and fixed the test  by providing different "expected" directories for Windows and non-Windows. However "reproducible builds" sounds like a valid contra-argument so I would handle it in the hash calculation. "Cons" of this approach was that we had to add a custom platform-specific handling logic that sounds fragile.

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: I suspect this is a bug of `org.gradle.testkit.runner.Build` and `org.junit.jupiter.api` -- gradle execution does  not wait for subprocesses to complete. Currently still running application (spawned by gradle task)  blocks temp directory for the test and it cannot be deleted on Windows causing the test to fail. I would try to find a more elegant solution for this, but if you know how to properly address it, please let me know

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: See: https://github.com/JetBrains/compose-hot-reload?tab=readme-ov-file#run-tasks

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: It is necessary. Otherwise explicitly set version won't be applied and we will apply ours.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureHotReload.kt
pjBooms: Added

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: done

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/internal/configureHotReload.kt
pjBooms: Sorry, removed

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: done

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
pjBooms: Handled this [in the commit](https://github.com/JetBrains/compose-multiplatform/pull/5444/commits/1bd3f058567672e8736b2d14e6064178407854c1).

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: It seems it is fixed only in Gradle [9.1.0](https://github.com/gradle/gradle/issues/7603)

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -107,6 +88,21 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask(), Resour
         ).forEach { it.writeTo(kotlinDir) }
     }
 
+    private fun File.resourceContentHash(type: ResourceType): Int {
+        if (System.getProperty("os.name").lowercase().contains("windows") &&
+            (type == ResourceType.DRAWABLE) && path.endsWith(".xml", true)
```
terrakok: Why is it required only for xmls?
Isn't it a problem for strings with line breaks?

pjBooms: We have a multiline string resource for the test -- no problem encountered. I guess XML reader handles it. Else the same problem would manifest for many projects.

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
sellmair: Makes sense! I strongly agree that any text-based resource should produce the same hashCode regardless of the formatting used. The 'hash' here is not an indicator of the actual text hash, but a hash for its 'semantic'. It should not change if something meaningless changed (e.g., if the format of line-endings changed). 
However, I wonder how projects would end up with different XML files in their project? Is there some generation happening that produces different XML files (e.g., uses different line-endings on different platforms)?

.github/workflows/gradle-plugin.yml
pjBooms: >Have you tried to use two distributions here? distribution: 'jetbrains', 'corretto'

I have. It does not work.

>It would be good to ask them about the feature too

Created [an issue](https://github.com/actions/setup-java/issues/945).

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
terrakok: please, add the link to the issue in the code

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
terrakok: I think a more correct way is to apply `'\r\n' -> '\n'` conversion for **any** text files on **Windows**.

pjBooms: How do we know that a file is a text file? We should apply it only for the text files for which we generate accessors. Currently, it is only XML vector drawables.

terrakok: SVG?

pjBooms: Correct! Handled .svg too.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -107,6 +88,22 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask(), Resour
         ).forEach { it.writeTo(kotlinDir) }
     }
 
+    private fun File.resourceContentHash(type: ResourceType): Int {
+        if (System.getProperty("os.name").lowercase().contains("windows") &&
+            (type == ResourceType.DRAWABLE) &&
+            (path.endsWith(".xml", true) || path.endsWith(".svg", true))
+        ) {
```
terrakok: Why do you use in one case `lowercase` in another `ignoreCase = true`?
Why do we need check type of the resource?

Please, introduce functions: `fun hostIsWindows(): Boolean` and `fun isTextFile(): Boolean`

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -107,6 +88,22 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask(), Resour
...
+            // Windows has different line endings in comparison with Unixes,
+            // thus the XML-based resource files binary differ there, so we need to handle this.
+            return readText().replace("\r\n", "\n").toByteArray().contentHashCode()
+        } else {
+            // Once a new text resource file is introduced, we have to catch it and handle its line endings.
+            check(type == ResourceType.DRAWABLE || type == ResourceType.FONT) { "Cannot calculate content hash for $type resource type!" }
```
terrakok: why do you check font type?
why should binary PNGs throw the error?

terrakok (changes requested):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
terrakok: add all type of resources in the test case, please. to check stability of the hashes

pjBooms: Currently we generate accessors only for font and drawable files (strings and raw files are handled separately). Here I am just checking this fact, so if we introduce a new **text** resource type, we could catch it and recall that we may need to do something with its line endings. It is reflected in the comment.

pjBooms: >add all type of resources in the test case, please.

Do you mean add `.svg` and raster images?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
pjBooms: done

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
pjBooms: Added two more files

terrakok: I don't agree. If we add a new text based resource type with a new extension, it will fail tests because hashes will be different. We don't need resource type here. And we don't need to check `type == ResourceType.DRAWABLE` as well. for any file based xml/svg resource we must do conversion on win machines

gradle-plugins/compose/src/test/test-projects/misc/commonResources/expected-with-hash-windows/commonMainResourceAccessors/app/group/resources_test/generated/resources/Plurals0.commonMain.kt
terrakok, line 1: why was it deleted as well as Font0 and other?

pjBooms: I deleted the whole `expected-with-hash-windows` directory.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
pjBooms: Ok, done. 

(let's hope that we don't forget to add tests for new text files)

sellmair: But just to double-check:
Any project would just check in those text files using a common line-ending, right? Will the git checkout then provide different line endings on different platforms?

pjBooms: Correct

PR #5453, opened 2025-10-13
Deprecate org.jetbrains.compose.ui.tooling.preview.Preview

pjBooms:
Deprecate `org.jetbrains.compose.ui.tooling.preview.Preview` annotation in favor of `androidx.compose.ui.tooling.preview.Preview`

Fixes [CMP-4869](https://youtrack.jetbrains.com/issue/CMP-4869/Reduce-variety-of-Preview-annotations)

### Migration Notes - Multiple Platforms
- Deprecate `org.jetbrains.compose.ui.tooling.preview.Preview` annotation in favor of `androidx.compose.ui.tooling.preview.Preview` to reduce variety of Preview annotations

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
@@ -113,6 +113,10 @@ abstract class ComposePlugin : Plugin<Project> {
 
     object CommonComponentsDependencies {
         val resources = composeDependency("org.jetbrains.compose.components:components-resources")
+        @Deprecated(
+            "Use org.jetbrains.compose.ui:ui-tooling-preview module instead",
+            ReplaceWith("compose.preview")
+        )
         val uiToolingPreview = composeDependency("org.jetbrains.compose.components:components-ui-tooling-preview")
     }
 


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -113,6 +113,10 @@ abstract class ComposePlugin : Plugin<Project> {
 
     object CommonComponentsDependencies {
         val resources = composeDependency("org.jetbrains.compose.components:components-resources")
+        @Deprecated(
+            "Use org.jetbrains.compose.ui:ui-tooling-preview module instead",
+            ReplaceWith("compose.preview")
+        )
         val uiToolingPreview = composeDependency("org.jetbrains.compose.components:components-ui-tooling-preview")
```
terrakok, line 120: In this case, I'd avoid to use `ReplaceWith` because the message and the action is counterintuitive 🤯

pjBooms: It is just for users convinience. `compose.preview` is alias to `org.jetbrains.compose.ui:ui-tooling-preview`. Until we have the DSL at all, it seems we can just replace

PR #5457, opened 2025-10-16
Kotlin 2.2.20

eymar:
## Testing
N/A

## Release Notes
N/A

### DIFF
--- a/html/build.gradle.kts
+++ b/html/build.gradle.kts
@@ -183,7 +183,6 @@ subprojects {
 
     repositories {
         gradlePluginPortal()
-        mavenLocal()
         mavenCentral()
         maven {
             url = uri("https://maven.pkg.jetbrains.space/public/p/compose/dev")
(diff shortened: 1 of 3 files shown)


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
```diff
@@ -106,7 +106,7 @@ class GradlePluginTest : GradlePluginTestBase() {
     fun testOldestKotlinJsMpp() = with(
         testProject(
             "application/jsMpp",
-            testEnvironment = defaultTestEnvironment.copy(kotlinVersion = oldestSupportedKotlinVersion)
+            testEnvironment = defaultTestEnvironment.copy(kotlinVersion = "2.2.0-RC2")
```
MatkovIvan: Why it's not the constant anymore? Why it's RC and not stable version?
It will be tricky to find it during the next update

eymar: the PRs about 2.2.20 are rather a draft so far :)

I recreated them again after rebasing. And might've missed some things when solving the conflicts.

MatkovIvan (approved):

PR #5469, opened 2025-10-31
update test projects for gradle plugin to use direct dependencies

kropp:
Update test projects for Gradle plugin to use direct dependencies instead of aliases

Fixes [CMP-9183](https://youtrack.jetbrains.com/issue/CMP-9183)

## Testing
Existing tests should continue to pass

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/test/test-projects/misc/kmpResourcePublication/appModule/build.gradle.kts
+++ b/gradle-plugins/compose/src/test/test-projects/misc/kmpResourcePublication/appModule/build.gradle.kts
@@ -1,5 +1,3 @@
-import org.jetbrains.compose.ExperimentalComposeLibrary
-
 plugins {
     id("org.jetbrains.compose")
     kotlin("multiplatform")


gradle-plugins/compose/src/test/test-projects/misc/androidAppWithResources/featureModule/build.gradle.kts
```diff
@@ -6,29 +6,26 @@ plugins {
 }
 
 kotlin {
+    jvmToolchain(11)
```
MatkovIvan: It requires JDK 11 to be installed, but it's not the case in our pipeline anymore.
It might be replaced like [this](https://github.com/JetBrains/compose-multiplatform/pull/5424/commits/51a87ca55cd5bc7e8b9cc78b6c7c1531bda2d96a) if you need to specify compatibility, or just drop it if it doesn't matter

gradle-plugins/compose/src/test/test-projects/misc/androidAppWithResources/featureModule/build.gradle.kts
```diff
@@ -6,29 +6,26 @@ plugins {
...
     jvm()
 
     androidLibrary {
+        experimentalProperties["android.experimental.kmp.enableAndroidResources"] = true
         namespace = "me.sample.feature"
         compileSdk = 35
         minSdk = 23
-        androidResources.enable = true
```
MatkovIvan, line 15: I guess it shouldn't be reverted to the previous version, please check conflicts/rebase and revert unrelated changes

PR #5474, opened 2025-11-04
Bump minimal supported Kotlin version to 2.1.0

pjBooms:
Bump minimal supported Kotlin version to 2.1.0

Fixes [CMP-9218](https://youtrack.jetbrains.com/issue/CMP-9218/Bump-minimal-supported-Kotlin-version-to-2.1.0-in-CMP-Gradle-Plugin)

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
+++ b/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
@@ -327,7 +327,7 @@ class ResourcesTest : GradlePluginTestBase() {
     fun testAndroidAppWithResources() {
         //FIXME delete the filter when https://issuetracker.google.com/456657404 is fixed
         Assumptions.assumeFalse {
-            currentOS == OS.Windows && defaultTestEnvironment.gradleVersion.contains("9.0.0")
+            currentOS == OS.Windows && defaultTestEnvironment.agpVersion.contains("9.0.0")
         }
         with(testProject("misc/androidAppWithResources", defaultTestEnvironment)) {
             gradle(":appModule:assembleDebug").checks {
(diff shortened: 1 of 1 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/CheckComposeCompilerPlugin.kt
```diff
@@ -28,20 +28,20 @@ internal fun Project.checkComposeCompilerPlugin() {
...
+internal const val minimalSupportedKgpVersion = "2.1.0"
 internal const val minimalSupportedKgpVersionError = "e: Configuration problem: " +
-        "Minimal supported Kotlin Gradle Plugin version is 2.0.0"
-internal const val newCompilerIsAvailableVersion = "2.0.0"
+        "Minimal supported Kotlin Gradle Plugin version is $minimalSupportedKgpVersion"
 internal const val newComposeCompilerKotlinSupportPluginId = "org.jetbrains.kotlin.plugin.compose"
 internal const val newComposeCompilerError =
-    "Since Kotlin 2.0.0 to use Compose Multiplatform " +
-            "you must apply \"$newComposeCompilerKotlinSupportPluginId\" plugin." +
+    "Starting with Compose Multiplatform 1.6.10, " +
```
MatkovIvan, line 36: It depends on Kotlin + compiler plugin, not on Compose Multiplatform version

pjBooms: I have just copied what we say in [the doc](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-compiler.html#migrating-a-compose-multiplatform-project) that we reference here. 
As CMP does not support Kotlin versions < 2.0, it actually requires for the plugin to be applied.

kropp: Both statements are valid, aligning the error message with documentation is a good idea.

kropp (approved):
According to the [investigation](https://youtrack.jetbrains.com/issue/CMP-9205/Desktop-Android-Doesnt-work-with-kotlin-2.1.10-compose-1.10.0-beta01-due-to-bundling-hot-reload#focus=Comments-27-12937457.0-0) in a related issue, compilation fails with unreadable errors when using Kotlin 2.0 with Compose Multiplatform 1.10, so we should raise it here in the explicit check as well.

PR #5494, opened 2025-11-26
copy existing license from root directory to all examples

kropp:
fixes CMP-8884

## Release Notes
N/A


examples/chat/LICENSE.txt
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
igordmn, line 1: There is no LICENSE.txt in graphics-2d and jetsnack. Is it expected?

igordmn (approved):

PR #5500, opened 2025-12-16
Implement iOS Tests example app

ASalavei:
Implement iOS Tests app to be able to run a minimal set of tests on the all range of supported iOS versions.

screenshot

Fixes https://youtrack.jetbrains.com/issue/CMP-7993/ios-stable.-test-on-iOS-13-14

## Release Notes
N/A

### DIFF
--- a/compose/ios-manual-tests/iosApp/iosApp/main.swift
+++ b/compose/ios-manual-tests/iosApp/iosApp/main.swift
@@ -1,17 +1,6 @@
 /*
- * Copyright 2025 The Android Open Source Project
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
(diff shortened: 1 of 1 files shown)


Kpotko (approved):

compose/ios-manual-tests/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,7 @@
+distributionBase=GRADLE_USER_HOME
```
igordmn, line 1: Is this example for our internal testing, or to show users how to include tests?

If it is the first, we have to move it out of `examples`, for example to `compose/ios-manual-tests`

ASalavei: moved

examples/validateExamplesIos.sh
```diff
@@ -36,3 +36,4 @@ runGradle codeviewer
 runGradle imageviewer
 runGradle graphics-2d
 runGradle jetsnack
+runGradle ios-tests
```
igordmn: If we moved the project, and want to check it on CI (better to check), we need to define it in another file.

I suggest just to copy it to `compose/validateProjects.sh`, define only `ios-manual-tests`, and add a new CI step

ASalavei: Done. TC config must be updated then.

compose/ios-manual-tests/iosApp/iosApp/main.swift
```diff
@@ -0,0 +1,33 @@
+/*
```
igordmn, line 1: If it isn't copied from AOSP, better to use the JetBrains copyright

ASalavei: done

PR #5511, opened 2026-01-14
Remove JetSnack sample in favor of JetCaster

kropp:
JetSnack sample is outdated and hard to maintain because of ConstraintLayout
Recently, there was another sample app ported from Jetpack samples to Multiplatform
https://github.com/kotlin-hands-on/jetcaster-kmp-migration

We don't need many similar samples, so let's keep the newer one only.

Fixes [CMP-9404](https://youtrack.jetbrains.com/issue/CMP-9404) Remove Jetsnack sample

## Testing
N/A

## Release Notes
N/A


eymar (approved):
WDYT about keeping the examples/jetsnack/README.md in case there some links to the project?
The readme can simple have a link to the new example?.

bashor:
As I see it, unlike JetSnack, JetCaster doesn't have non-mobile targets. Is it going to be changed? Should we keep JetSnack until then?

kropp:
JetSnack in its current state isn't a good multiplatform sample, as it is very outdated. Updating it is not feasible, as original Android implementation has been changed significantly, like Material → Material3 transition and many internal things, and the non-Android part has been also significantly different from it because of `ConstraintLayout`.
That's beeing said the only viable option to have it up to date is to port it again from scratch, for which we don't have capacity now.

PR #5516, opened 2026-01-16
CMP-9545: Cache Storage API extra check for protocol 

Maxi937:
Compose Multiplatform v1.10.0 introduced the use of the Cache Storage API to avoid repeat http requests for resources on Web ([#5379](https://github.com/JetBrains/compose-multiplatform/pull/5379)).

In a VsCode Extension Environment, the protocol of a Web View Panel that gets used is "vscode-webview:". This does not work with the Cache Storage API and results in an error when trying to load any compose resources. This PR just ensures the protocol is http/https in order to use Cache Storage API on Web Target.

Fixes https://youtrack.jetbrains.com/issue/CMP-9545/ComposeResources-on-Web-not-loading-because-Cache-Storage-API-is-not-supported-everywhere

## Testing
**Describe how you tested your changes (provide a snippet or/and steps)**
- I ran my vscode extension with the change and without the change

**without**
image

**with**
image

- I ran a regular browserDevelopmentRun to ensure caching still worked as expected
image

### Fixes - Resources
Fixes an issue where web resources failed to load when calling the Cache Storage API with unsupported protocols (e.g., vscode-webview: in VS Code webviews).

### DIFF
--- a/.github/workflows/gradle-plugin.yml
+++ b/.github/workflows/gradle-plugin.yml
@@ -17,8 +17,8 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-24.04', 'macos-15-xlarge', 'windows-2022']
-        gradle: ['8.13', '9.2.0']
-        agp: ['8.11.2', '8.12.3', '9.0.0-alpha13']
+        gradle: ['8.13', '9.3.0']
+        agp: ['8.12.3', '9.0.0']
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout Repository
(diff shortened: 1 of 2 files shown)


components/resources/library/src/webMain/kotlin/org/jetbrains/compose/resources/ResourceWebCache.web.kt
```diff
@@ -37,10 +37,15 @@ internal object ResourceWebCache {
...
     private val resetMutex = Mutex()
 
-    private val supportsCacheApi: Boolean by lazy { supportsCacheApi() }
+    private val supportsCacheApi: Boolean by lazy {
+        supportsCacheApi() && isCacheableProtocol()
+    }
 
     suspend fun load(path: String, onNoCacheHit: suspend (path: String) -> Response): Response {
-        if (!supportsCacheApi) return onNoCacheHit(path)
+        if (!supportsCacheApi) {
```
eymar: Hello! The change at this line seems to be unrelated to the fix. Could you please revert it here? This helps us when we use git history / git blame.

PR #5524, opened 2026-01-29
Migrate dev maven

kropp:
Change Dev Maven repository to public.jetbrains.team

Fixes [CMP-1460](https://youtrack.jetbrains.com/issue/CMP-1460) Migrate Compose dev builds from public.jetbrains.space

## Testing
All build configurations are green, packages are uploaded.

## Release Notes
N/A

### DIFF
--- a/benchmarks/multiplatform/benchmarks/build.gradle.kts
+++ b/benchmarks/multiplatform/benchmarks/build.gradle.kts
@@ -47,8 +47,6 @@ kotlin {
     @OptIn(ExperimentalWasmDsl::class)
     wasmJs {
         binaries.executable()
-        // todo: restore d8 benchmark after Compose update
-/*
         d8 {
             compilerOptions.freeCompilerArgs.add("-Xwasm-attach-js-exception")
             runTask {
@@ -56,7 +54,6 @@ kotlin {
                 // d8Args.add("--abort-on-uncaught-exception")
             }
         }
-*/
         browser()
 
         binaries.configureEach {
--- a/.github/workflows/benchmarks-smoke.yml
+++ b/.github/workflows/benchmarks-smoke.yml
@@ -41,6 +41,8 @@ jobs:
         ./gradlew :benchmarks:run
 
   benchmarks-wasm-d8:
+    # temporarily disabled after Compose update
+    if: false
     runs-on: ubuntu-24.04
     steps:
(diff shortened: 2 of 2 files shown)


kropp:
Related PR: https://jetbrains.team/p/ui/reviews/99/timeline
Build from branch https://teamcity.jetbrains.com/buildConfiguration/JetBrainsPublicProjects_Compose_AllPersonalBuild/5835949?buildTab=dependencies&focusLine=NaN&type=snapshot&mode=chain&groupAllProjects=false&groupAllComposite=false
Repo https://jetbrains.team/p/cmp/packages/maven/dev

benchmarks/multiplatform/benchmarks/build.gradle.kts
```diff
@@ -47,13 +47,16 @@ kotlin {
     @OptIn(ExperimentalWasmDsl::class)
     wasmJs {
         binaries.executable()
+        // todo: restore d8 benchmark after Compose update
```
eymar: To make CI pass, the job for d8 should be disabled too.

kropp: Thanks for the hint, I restored the target in the project and instead disabled the workflow.

benchmarks/multiplatform/settings.gradle.kts
```diff
@@ -3,7 +3,7 @@ pluginManagement {
         mavenLocal()
         mavenCentral()
         gradlePluginPortal()
-        maven("https://maven.pkg.jetbrains.space/public/p/compose/dev")
+        maven("https://packages.jetbrains.team/maven/p/cmp/dev")
```
igordmn, line 6: Can we setup an URL similar to Kotlin in the future (maybe not now). Something like:
```
https://redirector.kotlinlang.org/maven/cmp-dev/
```
This repo can be used by external users to use daily builds. `https://packages.jetbrains.team` looks like an internal link. Use cases:
- users don't want to wait for 2-weakly builds to try a just committed change
- we can ask users to try some build to investigate an issue or try a fix

kropp: Yes, that's the plan.
Unfortunately, there is already such a redirect to our old dev repo, so to not introduce further confusion I decide to migrate first, update redirect and then use it in our samples

PR #5529, opened 2026-02-04
Introduce real mode for benchmarks running

pjBooms:
In `real` mode benchmarks are run physically on the underlying target screen reporting fps metric.

Fixes [CMP-9696](https://youtrack.jetbrains.com/issue/CMP-9696)

## Testing
Manually

## Release Notes
N/A

### DIFF
--- a/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
+++ b/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
@@ -338,7 +338,7 @@ suspend fun runBenchmarks(
 @Composable
 fun BenchmarkRunner(
     benchmarks: List<Benchmark>,
-    frameRate: Int,
+    deviceFrameRate: Int,
     onExit: () -> Unit
 ) {
     var currentBenchmarkIndex by remember { mutableStateOf(0) }
@@ -364,7 +364,7 @@ fun BenchmarkRunner(
                 val duration = start.elapsedNow()
                 val stats = BenchmarkResult(
                     name = benchmark.name,
-                    frameBudget = (1.seconds.inWholeNanoseconds / frameRate).nanoseconds,
+                    frameBudget = (1.seconds.inWholeNanoseconds / deviceFrameRate).nanoseconds,
                     conditions = BenchmarkConditions(benchmark.frameCount, 0),
                     averageFrameInfo = FrameInfo(duration / benchmark.frameCount, Duration.ZERO),
                     averageFPSInfo = FPSInfo(benchmark.frameCount.toDouble() / duration.toDouble(DurationUnit.SECONDS)),
--- a/benchmarks/multiplatform/benchmarks/src/macosMain/kotlin/main.macos.kt
+++ b/benchmarks/multiplatform/benchmarks/src/macosMain/kotlin/main.macos.kt
@@ -8,13 +8,15 @@ import androidx.compose.ui.window.Window
 import kotlinx.coroutines.runBlocking
 import platform.AppKit.NSApp
 import platform.AppKit.NSApplication
+import platform.AppKit.NSScreen
+import platform.AppKit.maximumFramesPerSecond
 import kotlin.system.exitProcess
 
 fun main(args : Array<String>) {
     Config.setGlobalFromArgs(args)
     if (Config.isModeEnabled(Mode.REAL)) {
         NSApplication.sharedApplication()
-        val frameRate = 120 // TODO: get from device
+        val frameRate = (NSScreen.mainScreen?.maximumFramesPerSecond?.toInt()) ?: 120
         Window(
             "Benchmarks",
             DpSize(


benchmarks/multiplatform/benchmarks/src/macosMain/kotlin/main.macos.kt
```diff
@@ -2,9 +2,29 @@
...
+import platform.AppKit.NSApplication
+import kotlin.system.exitProcess
 
 fun main(args : Array<String>) {
     Config.setGlobalFromArgs(args)
-    runBlocking { runBenchmarks(graphicsContext = graphicsContext()) }
-}
+    if (Config.isModeEnabled(Mode.REAL)) {
+        NSApplication.sharedApplication()
+        val frameRate = 120 // TODO: get from device
```
ASalavei: NSScreen.mainScreen.maximumFramesPerSecond

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
```diff
@@ -262,21 +330,66 @@ suspend fun runBenchmarks(
...
+                }
+                repeat(benchmark.frameCount) {
+                    val frameStart = TimeSource.Monotonic.markNow()
+                    withFrameNanos { }
+                    frames[it] = BenchmarkFrame(frameStart.elapsedNow(), Duration.ZERO)
+                }
+                val duration = start.elapsedNow()
+                val stats = BenchmarkResult(
+                    name = benchmark.name,
+                    frameBudget = (1.seconds.inWholeNanoseconds / frameRate).nanoseconds,
```
eymar: Looks like `frameRate` is used only to calculate `frameBudget`.

What do you think about passing the correct `frameBudget` value in parameters and remove the calculation here?

It would make the function a bit more clear - when I saw the `frameRate` parameter first I thought the the BenchmarkRunner tries to enforce a certain frameRate.

benchmarks/multiplatform/benchmarks/src/wasmJsMain/kotlin/main.wasmJs.kt
```diff
@@ -15,23 +18,34 @@ fun main(args: Array<String>) {
...
 
     Config.setGlobalFromArgs(args)
 
-    MainScope().launch {
-        if (Config.saveStats() && !BenchmarksSaveServerClient.isServerAlive()) {
-            println("No benchmark server found.")
-            return@launch
+    val composeRoot = document.getElementById("root")!!
+    if (Config.isModeEnabled(Mode.REAL)) {
+        val frameRate = 120 // can we get this from device?
```
eymar, line 31: Unfortunately, there is no direct API.
But we can get it by running `requestAnimationFrame` for 1-2 seconds - https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame and then take the average.

ASalavei (approved):
LGTM!
I tried it on a real device and some tests, such as Visual Effects, showed slight improvements for parallel rendering.

After running tests on iOS, I would suggest several improvements:
- Write the total summary for all tests at the end, including the initial config. Currently it's hard to collect all the data because it can be interrupted with other log messages.
- Write number of total vs dropped frames for every test
- Write Duration for every text
- Add ability to configure warmup, and also it's good to have at least 2-3 seconds pause between tests and also after warmup to make easy separate tests in instruments.

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
pjBooms: I would like to avoid the same calculation for all targets. What about renaming it to `deviceFrameRate`?

benchmarks/multiplatform/benchmarks/src/wasmJsMain/kotlin/main.wasmJs.kt
pjBooms: I guess, we can postpone this, until we really need it

benchmarks/multiplatform/benchmarks/src/macosMain/kotlin/main.macos.kt
pjBooms: thanx, done

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
eymar: sounds good

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
```diff
@@ -262,21 +330,66 @@ suspend fun runBenchmarks(
...
+
+            LaunchedEffect(benchmark.name) {
+                val start = TimeSource.Monotonic.markNow()
+                val frames = MutableList(benchmark.frameCount) {
+                    BenchmarkFrame(Duration.ZERO, Duration.ZERO)
+                }
+                repeat(benchmark.frameCount) {
+                    val frameStart = TimeSource.Monotonic.markNow()
+                    withFrameNanos { }
+                    frames[it] = BenchmarkFrame(frameStart.elapsedNow(), Duration.ZERO)
```
igordmn, line 362: It looks like `frames[0]` will be near zero? Because it measures waiting for the next frame, not the frame itself.

It might skew the results.

To fix this, we can add additional `withFrameNanos { }` in the beginning of `LaunchedEffect`

benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/main.desktop.kt
```diff
@@ -3,15 +3,35 @@
...
 
 fun main(args: Array<String>) {
     Config.setGlobalFromArgs(args)
 
     if (Config.runServer) {
         // Start the benchmark server to receive results from browsers
         BenchmarksSaveServer.start()
+    } else if (Config.isModeEnabled(Mode.REAL)) {
+        val device = GraphicsEnvironment.getLocalGraphicsEnvironment().defaultScreenDevice
+        val frameRate = device.displayMode.refreshRate.takeIf { it > 0 } ?: 120
```
igordmn, line 22: Is it planned to be run only manually or/and on the server as one more indicator without replacing non-real benchmarks?

Asking because running on real devices is limited by vsync. And to know improvement we should run something heavy that always doesn't reach the vsync limit.

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
```diff
@@ -262,21 +330,66 @@ suspend fun runBenchmarks(
...
+    if (Config.isBenchmarkEnabled(benchmark.name)) {
+        Box(modifier = Modifier.fillMaxSize()) {
+            benchmark.content()
+
+            LaunchedEffect(benchmark.name) {
+                val start = TimeSource.Monotonic.markNow()
+                val frames = MutableList(benchmark.frameCount) {
+                    BenchmarkFrame(Duration.ZERO, Duration.ZERO)
+                }
+                repeat(benchmark.frameCount) {
```
igordmn, line 359: Should we skip `warmupCount` frames as in the "non-real" mode? (It also mentioned by Andrei)

benchmarks/multiplatform/benchmarks/src/desktopMain/kotlin/main.desktop.kt
pjBooms: For now, `real` mode is going to be used manually only for testing iOS performance improvements. 
Adding on CI  `real` mode is discussable but  there is no plans to replace `non-real` benchmarks.

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
pjBooms: I will add it in a subsequent PR.

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
pjBooms: I will test it, thanx

PR #5533, opened 2026-02-11
Benchmarks real mode improvements

pjBooms:
Improvements made by [this review comment](https://github.com/JetBrains/compose-multiplatform/pull/5529#pullrequestreview-3751067220):

- add warmup to real mode.
- add empty screen between benchmarks
- make warmupCount, frameCount, emptyScreenDelay configurable.
- introduce reportAtTheEnd config parameter
- report missed frames, average frame time for real mode
- normalize frame time for real mode by rounding its raw time to the nearest multiple of frame budget

Fixes [CMP-9696](https://youtrack.jetbrains.com/issue/CMP-9696)

## Testing
Manually

## Release Notes
N/A

### DIFF
--- a/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
+++ b/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
@@ -381,12 +381,12 @@ fun BenchmarkRunner(
                     withFrameNanos {  }
                     isEmptyScreen = false
                 } else {
-                    val start = TimeSource.Monotonic.markNow()
                     val frames = MutableList(benchmark.frameCount) {
                         BenchmarkFrame(Duration.ZERO, Duration.ZERO)
                     }
                     // skip first frame waiting
                     withFrameNanos {  }
+                    val start = TimeSource.Monotonic.markNow()
                     repeat(benchmark.frameCount) {
                         val frameStart = TimeSource.Monotonic.markNow()
                         withFrameNanos { }


benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Config.kt
```diff
@@ -58,9 +62,21 @@ object Args {
...
+            } else if (arg.startsWith("reportAtTheEnd=", ignoreCase = true)) {
+                reportAtTheEnd = arg.substringAfter("=").toBoolean()
+            } else {
+                println("WARNING: unknown argument $arg")
             }
         }
 
+        val defaultWarmupCount = if (modes.contains(Mode.REAL)) 0 else 100
+
         return Config(
```
ASalavei, line 80: it seems useful also to print config when running tests

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
```diff
@@ -342,39 +350,76 @@ fun BenchmarkRunner(
...
+                    repeat(benchmark.frameCount) {
+                        val frameStart = TimeSource.Monotonic.markNow()
+                        withFrameNanos { }
+                        val rawFrameTime = frameStart.elapsedNow()
+                        // rawFrameTime isn't reliable for missed-frame checks: small timing inaccuracies can push
+                        // it just over the frame budget even when the frame would effectively fit.
+                        // Instead, estimate how many vsync intervals the frame took by rounding the measured time to
+                        // the nearest multiple of nanosPerFrame.
+                        // Note: this is only an estimate — we can’t determine here whether a frame was truly missed.
+                        val normalizedFrameTime = ((rawFrameTime.inWholeNanoseconds + nanosPerFrame/2) / nanosPerFrame) * nanosPerFrame
```
igordmn, line 399: This also affects percentiles (and could affects FPS itself if we wouldn't use `averageFrameInfo`). This might give us a wrong value in cases:
- if the frame is 23ms, and we round it to 16ms (we miss a lot of accuracy)
- when we hardcode the refresh rate, not reading it from the current display
- the actual refresh rate isn't always strictly 60hz even if it returns 60.0 (it can be 59.98 and skew over time)

If we have an issue with missed frames, we can only fix its logic. What if we just multiply `frameBudget` by 1.5?

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
```diff
@@ -342,39 +350,76 @@ fun BenchmarkRunner(
...
+                    isEmptyScreen = false
+                } else {
+                    val start = TimeSource.Monotonic.markNow()
+                    val frames = MutableList(benchmark.frameCount) {
+                        BenchmarkFrame(Duration.ZERO, Duration.ZERO)
+                    }
+                    // skip first frame waiting
+                    withFrameNanos {  }
+                    repeat(benchmark.frameCount) {
+                        val frameStart = TimeSource.Monotonic.markNow()
```
igordmn, line 391: We should move it after `withFrameNanos { }`, as it can still wait for vsync in some cases (when we start another benchmark in the same window)

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Benchmarks.kt
igordmn: Discussed online:

> if the frame is 23ms, and we round it to 16ms (we miss a lot of accuracy)

> the actual refresh rate isn't always strictly 60hz even if it returns 60.0

Doesn't matter, as the average still be 16.7, and individual values only matter for missed frames.

> when we hardcode the refresh rate, not reading it from the current display

This is an issue, but we can look at it when needed (currently we hardcode only on Web)

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/Config.kt
pjBooms: Please provide PR on this

PR #5536, opened 2026-02-13
Update replaceVersion.sh to include components build files

eymar:
Also, update components (resources and other libs) to CMP 1.10.0 and Kotlin 2.3.0 (it has not been updated before because replaceVersion.sh missed it).

Other sub-projects already use CMP 1.10.0 and Kotlin 2.3.0

## Testing
N/A

## Release Notes
N/A

### DIFF
--- a/tools/replaceVersion.sh
+++ b/tools/replaceVersion.sh
@@ -11,6 +11,7 @@ declare -a folders=(
     "$ROOT/ci"
     "$ROOT/tutorials"
     "$ROOT/html"
+    "$ROOT/components"
 )
 
 if [ ! -z "$COMPOSE_TEMPLATES_FOLDER" ]; then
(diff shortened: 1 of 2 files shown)


tools/replaceVersion.sh
```diff
@@ -11,6 +11,7 @@ declare -a folders=(
     "$ROOT/ci"
     "$ROOT/tutorials"
     "$ROOT/html"
+    "$ROOT/components"
```
igordmn, line 14: I remember why I didn't include it in `replaceVersion.sh` in the first place. The intent was to update public projects with the latest version every release, but keep the publishing projects with a lesser version for compatibility.

Every release we should call this script, and if the latest Kotlin is 2.4.0, we should use it. But we should update publishing projects to Kotlin 2.4.0 only when we decide that we plan to break compatibility (it is usually some AOSP merge).

To keep this intent, we should remove
```
    "$ROOT/html"
    "$ROOT/components"
```
from the script, and either continue to change it manually, or add another script `replacePublishingKotlinVersion.sh`

PR #5538, opened 2026-02-16
Update oldest supported Kotlin version to 2.3.0 in GradlePluginTest

eymar:
Because we bump Kotlin version in CMP core and in components.

## Testing
N/A

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
+++ b/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
@@ -86,7 +86,7 @@ class GradlePluginTest : GradlePluginTestBase() {
             }
         }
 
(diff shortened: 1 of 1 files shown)


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
```diff
@@ -87,7 +87,7 @@ class GradlePluginTest : GradlePluginTestBase() {
         }
 
     // Note: we can't test non-jvm targets with Kotlin older than 2.2.0, because of klib abi version bump in 2.2.0
```
MatkovIvan: This comment needs to be updated

eymar: right :)

eymar: fixed

PR #5485, opened 2025-11-18
Add runtime libraries compatibility check for Compose projects

terrakok:
Add runtime libraries compatibility check for Compose projects

Fixes [CMP-9288](https://youtrack.jetbrains.com/issue/CMP-9288) Check compose libraries compatibility

image

To disable the check there is a new gradle property: `org.jetbrains.compose.library.compatibility.check.disable`

## Testing
- Added integration tests to verify version mismatches and override behavior.

### Features - Gradle Plugin
- Add a compatibility check for runtime libraries to ensure consistency with the expected Compose version.

PR #5541, opened 2026-02-20
Add runtime libraries compatibility check for skiko

MatkovIvan:
[CMP-9799](https://youtrack.jetbrains.com/issue/CMP-9799) Check skiko library compatibility
Addition to #5485

### Features - Gradle Plugin
- Add a compatibility check for skiko libraries to ensure consistency

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/RuntimeLibrariesCompatibilityCheck.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/RuntimeLibrariesCompatibilityCheck.kt
@@ -2,8 +2,15 @@ package org.jetbrains.compose
 
 import org.gradle.api.DefaultTask
 import org.gradle.api.Project
+import org.gradle.api.artifacts.ModuleVersionIdentifier
+import org.gradle.api.artifacts.component.ComponentIdentifier
+import org.gradle.api.artifacts.component.ComponentSelector
+import org.gradle.api.artifacts.component.ModuleComponentIdentifier
 import org.gradle.api.artifacts.component.ModuleComponentSelector
+import org.gradle.api.artifacts.component.ProjectComponentIdentifier
+import org.gradle.api.artifacts.component.ProjectComponentSelector
 import org.gradle.api.artifacts.result.ResolvedDependencyResult
+import org.gradle.api.artifacts.result.ResolvedComponentResult
 import org.gradle.api.provider.Property
 import org.gradle.api.provider.ProviderFactory
 import org.gradle.api.provider.SetProperty
@@ -72,7 +79,7 @@ internal abstract class RuntimeLibrariesCompatibilityCheck : DefaultTask() {
             "org.jetbrains.compose.foundation:foundation",
             "org.jetbrains.compose.ui:ui"
         )
-        const val skikoGroup = "org.jetbrains.skiko"
+        val skikoLibraryForCheck = "org.jetbrains.skiko:skiko"
 
         private val majorMinorRegex = """^(\d+)\.(\d+)""".toRegex()
         fun majorMinorVersion(version: String): String {
@@ -81,13 +88,6 @@ internal abstract class RuntimeLibrariesCompatibilityCheck : DefaultTask() {
             val minor = match.groupValues[2]
             return "$major.$minor"
         }
-
-        fun hasIncompatibleVersions(usages: List<DependencyUsage>): Boolean =
-            usages
-                .flatMap { usage -> listOfNotNull(usage.requestedVersion, usage.selectedVersion) }
-                .map(::majorMinorVersion)
-                .distinct()
-                .size > 1
     }
 
     @get:Inject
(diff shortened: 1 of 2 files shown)


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/unit/SkikoRuntimeLibrariesCompatibilityCheckTest.kt
```diff
@@ -0,0 +1,35 @@
+package org.jetbrains.compose.test.tests.unit
+
+import org.jetbrains.compose.RuntimeLibrariesCompatibilityCheck
+import org.jetbrains.compose.RuntimeLibrariesCompatibilityCheck.Companion.skikoLibraryForCheck
+import org.jetbrains.compose.RuntimeLibrariesCompatibilityCheck.ProblemLibrary
+import kotlin.test.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class SkikoRuntimeLibrariesCompatibilityCheckTest {
```
igordmn: We need to add a test for the case:
1. a project depends on current Compose UI version
2. a project depends on some old third-party library that depends on an old Compose

We have different Skiko here, but transitively via different Compose versions, which should be resolved into single current Skiko version.

We shouldn't print a warning in this case, as it is a valid case.

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/unit/SkikoRuntimeLibrariesCompatibilityCheckTest.kt
```diff
@@ -0,0 +1,35 @@
...
+    @Test
+    fun patchVersionDifferencesAreIgnored() {
```
igordmn: These tests doesn't look helpful enough, they just check a small List check function, which probably doesn't contain bugs, but doesn't check the retrieving libraries logic, which could contain bugs.

We need to add tests on the project level.

igordmn: After we do that, we need to be sure that we test this case:
1. a project depends on current Compose UI version
2. a project depends on some old third-party library that depends on an old Compose

We have different Skiko here, but transitively via different Compose versions, which should be resolved into single current Skiko version.

We shouldn't print a warning in this case, as it is a valid case.

MatkovIvan: I guess it should work, but need to re-check, right

MatkovIvan: They already triggered locally that `Version` class cannot be re-used here because it doesn't support 4 numbers 😅 

For project level I also add a simple test (not full suite)

MatkovIvan: Done

MatkovIvan: Done

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/RuntimeLibrariesCompatibilityCheck.kt
```diff
@@ -92,47 +114,106 @@ internal abstract class RuntimeLibrariesCompatibilityCheck : DefaultTask() {
...
+                    projectPath.get(),
+                    configurationName.get(),
+                    composeProblems,
+                    expectedRuntimeVersion
+                )
+            )
         }
 
-        if (problems.isNotEmpty()) {
+        val dependencyUsages = allDependencies.get().mapNotNull { dep ->
```
igordmn:
```suggestion
        val skikoDependencyUsages = allDependencies.get().mapNotNull { dep ->
```

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/RuntimeLibrariesCompatibilityCheck.kt
```diff
@@ -92,47 +114,106 @@ internal abstract class RuntimeLibrariesCompatibilityCheck : DefaultTask() {
...
+                val requested = usage.requestedName?.let { "$it:${usage.requestedVersion}" } ?: "<unknown>"
+                val selected = usage.selectedName?.let { "$it:${usage.selectedVersion}" } ?: "<unknown>"
+                appendLine("    requested: '$requested'")
+                appendLine("    selected:  '$selected'")
+                appendLine()
+            }
+        appendNoteAboutDependencyMismatch(projectName, configurationName)
+        appendLine()
+        appendLine("Note: Skiko is considered implementation detail in Compose Multiplatform and might be incompatible across versions.")
+        appendLine("Please align Skiko dependencies to the same version. If possible, avoid direct skiko references and use Compose APIs instead.")
```
igordmn:
```suggestion
        appendLine("Please align Skiko dependencies to the same version. If possible, avoid direct skiko references and use Compose APIs instead.")
        appendLine("If the different version is used by a third-party library, ask the library author to update it using the latest Compose.")
```

Can we also write dependencies that use a wrong Skiko? Or at least a command that can show that?

Otherwise in most cases, in is not informative for app developers.

MatkovIvan: We're writing detailed instruction about how to find it via `dependencies` gradle task

igordmn: I see `appendNoteAboutDependencyMismatch`, indeed.

igordmn (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/RuntimeLibrariesCompatibilityCheck.kt
MatkovIvan: Refactored this part, so this comment is outdated

PR #5570, opened 2026-03-30
[CMP-7202] Introduce Unified Multiplatform Pointer Click Modifier

gagik894:
## Summary
Introduces `Modifier.onPointerClick` in `org.jetbrains.compose.components.input`, providing a unified multiplatform API for handling pointer interactions (mouse buttons, stylus, keyboard modifiers) while preserving touch semantics.

Resolves: [CMP-7202](https://youtrack.jetbrains.com/issue/CMP-7202)

---

## Problem

- **Metadata Loss:** `Modifier.clickable` does not expose hardware details (button type, modifiers, pointer type).
- **Inefficient Workarounds:** Desktop/Web solutions rely on stacking multiple modifiers, increasing complexity and risk of inconsistent behavior.

---

## Solution

`Modifier.onPointerClick` provides a single entry point for pointer-aware interactions via a structured `PointerClickEvent`.

### Event Payload
- `position`: Click position relative to component
- `buttons`: PointerButtons (nullable for accessibility/synthesized events)
- `type`: PointerType (Mouse, Touch, Stylus, etc.)
- `keyboardModifiers`: Shift/Ctrl/Alt/Meta state
- Helpers: `isPrimaryAction`, `isSecondaryAction`, `isEraser`

---

## Implementation

- Built as a `DelegatingNode` using pointer + semantics nodes
- Processes events across pointer passes for correct cancellation behavior
- Safely integrates keyboard modifiers and accessibility-triggered clicks
- Decouples click handling from visual indication (primary actions trigger ripple by default)

---

## Testing

- Covers button routing, modifier propagation, cancellation, and bounds handling
- Includes interactive demo for inspection and validation


kropp:
Hi @gagik894 
unfortunately, we can not accept this PR:
* `components` is a wrong module, this modifier should be placed near other similar ones
* all common APIs should be added directly to AOSP (androidx) repo, otherwise it won't be available on Android. And androidx repo is the source of truth in this case.

PR #5582, opened 2026-04-20
[Resources] Support BCP 47 script code qualifiers in resource folders

### LINKED ISSUE #4449: Ability to support more language and region qualifiers
GenerateResClassTask was failed:
java.lang.IllegalStateException: values-b+zh+Hans/strings.xml contains unknown qualifier: 'b+zh+Hans'.

Can values-b+zh+Hans and values-b+zh+Hant be supported

https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-images-resources.html#language-and-regional-qualifiers
https://developer.android.google.cn/guide/topics/resources/app-languages

2KAbhishek:
Resource folder parsing rejected valid Android BCP 47 folder names like `values-b+sr+Latn` (script code), `values-b+zh+Hans+CN` (multi-segment), and `values-b+es+419` (numeric region) with "unknown qualifier" errors when placed under `commonMain/composeResources`.
The Gradle plugin split folder names on `-` without understanding the `b+lang+code` segment format, and qualifier validation only accepted 2-letter region codes.

This PR adds support for BCP 47 script codes (ISO 15924) and numeric region codes in resource folder qualifiers, propagates the script through both the suspend (`getString()`) and Composable (`stringResource()`) lookup paths.

Fixes [CMP-4449](https://youtrack.jetbrains.com/issue/CMP-4449)
Fixes #4449

## Testing

- `testGetPathByEnvironmentWithScript` in `ResourceTest` covers script matching (`sr-Latn`, `sr-Cyrl`, `zh-Hans`, `zh-Hant`), region narrowing, empty-script fallback, cross-language default resolution, and the script-priority guard.
- `testDefaultComposeEnvironmentPropagatesSystemScript` and `testDefaultComposeEnvironmentDropsSystemScriptOnLanguageOverride` cover the Composable `stringResource()` script lookup path.
- `testBcpFolderQualifiers` and `testBcpFolderQualifiersInvalid` integration tests cover language-only, script, alpha-2 and numeric regions, multi-segment chains, and trailing theme/density qualifiers, plus rejection of malformed segments, out-of-order subtags, trailing locale qualifiers after `b+...`, and standalone script qualifiers without `b+`.
- Verified end-to-end on Android emulator (API 36) with a sample app

### Fixes - Resources

- Support BCP 47 script (`values-b+sr+Latn`, `values-b+zh+Hans`) and numeric region (`values-b+es+419`) qualifiers, including multi-segment locales and trailing theme/density qualifiers, in `commonMain` resource folders

> Authored with AI assistance, reviewed and tested manually before submitting.


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
```diff
@@ -166,35 +170,46 @@ private fun List<ResourceItem>.filterByDensity(density: DensityQualifier): List<
...
+        val noScript = withLanguage.filter { item -> item.qualifiers.none { it is ScriptQualifier } }
+        noScript.ifEmpty { withLanguage }
+    } else {
+        val withScript = withLanguage.filter { item -> item.qualifiers.any { it == script } }
+        if (withScript.isNotEmpty()) {
+            withScript
+        } else {
+            withLanguage.filter { item -> item.qualifiers.none { it is ScriptQualifier } }
+        }
     }
```
terrakok, line 206: why do you use different code styles here? (suspected AI)

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
```diff
@@ -166,35 +170,46 @@ private fun List<ResourceItem>.filterByDensity(density: DensityQualifier): List<
...
+
     val withLanguage = filter { item ->
         item.qualifiers.any { it == language }
     }
-
-    val withExactLocale = withLanguage.filter { item ->
-        item.qualifiers.any { it == region }
+    if (withLanguage.isEmpty()) return noLocaleItems
+
+    val scriptCandidates = if (script.script.isEmpty()) {
```
terrakok, line 196: instead of `script.script.isEmpty()` better to add a function to the `ScriptQualifier`.

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/ResourceTest.kt
```diff
@@ -111,4 +112,64 @@ class ResourceTest {
             assertEquals("Resource with ID='ImageResource:test3' has more than one file: en1, en2", msg)
         }
     }
+
+    @Test
+    fun testGetPathByEnvironmentWithScript() {
```
terrakok, line 117: please check all cases from the description:
```
// Filter by language, script, and region together (extended from the original lang+region logic):
// 1) exact language + script + region -> use it
// 2) language + script (no region) -> use it
// 3) language + region (no script) -> use it
// 4) language only (no script, no region) -> use it
// 5) items with NO locale qualifiers at all (default)
// When the environment script is empty (e.g. DefaultComposeEnvironment), prefer items without
// a ScriptQualifier first; fall back to script-tagged items only if nothing else matches.
```

and add references on the cases in comments

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -141,6 +139,34 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask() {
...
     }
 
+    /**
+     * Parses Android folder names to extract qualifiers.
+     * Handles both standard qualifiers and Android BCPF format.
+     *
+     * Examples: values-en-rUS -> ["en", "rUS"]
+     *           values-b+sr+Latn -> ["sr", "Latn"]
+     */
+    private fun parseAndroidFolderName(dirName: String): List<String>? {
```
terrakok, line 149: parseComposeResourceLocaleQualifiers

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -141,6 +139,34 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask() {
...
+        val bcpfPattern = Regex("""^([a-z]+)-b\+([a-z]{2,3})\+([A-Za-z0-9]+)$""")
```
terrakok, line 150: should it work for `values-dark-b+zh+Hant`?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -141,6 +139,34 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask() {
...
+        val bcpfMatch = bcpfPattern.matchEntire(dirName)
+        if (bcpfMatch != null) {
+            val language = bcpfMatch.groupValues[2]
+            val code = bcpfMatch.groupValues[3]
+            return if (code.first().isDigit()) {
+                listOf(language, "r$code")
+            } else {
+                listOf(language, code)
+            }
```
terrakok, line 159: could you explain the logic?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
terrakok: or when `b+sr+Latn+RS`

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -48,6 +48,25 @@ class ResourcesTest : GradlePluginTestBase() {
         }
     }
 
+    @Test
+    fun testBcpFolderQualifiers() {
```
terrakok, line 52: add more test cases:
 - more than 2 bcp47 tags -> error?
 - mix bcp47 and region -> error
 - more different bcp47 cases (only lang, numeric region, etc)

terrakok (changes requested):

2KAbhishek:
@terrakok thank you for the review comments, I will be resolving those and get back

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
2KAbhishek: Yes, I did use AI's help for this PR

I will be fixing the style mismatch, thanks for the catch

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
2KAbhishek: On this PR I have only added support for 2 separator resource like `b+sr+Latn`, was hoping to build on this once it gets merged and then add more - like the cases you mentioned.

What do you suggest?

components/resources/library/src/commonTest/kotlin/org/jetbrains/compose/resources/ResourceTest.kt
2KAbhishek: I will add these, thank you

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
2KAbhishek: I added this to support numeric region locales like `es-419`.

When parsing values-b+... folders we extract the language + next subtag, if that subtag is numeric (419) it represents a region, and Android resource qualifiers require the r prefix (values-es-r419).

So this normalizes 419 -> r419

> Note: the region language fallback is not fully baked in on this PR, i.e es-MX -> es-419 is not part of these changes, I am hoping to add that as a subsequent PR.

Please let me know if that sounds ok to you, or if you want me to clean this up.

terrakok: I mean why didn't you add r when it is letters?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
terrakok: cases with other qualifiers must work. like `values-dark-b+zh+Hant`

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
terrakok: then you must mention it at least, and review your code twice before submitting PR

westnordost:
In one comment you write you also added support for [**UN M49 region codes**](https://en.wikipedia.org/wiki/UN_M49).
From a multiplatform perspective, all you have in order to decide which resource to use is the `androidx.compose.ui.text.intl.Locale`, i.e. ISO 639 `language` code, ISO 15924 `script` code and ISO 3166 `region` code. To be able to interpret UN M49 region codes correctly, you'd need a mapping from these to ISO 3166 region codes and I don't see this anywhere in this PR.
Anyway, you *do* mention that support for these "is not fully baked in yet", but I'd suggest to leave this out of this PR completely. The utility of ISO 15924 `script` code is acknowledged by maintainers, support for UN M49 identifiers would be a different issue.

~**Addendum**: Note however that [IETF BCP47 language tags](https://en.wikipedia.org/wiki/IETF_language_tag) actually allow for UN.49 region codes in place of a ISO 3166 region code, e.g. `es-Latn-419` would indeed be a valid IETF BCP47 language tag. This means, that any platform that has full support for parsing such a language tag would also understand UN.49 region codes out of the box.~

(see https://github.com/JetBrains/compose-multiplatform/pull/5582#issuecomment-4334015384)

westnordost:
As a side-note, I always wondered about the odd `b+<language code>-<script code>` syntax in Android. It seems to me that a unique identifier for language + script + region code would be perfectly possible without the `b+` in front. And also, about the `<language code>-r<region code>` syntax, i.e. with the `r` in the middle.

Reading the [Android documentation on this matter](https://developer.android.com/guide/topics/resources/providing-resources#AlternativeResources), it looks like the above mentioned syntax simply came to be for historic reasons. Only with Android 7.0, the `b+<language code>-<script code>` was added.

So, wouldn't it be fine to just use 1:1 [IETF BCP47 language tags](https://en.wikipedia.org/wiki/IETF_language_tag) as directory names instead of inheriting this legacy Android syntax? E.g.

- `sr-Latn-RS`
- `zh-Hans`
- `en-GB`
 
instead of

- `b+sr+Latn+RS`
- `b+zh+Hans`
- `en-rGB`

Of course, using `en-GB` instead of `en-rGB` would be a **breaking change**, so, to not break existing software using Compose Multiplatform resources, `en-rGB` would still need to be understood correctly for backward compatibility.

**Disclaimer**: I am no maintainer here, this is just my opinion. It remains with maintainers such as @terrakok to make any decisions in this regard. I just want to allude to that now is probably the only sensible opportunity to drop that legacy syntax.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
westnordost: @terrakok I believe it would be `values-b+zh+Hant-dark`?

[Jetbrain's official documentation on resources](https://kotlinlang.org/docs/multiplatform/compose-multiplatform-resources-setup.html#qualifiers) doesn't mention in which order the qualifiers need to be listed. [The Android documentation on resource qualifiers](https://developer.android.com/guide/topics/resources/providing-resources#AlternativeResources) however _does_ mention an order, and the order is

1. language, script, region
2. theme
3. screen density

(Incidentally, that's the order in which the qualifiers are mentioned in Jetbrain's documentation too, the documentation just doesn't state that the qualifiers must be specified in that order exactly.)

2KAbhishek: Noticed this in the example on Jetbrain's docs

drawable-en-rUS-mdpi-dark

1. Lang script region
2. Density 
3. Theme

We should follow the same order right?

2KAbhishek: @terrakok need your inputs on this please, I will proceed accordingly

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
2KAbhishek: Resolved this

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
2KAbhishek: Added now

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
2KAbhishek: mentioned it on the updated description

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
2KAbhishek: I have went with [this pattern](https://github.com/JetBrains/compose-multiplatform/pull/5582#discussion_r3121487394) as mentioned in the docs, e.g: `values-b+zh+Hant-dark`

2KAbhishek:
> In one comment you write you also added support for [**UN M49 region codes**](https://en.wikipedia.org/wiki/UN_M49). From a multiplatform perspective, all you have in order to decide which resource to use is the `androidx.compose.ui.text.intl.Locale`, i.e. ISO 639 `language` code, ISO 15924 `script` code and ISO 3166 `region` code. To be able to interpret UN M49 region codes correctly, you'd need a mapping from these to ISO 3166 region codes and I don't see this anywhere in this PR. Anyway, you _do_ mention that support for these "is not fully baked in yet", but I'd suggest to leave this out of this PR completely. The utility of ISO 15924 `script` code is acknowledged by maintainers, support for UN M49 identifiers would be a different issue.

@westnordost can you please help by creating an issue for this on YouTrack? I think it will be better to take the UN M49 mapping as a subsequent fix and not include it on this PR - I can work on adding it

2KAbhishek:
> As a side-note, I always wondered about the odd `b+<language code>-<script code>` syntax in Android. It seems to me that a unique identifier for language + script + region code would be perfectly possible without the `b+` in front. And also, about the `<language code>-r<region code>` syntax, i.e. with the `r` in the middle.

@westnordost Appreciate the suggestion, but for this PR I'd like to keep scope tight: support what Android's resource system already supports - b+ and lang-rREGION, so resources can be ported from Android to CMP without extra steps.

Moving to IETF-pure folder syntax (sr-Latn-RS, en-GB) is a worthwhile discussion but I think it's a separate concern from script-code support and would need its own deprecation story for the existing region syntax. Happy to revisit as a follow-up issue if needed.

2KAbhishek:
@terrakok I have resolved the suggestions and did some clean up, can you please take another look 🙏🏽

westnordost:
> @westnordost Appreciate the suggestion, but for this PR I'd like to keep scope tight: support what Android's resource system already supports - b+ and lang-rREGION, so resources can be ported from Android to CMP without extra steps.

Right, but the resource directory names for Compose Multiplatform Resources *already* deviate from those from Android: Android uses `night`, `notnight`, CMP uses `dark`, `light`. Given that such a quite unnecessary change was already made, I am inferring that CMP-folks are more open towards deviating from the legacy Android resource directory naming scheme. Well, all that is just speculation from my part, though.

westnordost:
One note about UN49 support:

I wrote earlier...

> Addendum: Note however that [IETF BCP47 language tags](https://en.wikipedia.org/wiki/IETF_language_tag) actually allow for UN.49 region codes in place of a ISO 3166 region code, e.g. es-Latn-419 would indeed be a valid IETF BCP47 language tag. This means, that any platform that has full support for parsing such a language tag would also understand UN.49 region codes out of the box.

This is wrong, or at least not helpful. While in e.g. JavaScript, `es-419` is parsed without error (i.e. it is supported) `"419"` is just put into the `region` field. So what's actually missing is an API that would allow one to check whether `CL` is in `419`. Hence, regardless, a mapping UN.49 <-> ISO 3166 code in Kotlin would still be necessary for UN.49 support in directory names. Sorry for the noise.

2KAbhishek:
@terrakok just wanted to follow up for a re-review, can you please take another look

2KAbhishek:
@igordmn @eymar @Schahen @MatkovIvan 
Sorry for making noise, I wanted to get some eyes on this PR as it has been waiting for some time

Will be super helpful if you can take a look or let me know who can help with this 🙏🏽 
Thanks

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
```diff
@@ -110,11 +116,9 @@ internal abstract class GenerateResourceAccessorsTask : IdeaImportTask() {
     ): List<ResourceItem>? {
         val file = this
         val dirName = file.parentFile.name ?: return null
-        val typeAndQualifiers = dirName.split("-")
-        if (typeAndQualifiers.isEmpty()) return null
+        val qualifiers = parseComposeResourceQualifiers(dirName)
```
terrakok: Architecture: every qualifier parsing shold happen in single place (GeneratedResClassSpec.kt as it was before). At the moment the parsing logic lies in two different places.

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
```diff
@@ -48,8 +51,17 @@ internal val DefaultComposeEnvironment = object : ComposeEnvironment {
 
         //cache ResourceEnvironment unless compose environment is changed
         return remember(composeLocale, composeTheme, composeDensity) {
+            // androidx.compose.ui.text.intl.Locale doesn't expose script yet; borrow it
+            // from the system only when languages match, so locale overrides don't inherit a wrong script.
```
terrakok: attach a link to the issue in AOSP tracker, please

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
```diff
@@ -48,8 +51,17 @@ internal val DefaultComposeEnvironment = object : ComposeEnvironment {
...
+            val systemEnvironment = getSystemResourceEnvironment()
```
terrakok: this is expensive operation (as it is even documented!) don't use it here.
If you need to workaround it on android only then make a special expect/actual

components/resources/library/src/macosMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.macos.kt
```diff
@@ -30,8 +31,9 @@ internal actual fun getSystemEnvironment(): ResourceEnvironment {
 
     return ResourceEnvironment(
         language = LanguageQualifier(languageCode),
+        script = ScriptQualifier(scriptCode),
         region = RegionQualifier(regionCode.orEmpty()),
```
terrakok: follow the current style, please: scriptCode.orEmpty() here (not in the declaration)

components/resources/library/src/iosMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.ios.kt
```diff
@@ -18,8 +19,9 @@ internal actual fun getSystemEnvironment(): ResourceEnvironment {
     val density = mainScreen.scale.toFloat()
     return ResourceEnvironment(
         language = LanguageQualifier(languageCode),
+        script = ScriptQualifier(scriptCode),
```
terrakok: follow the current style, please: scriptCode.orEmpty() here (not in the declaration)

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
```diff
@@ -48,8 +51,17 @@ internal val DefaultComposeEnvironment = object : ComposeEnvironment {
...
+            val systemScript = if (composeLocale.language == systemEnvironment.language.language) {
```
terrakok: this is not clear logic. we need tests for that

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
terrakok: pass a bcp qualifier (string with pluses inside) to ResourceItem. and then inside `CodeBlock.Builder.addQualifier` extract real locale qualifiers

terrakok (changes requested):

westnordost:
(I just wanted to say, thanks for your second review, @terrakok , I stumbled upon this missing feature just now and saw with delight that there's progress here. I'll also read through the PR a bit)

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
westnordost: (The comment is also confusing. `androidx.compose.ui.text.intl.Locale` [**does** expose `script`](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/intl/Locale#script()), so you must mean something else, but what? )

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -48,6 +48,118 @@ class ResourcesTest : GradlePluginTestBase() {
...
+            // BCP 47 segment must start with a language subtag
+            "values-b+EN" to "BCP 47 segment must start with a language subtag",
+            // Folder name with no resource type prefix surfaces as 'Unknown resource type'
+            "-foo" to "Unknown resource type: ''",
+            // BCP 47 segment must be locale-first; theme/density before it is rejected
+            "values-dark-b+zh+Hant" to "BCP 47 'b+' segment must be the first qualifier",
+            "values-mdpi-b+zh" to "BCP 47 'b+' segment must be the first qualifier",
+            // two regions in one BCP 47 segment
+            "values-b+sr+US+RS" to "repetitive qualifiers",
+            "values-b+sr+Latn+RS+US" to "repetitive qualifiers",
```
westnordost: You could add tests for

```
"values-b+sr+sr+RS" to "repetitive qualifiers",
"values-b+sr+Latn+Latn" to "repetitive qualifiers",
```

Testing for duplicate language code, duplicate script code.

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -48,6 +48,118 @@ class ResourcesTest : GradlePluginTestBase() {
...
+        val invalidCases = listOf(
+            // malformed BCP 47 subtag (numeric where letters expected)
+            "values-b+sr+1234" to "Malformed BCP 47 subtag '1234'",
+            // junk subtag (not a lang / script / region pattern)
+            "values-b+sr+ABCD" to "Malformed BCP 47 subtag 'ABCD'",
+            // empty subtags inside a BCP 47 segment
+            "values-b+" to "Malformed BCP 47 subtag ''",
+            "values-b++Latn" to "Malformed BCP 47 subtag ''",
+            // BCP 47 segment must start with a language subtag
+            "values-b+EN" to "BCP 47 segment must start with a language subtag",
```
westnordost, line 132: You could add

```
            "values-b+Latn" to "BCP 47 segment must start with a language subtag",
```

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -48,6 +48,118 @@ class ResourcesTest : GradlePluginTestBase() {
...
+    @Test
+    fun testBcpFolderQualifiersInvalid() {
+        val invalidCases = listOf(
+            // malformed BCP 47 subtag (numeric where letters expected)
+            "values-b+sr+1234" to "Malformed BCP 47 subtag '1234'",
+            // junk subtag (not a lang / script / region pattern)
+            "values-b+sr+ABCD" to "Malformed BCP 47 subtag 'ABCD'",
+            // empty subtags inside a BCP 47 segment
+            "values-b+" to "Malformed BCP 47 subtag ''",
+            "values-b++Latn" to "Malformed BCP 47 subtag ''",
```
westnordost, line 126: You could add
```
            "values-b+sr+Latn+" to "Malformed BCP 47 subtag ''",
            "values-b+sr++" to "Malformed BCP 47 subtag ''",
            "values-b+sr++RS" to "Malformed BCP 47 subtag ''",
            "values-b+sr+RS+" to "Malformed BCP 47 subtag ''",
```

(testing for trailing plus, empty script, empty region)

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -48,6 +48,118 @@ class ResourcesTest : GradlePluginTestBase() {
...
+                    """
+                    <resources>
+                        <string name="bad">test</string>
+                    </resources>
+                    """.trimIndent()
+                )
+                gradleFailure("prepareKotlinIdeaImport").checks {
+                    check.logContains(errorMsg)
+                }
+                bad.deleteRecursively()
```
westnordost: Is this actually executed if the test fails here? A failing test should also clean up after itself.

Same for the positive tests.

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -48,6 +48,118 @@ class ResourcesTest : GradlePluginTestBase() {
...
+                "values-b+zh",
+                // language + script
+                "values-b+sr+Latn",
+                "values-b+zh+Hans",
+                // language only inside the b+ segment
+                "values-b+sr",
+                // language + letter region (no script)
+                "values-b+en+US",
+                // language + numeric (UN M.49) region
+                "values-b+es+419",
```
westnordost, line 65: I thought this should not be supported in this PR?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -48,6 +48,118 @@ class ResourcesTest : GradlePluginTestBase() {
...
+                    """
+                    <resources>
+                        <string name="bcp_test">test</string>
+                    </resources>
+                    """.trimIndent()
+                )
+            }
+            gradle("prepareKotlinIdeaImport").checks {
+                check.logDoesntContain("unknown qualifier")
+                check.logDoesntContain("repetitive qualifiers")
```
westnordost: But there are other error messages, no?? Hardcoding certain expected error messages appears fickle to me as it does not find all issues. Isn't there something (in the logs) which detects all errors?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
```diff
@@ -48,6 +48,118 @@ class ResourcesTest : GradlePluginTestBase() {
...
+            "values-b+sr+Latn+RS+US" to "repetitive qualifiers",
+            // BCP 47 segment cannot be combined with trailing locale-shaped qualifiers
+            "values-b+sr+Latn-rRS" to "Locale qualifier 'rRS' cannot follow a BCP 47",
+            "values-b+sr-rRS" to "Locale qualifier 'rRS' cannot follow a BCP 47",
+            // BCP 47 subtags must follow language -> script -> region order
+            "values-b+sr+RS+Latn" to "BCP 47 subtags must follow language -> script -> region order",
+            // Standalone script qualifiers (without b+) are not supported
+            "values-sr-Latn" to "Script qualifier 'Latn' must be inside a BCP 47",
+            "values-sr-rRS-Latn" to "Script qualifier 'Latn' must be inside a BCP 47",
+        )
```
westnordost, line 154: Definitely missing in the failure-tests is testing that `b+sr+RS` may not be used together with `sr-rRS`, i.e. its either-or. E.g.

```
"values-b+sr+RS-sr-rRS"
"values-sr-rRS-b+sr+RS"
```

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
terrakok: You added the comment that something doesn't work on Android 🤷

terrakok: (oh. I meant @2KAbhishek )

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GenerateResourceAccessorsTask.kt
2KAbhishek: Done, I have refactored the logic

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
2KAbhishek: This comment was misleading, I have removed it

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
2KAbhishek: Dome, I have implemented the expect / actual approach

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
2KAbhishek: Thanks for the catch, fixed it

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
2KAbhishek: Correct, full runtime fallback matching for numeric regions (like es-MX falling back to es-419) is deferred to a future PR. This PR only adds code-gen support to parse and normalize these subtags.

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
2KAbhishek: This is intentional to verify that the build fails precisely due to our validation rules, if you have any alternate solution, do let me know

2KAbhishek:
@terrakok I have resolved the change requests, can you please take another look and let me know what you think 

Thanks for the reviews

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/ResourcesTest.kt
westnordost: But this test in particular doesn't test that the build fails, it tests that the build *doesn't* fail.

westnordost:
By the way, if you force push, it looks like GitHub UI forgets which files one already had a look at (and didn't change since the last change). This makes it harder for reviewers, because then one has to read *everything* again, also the things that were fine, didn't need change and also weren't changed.

westnordost:
Nevermind my previous comment. Two files were still marked as viewed, so this disproves my prior conclusion.

components/resources/library/src/jsMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.js.kt
```diff
@@ -20,8 +24,20 @@ internal actual fun getSystemEnvironment(): ResourceEnvironment {
...
+
+internal actual fun androidx.compose.ui.text.intl.Locale.getScript(): String {
+    val tags = toLanguageTag().split('-', '_')
+    for (tag in tags) {
+        if (tag.length == 1) break
+        if (tag.length == 4 && tag.all { it.isLetter() }) {
+            return tag
+        }
+    }
+    return ""
```
westnordost: Why? As I wrote, `Locale.script` is defined.

components/resources/library/src/androidMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.android.kt
```diff
@@ -11,8 +11,20 @@ internal actual fun getSystemEnvironment(): ResourceEnvironment {
...
+internal actual fun androidx.compose.ui.text.intl.Locale.getScript(): String {
+    val script = platformLocale.script
+    if (script.isNotEmpty()) return script
+
+    val defaultLocale = java.util.Locale.getDefault()
+    if (language == defaultLocale.language) {
+        return defaultLocale.script
+    }
+    return ""
+}
```
westnordost: But [`Locale.script` exists](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/intl/Locale#script()), so no need to have some expect-actual here to define this function.
So now there is `Locale.script` and `Locale.getScript()`, the latter having some additional logic *but only on Android/JVM* platform.

IMO
1. that logic doesn't belong here; if it is necessary for one reason or another (please explain), then it should be done explicitly at the call site instead of "hiding" in a getter
2. I don't understand why this is necessary at all (add comment that explains it?)
3. I can not follow why the same logic should not be applied on all platforms (can be done without expect/actual)

2KAbhishek: you are right, I tested and verified, these getters are redundant, I will clean them up

components/resources/library/src/jsMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.js.kt
2KAbhishek: it was redundant, has been cleaned up, thanks

components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ResourceEnvironment.kt
```diff
@@ -40,6 +43,8 @@ internal interface ComposeEnvironment {
 }
 
 internal val DefaultComposeEnvironment = object : ComposeEnvironment {
+    private var cachedSystemEnvironment: ResourceEnvironment? = null
+
```
westnordost: What is this for? It doesn't seem to be used at all

2KAbhishek: this was from a prev iteration, I will remove and verify if anything else was missed

westnordost (commented):
LGTM. I didn't look through `GeneratedResClassSpec` closely.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
```diff
@@ -100,32 +106,103 @@ private fun CodeBlock.Builder.addQualifiers(resourceItem: ResourceItem): CodeBlo
...
+            error("Region qualifier must be declared after language: '$lang-$q'.\nFile: ${resourceItem.path}")
         }
-        add("%T(\"${q.takeLast(2)}\"), ", regionQualifier)
+        add("%T(\"$regionCode\"), ", regionQualifier)
     }
 
     return this
 }
 
+private fun parseComposeResourceQualifiers(qualifiers: List<String>, path: Path): List<String> {
```
terrakok: I don't understand why you want to expand `b+` to a list of strings.
Instead of that, add a one new branch in:
```kotlin
"mdpi",
"hdpi",
"xhdpi",
"xxhdpi",
"xxxhdpi",
"ldpi" -> {
    saveQualifier(densityQualifier, q)
}

else -> when {
    q.startsWith("b+") -> {
        val (language, region, script) = expandBcpQualifier(q)
        language?.let { saveQualifier(languageQualifier, it) }
        region?.let { saveQualifier(languageQualifier, it) }
        script?.let { saveQualifier(languageQualifier, it) }
    }

    q.matches(languageRegex) -> {
        saveQualifier(languageQualifier, q)
    }

    /...
}
```

terrakok (changes requested):
almost perfect 👌
one more comment

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
```diff
@@ -61,15 +61,19 @@ private val internalAnnotation = AnnotationSpec.builder(internalAnnotationClass)
 
 private val resourceContentHashAnnotationClass = ClassName("org.jetbrains.compose.resources", "ResourceContentHash")
 
+private val languageRegex = Regex("[a-z]{2,3}") // ISO 639
+private val scriptRegex = Regex("[A-Z][a-z]{3}") // ISO 15924
+private val regionAlpha2Regex = Regex("[A-Z]{2}") // ISO 3166-1 alpha-2
+private val regionNumericRegex = Regex("[0-9]{3}") // UN M.49
+private val androidRegionRegex = Regex("r[A-Z]{2}|r[0-9]{3}") // Android `r` prefix
```
terrakok, line 68: do we need global values here?

westnordost: How often is `CodeBlock.Builder.addQualifiers` called? Parsing 5 regexes each time the function is called might have performance impacts.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
2KAbhishek: Refactored this, thanks

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
2KAbhishek: Good point. addQualifiers is invoked per ResourceItem during accessor generation, so creating Regex(...) inside the function would probably recompile patterns repeatedly.

I kept these regexes as top-level private vals so they are compiled once and reused across calls.

2KAbhishek:
@terrakok resolved the comments, we should be ready for the final pass 😄 
Thank you

terrakok (approved):
I will add small fixes after the merge

terrakok:
@igordmn New API. need a second review

igordmn (approved):

terrakok:
https://github.com/JetBrains/compose-multiplatform/pull/5618
@2KAbhishek

PR #5585, opened 2026-04-29
Remove iosX64 and macosX64 targets across examples

eymar:
Compose-Runtime doesn't support iosX64 anymore due to deprecation in Kotlin. We remove it too.

## Release Notes
N/A

### DIFF
--- a/examples/chat/shared/build.gradle.kts
+++ b/examples/chat/shared/build.gradle.kts
@@ -30,13 +30,6 @@ kotlin {
         browser()
     }
 
-    macosX64 {
-        binaries {
-            executable {
-                entryPoint = "main"
-            }
-        }
-    }
     macosArm64 {
         binaries {
             executable {
@@ -85,9 +78,6 @@ kotlin {
 
         val macosMain by creating {
             dependsOn(commonMain)
-        }
-        val macosX64Main by getting {
-            dependsOn(macosMain)
         }
         val macosArm64Main by getting {
             dependsOn(macosMain)
--- a/examples/chat/shared/build.gradle.kts
+++ b/examples/chat/shared/build.gradle.kts
@@ -15,7 +15,6 @@ kotlin {
     jvm("desktop")
 
     listOf(
-        iosX64(),
         iosArm64(),
         iosSimulatorArm64()
     ).forEach { iosTarget ->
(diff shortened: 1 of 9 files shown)


MatkovIvan (commented):
Could you please fix the script that didn't report these failures to actually verify this?

eymar:
> Could you please fix the script that didn't report these failures to actually verify this?

```
19:32:35   e: ❌ KMP Dependencies Resolution Failure
19:32:35   Source set 'commonMain' couldn't resolve dependencies for all target platforms
19:32:35   Couldn't resolve dependency 'org.jetbrains.compose.runtime:runtime' in 'commonMain' for all target platforms.
19:32:35   The dependency should target platforms: [android, android, desktop, iosArm64, iosSimulatorArm64, iosX64]
19:32:35   Unresolved platforms: [iosX64]
```

The error we see is defined as STRONG_WARNING in KGP - https://github.com/JetBrains/kotlin/blob/master/libraries/tools/kotlin-gradle-plugin/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/diagnostics/KotlinToolingDiagnostics.kt#L120 

> object PartiallyResolvedKmpDependencies : ToolingDiagnosticFactory(STRONG_WARNING, DiagnosticGroup.Kgp.Misconfiguration)

That's why it doesn't fail the CI job.

Since there is even a KGP [test](https://github.com/JetBrains/kotlin/blob/b007223c59fc161eaed7b50fcabdf8476046d011/libraries/tools/kotlin-gradle-plugin-integration-tests/src/test/kotlin/org/jetbrains/kotlin/gradle/KmpPartiallyResolvedDependenciesCheckerIT.kt#L43) for this relaxed error reporting, it's intentional. But it would fail as soon as we attempt to build the missing target. 

I think we can just remove iosX64 and macosX64.

eymar:
Additionaly: https://youtrack.jetbrains.com/issue/KT-81536/Couldnt-resolve-dependency-in-commonMain-for-all-target-platforms-in-2.2.20-Beta2#focus=Comments-27-12867126.0-0

> this error is nothing but an annoying warning; it doesn't fail your builds or stop from using IDE. You should be able to compile, and write code without any problems.

MatkovIvan (approved):

PR #5591, opened 2026-05-04
CMP-4196 Add support for subdirectories in resource folders

cmelchior:
Fixes https://youtrack.jetbrains.com/issue/CMP-4196/Add-ability-to-have-subfolders-in-composeResources

This commit adds support for subdirectories in `/drawable-*` and `/font-*/` folders. The feature can be controlled using `ResourceHolder.allowAccessorSubDirs`.

**The following logic has been added:**

- [x] Subdirectories are transformed to object classes under the appropriate `Res.<type>`, e.g., `Res.drawable.subdir`. This puts some limitations on the naming, e.g., no symbols, no starting digit, no whitespace. Conversion rules can be added later, similar to how some file names are transformed.
- [x] Qualified resources will match across subdirectories, similar to how it is done now on the top-level.
- [x] It is disallowed to have directories and files create the same key, .e.g., `/drawable/subdir/` and `/drawable/subdir.png` will throw an exception.

**The logic for validating resource files has been improved:**

- [x] Calculating available ResourceItem's have been isolated in a new class: `ResourceHolder`, making it easier to unit test the logic and keep Gradle Tasks cleaner.
- [x] All invalid input errors are now `GradleException`, which provides a better experience in the Gradle output.
- [x] Illegal characters in file names, like whitespace or `.` will now show a proper exception message.
- [x] Files that would accidentally merge resource keys now throw an exception, e.g., `/drawable/compose.png` and `/drawable/compose.jpg`.
- [x] We now ignore case when collapsing resource directories. Previously, `/drawable/COMPOSE.png` and `/drawable/compose.png` would be two different keys, but this logic would break on case-insensitive filesystems. Now they collapse to the same key. The keyfrom the unqualified directory (e.g., `/drawable`) is used.
- [x] String keys across files (e.g., `strings.xml` and `strings2.xml`) could be duplicated without error, making it undefined which one was returned. This will now throw an exception

**Testing:**
- [x] All tests in `ResourceTest` pass.
- [x] Added new validation logic tests in `ResourceHolderValidationTest`.
- [x] https://github.com/JetBrains/compose-multiplatform/tree/master/components/resources example has been modified to show it working.
- [x] Manual testing shows that Android Studio preview still works
- [x] Check for the platform (Win/Linux/Mac) before running some of the validation cases. Some of them will be prevented outright by the host filesystem.
- [x] Fix Release Notes 

**Open Questions:**

- [ ] In this PR, the subdirectory support is enabled by default. However, the Compose Multiplatform Plugin is not yet aware of these changes. Because it optimistically re-creates the accessors when a user adds new resources or renames them, the plugin will break subdir support in these cases. This PR can be merged with `ResourceHolder.allowAccessorSubDirs = false`, and then the flag can be toggled once the Plugin has been updated. I'm not sure what the best approach is here?

- [ ] Some of the validation logic is now stricter. It is technically a breaking change to collapse resource names while ignoring their casing, but the previous implementation had bugs that didn't show up until the app was run on the "wrong" platform. If it is not acceptable to break this in this PR, a migration strategy should probably be put in place instead. But guidance here is needed.

### Features - Multiple Platforms
- Support for subfolders in `/composeResources/drawables` and `/composeResources/fonts`


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
```diff
@@ -189,8 +217,13 @@ internal fun getResFileSpec(
                     .build()
             )
 
-            ResourceType.values().forEach { type ->
-                resObject.addType(TypeSpec.objectBuilder(type.accessorName).build())
+            ResourceType.entries.forEach { type ->
+                val subdirPaths = typeToSubdirPaths[type].orEmpty()
+                resObject.addType(
+                    TypeSpec.objectBuilder(type.accessorName)
+                        .addSubdirObjects(subdirPaths)
```
terrakok, line 224: There’s a problem here that I’ve solved before regarding files: if there are too many directories, the compiler throws an error saying the generated file is too big. For example, subdirectories for a hundred locales. I solved this by splitting them into separate files with extensions.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/utils/gradleUtils.kt
```diff
@@ -72,3 +73,6 @@ internal inline fun <reified SubT> DomainObjectCollection<*>.configureEachWithTy
         }
     }
 }
+
+@Suppress("NOTHING_TO_INLINE")
+internal inline fun gradleError(message: String): Nothing = throw GradleException(message)
```
terrakok, line 78: What is the difference? If it is required then it should be updated every where

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/resources/GeneratedResClassSpec.kt
```diff
@@ -114,23 +131,34 @@ private fun CodeBlock.Builder.addQualifiers(resourceItem: ResourceItem): CodeBlo
...
         }
         add("%T(\"${q.takeLast(2)}\"), ", regionQualifier)
     }
 
     return this
 }
 
+private fun TypeSpec.Builder.addSubdirObjects(paths: Set<List<String>>): TypeSpec.Builder {
+    paths.groupBy { it.first() }.forEach { (segment, subPaths) ->
+        val nested = TypeSpec.objectBuilder(segment).addModifiers(KModifier.PUBLIC)
```
terrakok, line 148: does it support names like `is` or `package`?

terrakok (commented):
Hi! Thank you for the contribution! Sorry for the long delay. I reviewed the code. It is well in general but I have few comments:

1) The case of many directories and deep nesting is not handled — the generated `Res` file will cause a compilation error.

2) Subdirectories are supported only for images and fonts; This is strange not to have the support for strings as well.

3) There's a feature to get a list of all drawables for example. When subdirectories appear, it makes sense to be able to:
   - get files only from a specific subdirectory, or
   - get the full directory hierarchy.

4) Support is required in IDEA and in the Amper plugin (they forked our code).

5) This feature looks appealing at first glance but is not clearly in high demand:
   - Android does not have it; the existing requests are minor.
   - Problems may arise if Android later adds a similar feature but with different behavior — migration support could become difficult.
   - There is still a workaround: to use prefixes instead of subdirs.

We discussed it internally and agreed that we don't want to support the feature **for now** and we don't have enough resources to support it at the moment for such minor feature.

terrakok:
If we change our minds and decide to support this feature, I'll come back to this. 👍

PR #5598, opened 2026-05-11
Refactor gradle API

terrakok:
Migrate from `nullableProperty` to `notNullProperty` for improved property handling. Remove deprecated `nullableProperty` utility.

Fixes https://youtrack.jetbrains.com/issue/CMP-10189

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/MacOSNotarizationSettings.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/MacOSNotarizationSettings.kt
@@ -12,7 +12,7 @@ import org.gradle.api.tasks.Input
 import org.gradle.api.tasks.Internal
 import org.gradle.api.tasks.Optional
 import org.jetbrains.compose.desktop.application.internal.ComposeProperties
-import org.jetbrains.compose.internal.utils.notNullProperty
+import org.jetbrains.compose.internal.utils.property
 import javax.inject.Inject
 
 abstract class MacOSNotarizationSettings {
@@ -24,25 +24,25 @@ abstract class MacOSNotarizationSettings {
 
     @get:Input
     @get:Optional
-    val appleID: Property<String> = objects.notNullProperty<String>().apply {
+    val appleID: Property<String> = objects.property<String>().apply {
         set(ComposeProperties.macNotarizationAppleID(providers))
     }
 
     @get:Input
     @get:Optional
-    val password: Property<String> = objects.notNullProperty<String>().apply {
+    val password: Property<String> = objects.property<String>().apply {
         set(ComposeProperties.macNotarizationPassword(providers))
     }
 
     @get:Input
     @get:Optional
-    val teamID: Property<String> = objects.notNullProperty<String>().apply {
+    val teamID: Property<String> = objects.property<String>().apply {
         set(ComposeProperties.macNotarizationTeamID(providers))
     }
 
     @Deprecated("This option is no longer supported and got replaced by teamID", level = DeprecationLevel.ERROR)
     @get:Internal
-    val ascProvider: Property<String> = objects.notNullProperty<String>().apply {
+    val ascProvider: Property<String> = objects.property<String>().apply {
         set(providers.provider {
             throw UnsupportedOperationException("This option is not supported by notary tool and was replaced by teamID")
         })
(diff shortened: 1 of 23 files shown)


MatkovIvan (approved):
Public API change. Requires release notes + second approval

(and green CI)

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/utils/providerUtils.kt
```diff
@@ -13,34 +13,28 @@ import org.gradle.api.provider.Provider
 import org.gradle.api.provider.ProviderFactory
 import org.jetbrains.kotlin.gradle.plugin.extraProperties
 
-internal inline fun <reified T> ObjectFactory.new(vararg params: Any): T =
+internal inline fun <reified T : Any> ObjectFactory.new(vararg params: Any): T =
     newInstance(T::class.java, *params)
 
-@SuppressWarnings("UNCHECKED_CAST")
-internal inline fun <reified T : Any> ObjectFactory.nullableProperty(): Property<T?> =
```
igordmn, line 20: Did you check that all values that returned nullable are handled properly with `isPresent` or `orNull`?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/utils/providerUtils.kt
```diff
@@ -13,34 +13,28 @@ import org.gradle.api.provider.Provider
...
-    property(T::class.java) as Property<T?>
-
 internal inline fun <reified T : Any> ObjectFactory.notNullProperty(): Property<T> =
```
igordmn: `notNull` is misleading if all properties considered not null. 

It looks like we can just rename it to `property`

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/utils/providerUtils.kt
```diff
@@ -13,34 +13,28 @@ import org.gradle.api.provider.Provider
...
 
-internal inline fun <reified T> Task.provider(noinline fn: () -> T): Provider<T> =
+internal inline fun <reified T : Any> Task.provider(noinline fn: () -> T): Provider<T> =
     project.provider(fn)
 
-internal fun ProviderFactory.valueOrNull(prop: String): Provider<String?> =
-    provider {
-        gradleProperty(prop).orNull
-    }
+internal fun ProviderFactory.valueOrNull(prop: String): Provider<String> =
```
igordmn: This is misleading now if we can't have null properties. Let's just remove it and use `gradleProperty` directly

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/internal/utils/providerUtils.kt
terrakok: It is already uses `.orNull` because it wouldn't work before

igordmn: Do you mean that we can't pass null at all, and there is no place where we check on null instead of using `orNull`?

terrakok: instead of null there should be an empty state.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/MacOSNotarizationSettings.kt
```diff
@@ -24,25 +24,25 @@ abstract class MacOSNotarizationSettings {
 
     @get:Input
     @get:Optional
-    val appleID: Property<String?> = objects.nullableProperty<String>().apply {
+    val appleID: Property<String> = objects.property<String>().apply {
```
igordmn, line 27: Will users who pass `null` explicitly face a compilation error?

If so, add a Migration Note.

igordmn (approved):

MatkovIvan:
Please don't forget about release notes before merge

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/MacOSNotarizationSettings.kt
terrakok: Image

it is possible to set null there 😄

terrakok:
So, there is no visible changes

PR #5610, opened 2026-05-20
Changelog script fixes

mazunin-v-jb:
Refactored to omit using `__FILE__`, since it has [a caching bug](https://youtrack.jetbrains.com/issue/KT-77524/)

## Release Notes
N/A

### DIFF
--- a/tools/changelog/changelog.main.kts
+++ b/tools/changelog/changelog.main.kts
@@ -55,8 +55,10 @@ import java.time.format.DateTimeFormatter
 import java.util.*
 import kotlin.system.exitProcess
 
-val changelogFile = __FILE__.resolve("../../../CHANGELOG.md").canonicalFile
-val prFormatFile = File("PR_FORMAT.md")
+val scriptDir = getScriptPathFromArgs().parentFile
+
+val changelogFile = scriptDir.parentFile.parentFile.resolve("CHANGELOG.md")
+val prFormatFile = scriptDir.resolve("PR_FORMAT.md")
 val prFormatLink = "https://github.com/JetBrains/compose-multiplatform/blob/master/tools/changelog/PR_FORMAT.md"
 
 val argsKeyless = args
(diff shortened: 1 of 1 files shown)


tools/changelog/changelog.main.kts
```diff
@@ -55,8 +55,16 @@ import java.time.format.DateTimeFormatter
 import java.util.*
 import kotlin.system.exitProcess
 
-val changelogFile = __FILE__.resolve("../../../CHANGELOG.md").canonicalFile
-val prFormatFile = File("PR_FORMAT.md")
+val scriptDir = generateSequence(File(".").absoluteFile.canonicalFile) { dir ->
```
igordmn: Was the issue only with `File("PR_FORMAT.md")`, or with `__FILE__.resolve` too?

If only with `PR_FORMAT`, I would suggest using:
```
val scriptDir = __FILE__.parentFile
```

because it doesn't depend on where you call the script. With current code it seems you can't call it outside of the repo.

mazunin-v-jb: > Was the issue only with File("PR_FORMAT.md"), or with __FILE__.resolve too?

With both, actually. If I start it from the repo root, it can't find PR_FORMAT.md. If I start it in tools/changelog - `__FILE__.resolve("../../../CHANGELOG.md")` somehow resolves to just CHANGELOG.md (like ../../.. ignored completely).

> With current code it seems you can't call it outside of the repo.

Yes, I decided to add this to prevent it going to the system parent directories. I can revert it if needed.

igordmn: `__FILE__` is supposed to return the exact path to the script file and shouldn't depend on the working dir.

If it works correctly, we could use it to determine the `scriptDir` instead of iterating over parent folders. But it looks like it has [a caching bug](https://youtrack.jetbrains.com/issue/KT-77524/Kotlin-script-moving-a-script-doesnt-affect-its-FILE-variable#focus=Comments-27-13794669.0-0)

I tried to use this code and it works no matter where we call it:
```
fun getScriptPathFromArgs(): File =
    System.getProperty("sun.java.command")
        ?.split(" ")
        ?.find { it.endsWith(".kts") }
        ?.let { File(it).canonicalFile }
        ?: error("Can't find script path from args: ${System.getProperty("sun.java.command")}")

val scriptDir = getScriptPathFromArgs().parentFile
```

Let's use it?

igordmn (approved):

PR #5612, opened 2026-05-28
Add packagingResourcesRootDir to expose jpackage --resource-dir

CloudyDino:
jpackage's `--resource-dir` is the documented mechanism for overriding the packaging tool's own resources — the DMG volume icon (`<package-name>-volume.icns`), the DMG background, Linux `preinst`/`postinst` scripts, WiX templates, and so on (see the [jpackage docs](https://docs.oracle.com/en/java/javase/21/jpackage/override-jpackage-resources.html)).

Today Compose fully owns that directory and gives users no supported way to put files into it: `AbstractJPackageTask` points `--resource-dir` at `build/compose/tmp/resources`, and on every run it calls `clearDirs(...)` and then writes the generated `Info.plist` there. Because the clear happens inside the task action, dropping a file in via a `doFirst` is always wiped before jpackage runs. This blocks a recurring class of customizations.

This PR adds a `packagingResourcesRootDir` DSL property that mirrors the existing `appResourcesRootDir` (with `common`, `<os>` and `<os>-<arch>` subdirectories). Its files are copied into the resource dir **after** the dir is cleared but **before** the generated `Info.plist`/`product-def.plist` are written, so the Compose-managed plists always take precedence over user-supplied files.

```kotlin
compose.desktop {
    application {
        nativeDistributions {
            // drops e.g. packaging-resources/macos/MyApp-volume.icns into --resource-dir
            packagingResourcesRootDir.set(project.layout.projectDirectory.dir("packaging-resources"))
        }
    }
}
```

This revives the approach a maintainer suggested on #1843 (exposing the jpackage resource dir via the DSL rather than reusing `appResourcesDir`) and supersedes the abandoned #3647. Unlike #1843 — which repointed the internal `jpackageResources` dir at a `Sync` output and so had its files deleted by the task's own `clearDirs(...)` (the unresolved concern raised in that PR's review) — this copies the files in after the clear, so nothing is wiped.

Fixes [CMP-2042](https://youtrack.jetbrains.com/issue/CMP-2042) and [CMP-6122](https://youtrack.jetbrains.com/issue/CMP-6122) (both the DMG volume icon, set via `<package-name>-volume.icns`). Also helps [CMP-6411](https://youtrack.jetbrains.com/issue/CMP-6411), [CMP-1972](https://youtrack.jetbrains.com/issue/CMP-1972), and [CMP-9837](https://youtrack.jetbrains.com/issue/CMP-9837).

## Testing

Added an integration test (`DesktopApplicationTest.packagingResources`) with a new `application/packagingResources` test project. It runs `:createDistributable` and asserts that both the `common` file and the current-OS file are copied into `build/compose/tmp/resources` (the `--resource-dir` passed to jpackage). Verified locally on macOS:

```
DesktopApplicationTest > packagingResources(kotlin=2.3.20, gradle=9.3.1, agp=9.1.1) PASSED
```

`:compose:compileKotlin` and `:compose:compileTestKotlin` both build cleanly.

### Features - Gradle Plugin
- Added `nativeDistributions.packagingResourcesRootDir` to supply custom files to the packaging tool's resource directory (jpackage's `--resource-dir`), allowing overrides such as the DMG volume icon, DMG background, Linux `preinst`/`postinst` scripts, and WiX templates.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/AbstractDistributions.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/AbstractDistributions.kt
@@ -30,6 +30,20 @@ abstract class AbstractDistributions {
     var description: String? = null
     var vendor: String? = null
     val appResourcesRootDir: DirectoryProperty = objects.directoryProperty()
+
+    /**
+     * Root directory for files passed to the underlying packaging tool (jpackage)
+     * via `--resource-dir`, used to override the tool's own packaging resources (e.g.
+     * a `<package-name>-volume.icns` DMG volume icon, a DMG background, Linux
+     * `preinst`/`postinst` scripts, or WiX templates). See
+     * https://docs.oracle.com/en/java/javase/21/jpackage/override-jpackage-resources.html
+     *
+     * Files are resolved from `common`, `<os>` and `<os>-<arch>` subdirectories, the
+     * same layout as [appResourcesRootDir]. The generated `Info.plist` and
+     * `product-def.plist` are managed by Compose and always take precedence.
+     */
+    val packagingResourcesRootDir: DirectoryProperty = objects.directoryProperty()
+
     val licenseFile: RegularFileProperty = objects.fileProperty()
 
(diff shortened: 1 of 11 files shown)


Siedlerchr:
It would be really great to have this merged soon!

PR #5613, opened 2026-05-29
Bump CHR version to 1.2.0-alpha01

pjBooms:
+ use JDK 25 as CHR requires it since 1.2.0-alpha01

Fixes [CMP-10253](https://youtrack.jetbrains.com/issue/CMP-10253)

### Features - Desktop
- The default ProGuard version is set to 7.8.0
- _(prerelease fix)_ Bump Compose Hot Reload to [1.2.0-alpha01](https://github.com/JetBrains/compose-hot-reload/releases/tag/v1.2.0-alpha01)

### DIFF
--- a/.github/workflows/gradle-plugin.yml
+++ b/.github/workflows/gradle-plugin.yml
@@ -24,11 +24,11 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v5
 
-      - name: Setup JDK 21
+      - name: Setup JDK 25
         uses: actions/setup-java@v4
         with:
           distribution: 'jetbrains'
-          java-version: '21'
+          java-version: '25'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/ci/docker/linux-tests/Dockerfile
+++ b/ci/docker/linux-tests/Dockerfile
@@ -44,7 +44,7 @@ ENV LANG=en_US.UTF-8 LANGUAGE=en_US:en LC_ALL=en_US.UTF-8
 ENV JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF-8
 
 # Install JBR
-ARG JBR_DISTR=jbr_jcef-21.0.8-linux-x64-b1163.62
+ARG JBR_DISTR=jbr_jcef-25.0.3-linux-x64-b496.62
 ARG JBR_DISTR_ARCH=$JBR_DISTR.tar.gz
 RUN wget https://cache-redirector.jetbrains.com/intellij-jbr/$JBR_DISTR_ARCH && \
     tar xzf $JBR_DISTR_ARCH && \
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
@@ -12,7 +12,7 @@ import org.jetbrains.compose.internal.utils.notNullProperty
 import org.jetbrains.compose.internal.utils.nullableProperty
 import javax.inject.Inject
 
-private const val DEFAULT_PROGUARD_VERSION = "7.7.0"
+private const val DEFAULT_PROGUARD_VERSION = "7.8.0"
 
 abstract class ProguardSettings @Inject constructor(
     objects: ObjectFactory,
--- a/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
+++ b/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
@@ -103,7 +103,7 @@ class HotReloadTest : GradlePluginTestBase() {
 
     @Test
     fun testExternalHotReload() = with(testProject("application/mpp")) {
-        val externalHotReloadVersion = "1.0.0-rc01"
+        val externalHotReloadVersion = "1.1.0"
         modifyText("settings.gradle") {
             //  Set the explicit version of Compose Hot Reload in the "pluginManagement {" block
             it.replace(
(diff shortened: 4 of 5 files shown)


.github/workflows/gradle-plugin.yml
```diff
@@ -24,11 +24,11 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v5
 
-      - name: Setup JDK 21
+      - name: Setup JDK 25
         uses: actions/setup-java@v4
         with:
           distribution: 'jetbrains'
-          java-version: '21'
+          java-version: '25'
```
kropp, line 31: What would be the new Gradle requirements given the updated JDK for the plugin?

pjBooms: It seems we don't have any explicitly stated requirements for Gradle now but Gradle supports JDK 25 starting from 9.1.0

ci/docker/linux-tests/Dockerfile
MatkovIvan, line 1: Please test the TeamCity pipeline with a new `Dockerfile`
(publish it from your branch **with custom tag** and pass this custom tag into SNAPSHOT build)

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
```diff
@@ -12,7 +12,7 @@ import org.jetbrains.compose.internal.utils.notNullProperty
 import org.jetbrains.compose.internal.utils.nullableProperty
 import javax.inject.Inject
 
-private const val DEFAULT_PROGUARD_VERSION = "7.7.0"
+private const val DEFAULT_PROGUARD_VERSION = "7.8.0"
```
MatkovIvan, line 15: Should we document it?
I remember such updates were requested externally before - [CMP-3818](https://youtrack.jetbrains.com/issue/CMP-3818)

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
```diff
@@ -124,6 +124,15 @@ class HotReloadTest : GradlePluginTestBase() {
                 """.trimIndent()
             )
         }
+        modifyText("build.gradle") {
+            //  Set JDK 25 toolchain as we use JBR 25 on CI
+            it.replace("jvm()",
+                """
+                    jvm()
+                    jvmToolchain(25)
```
MatkovIvan, line 132: "jvmToolchain" requires 25 to be installed even when we'll use 26+
Can we please prefer not-so-narrow enforcement (like in other examples in this repo)?

ci/docker/linux-tests/Dockerfile
```diff
@@ -44,7 +44,7 @@ ENV LANG=en_US.UTF-8 LANGUAGE=en_US:en LC_ALL=en_US.UTF-8
 ENV JAVA_TOOL_OPTIONS=-Dfile.encoding=UTF-8
 
 # Install JBR
-ARG JBR_DISTR=jbr_jcef-21.0.8-linux-x64-b1163.62
+ARG JBR_DISTR=jbr_jcef-25.0.3-linux-x64-b496.62
```
MatkovIvan, line 47: Does it mean that CHR won't work with JBR 21 anymore?
Did we agree on such requirements bump?
Is it documented?

---

If not, how we can make sure that it still works with JBR 21 if we remove such environment from testing?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
pjBooms: Do we have a place to document it? Ideally, we should periodically check all external dependencies and update them.

ci/docker/linux-tests/Dockerfile
pjBooms: > Does it mean that CHR won't work with JBR 21 anymore?

If a project has `jvmToolchain(21)` or earlier, CHR picks up JBR 21 by default. Else it uses JBR 25 (can be configured  though). 

>If not, how we can make sure that it still works with JBR 21 if we remove such environment from testing?

CHR has its own tests that test different versions of JBR starting from JBR 11. I don't think that we need to test JBR 21 for integration with CMP, default is enough, IMHO.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
MatkovIvan: I meant release notes at least, not sure that we have this somewhere else

ci/docker/linux-tests/Dockerfile
MatkovIvan: Can we use the version from CMP test environment in this case? Why do we need to update the test environment to update CHR?
We probably can, but better to do it separately

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/HotReloadTest.kt
pjBooms: The test tests some "previous" CHR, so if we set its version for a project it will be choosen and not the bundled one. Now we use CHR 1.1.0 as that "previous" CHR that supports JBRs upto JBR 25. However as default for it is 21 we have to set `jvmToolchain(25)`  so it picks up JBR 25. When we will use 26+ on CI (it will happen once we change default JBR for CHR again), we will need to update this test again to choose CHR version that supports 26+ and that toolchain should be updated too. Unfortunately, it works like this for this test.

ci/docker/linux-tests/Dockerfile
pjBooms: Done. Please check

ci/docker/linux-tests/Dockerfile
pjBooms: I am not sure what you mean. CHR test environment lives separately from CMP's now. 

>  Why do we need to update the test environment to update CHR?

We need to do that only when default JBR for CHR changes. It happens rarely.  Default JBR for CHR has to be present in CMP test environment to test integration of CHR into CMP.

Does it answer your question?

MatkovIvan (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/ProguardSettings.kt
pjBooms: Added to release notes

PR #5617, opened 2026-06-09
Web: empty UI when resource is cancelled

terrakok:
Extract the loading job out of the UI scope

Fixes https://youtrack.jetbrains.com/issue/CMP-10137

## Testing
Added unit tests

### Fixes - Resources
- Fix an issue when resource loading might be cancelled and UI became empty


Schahen (approved):
In ideal world I clearly can see that there are some cases where we'll benefit from cancellation but even if so, this should be designed and discussed separately.
This particular PR achieves what it aims for

PR #5620, opened 2026-06-11
Bump CHR version to 1.2.0-alpha02

pjBooms:
Fixes [CMP-10314](https://youtrack.jetbrains.com/issue/CMP-10314)

### Features - Desktop
- _(prerelease fix)_ Bump Compose Hot Reload to [1.2.0-alpha02](https://github.com/JetBrains/compose-hot-reload/releases/tag/v1.2.0-alpha02)

### DIFF
--- a/gradle-plugins/gradle/libs.versions.toml
+++ b/gradle-plugins/gradle/libs.versions.toml
@@ -6,7 +6,7 @@ plugin-android = "8.10.1"
 shadow-jar = "9.1.0"
 publish-plugin = "1.2.1"
 # we use "prefer" here for the strategy: "explicitly specified hot reload plugin always wins".
-plugin-hot-reload = { prefer = "1.2.0-alpha01" }
+plugin-hot-reload = { prefer = "1.2.0-alpha02" }
 
 [libraries]
 download-task = { module = "de.undercouch:gradle-download-task", version.ref = "gradle-download-plugin" }


gradle-plugins/gradle/libs.versions.toml
```diff
@@ -6,7 +6,7 @@ plugin-android = "8.10.1"
 shadow-jar = "9.1.0"
 publish-plugin = "1.2.1"
 # we use "prefer" here for the strategy: "explicitly specified hot reload plugin always wins".
-plugin-hot-reload = { prefer = "1.2.0-alpha01" }
+plugin-hot-reload = { prefer = "1.2.0-alpha02" }
```
MatkovIvan, line 9: The next release will be beta01 (we already did a build cut for CMP alpha02).
That will require CHR beta too. So we do not expect to have any release with this version => we don't really need it. It applies to release notes too

pjBooms: I have already discussed this issue with @kropp and we agreed that it is ok to have CHR 1.2.0-alpha02 in CMP 1.12.0-beta01. @kropp please confirm

igordmn: As for an utility that doesn't affect user apps, this is okay.

If this would be a public (not internal) library then we would have an issue.

I suspect that this was driven by the too complex and separate release processes - we might need to take this into account when we revisit them.

kropp: Let's have it as an exception this time, due to issues with CI preventing CHR release on time and the need to progress on the topic as fast as possible.

pjBooms: Please approve someone then

igordmn (approved):

MatkovIvan (approved):

PR #5621, opened 2026-06-11
Web: ensure Skiko runtime is initialized before JS browser tests start

terrakok:
`configureWebApplication` now wires JS test targets with generated Karma config that:
 - serves `js-reexport-symbols.mjs`, `skiko.mjs`, and `skiko.wasm`,
 - injects a loader before test bundle execution,
 - delays `window.__karma__.loaded` until `js-reexport-symbols.mjs` is imported and `api.awaitSkiko` is resolved.

This removes Skiko init race in jsBrowserTest for projects without explicit app executable setup and prevents runtime failures like missing/invalid `org_jetbrains_skia_*` symbols.

Fixes https://youtrack.jetbrains.com/issue/CMP-4906

## Testing
Added test case for the issue

## UPDATE
Reverted: https://github.com/JetBrains/compose-multiplatform/pull/5628

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
@@ -222,11 +222,15 @@ private fun configureJsBrowserTestsSkikoLoading(
                     let skikoReady = null;
                     window.__karma__.loaded = function() {
                       if (!skikoReady) {
-                        skikoReady = import("/base/kotlin/js-reexport-symbols.mjs")
-                          .then((mod) => mod?.api?.awaitSkiko || Promise.resolve());
+                        const servedFiles = window.__karma__.files || {};
+                        const reexportUrl = Object.keys(servedFiles)
+                          .find((url) => url.endsWith("js-reexport-symbols.mjs"));
+                        skikoReady = reexportUrl
+                          ? import(reexportUrl).then((mod) => mod?.api?.awaitSkiko ?? Promise.resolve())
+                          : Promise.resolve();
                       }
                       skikoReady.then(() => originalLoaded()).catch((error) => {
-                        const message = error && error.stack ? error.stack : String(error);
+                        const message = error?.stack ?? String(error);
                         window.__karma__.error(message);
                       });
                     };
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
@@ -5,6 +5,7 @@
 
 package org.jetbrains.compose.web.internal
 
+import org.gradle.api.DefaultTask
 import org.gradle.api.Project
 import org.gradle.api.artifacts.Configuration
 import org.gradle.api.artifacts.ResolvedDependency
@@ -20,7 +21,11 @@ import org.jetbrains.compose.internal.utils.file
 import org.jetbrains.compose.internal.utils.registerTask
 import org.jetbrains.compose.web.WebExtension
 import org.jetbrains.compose.web.tasks.UnpackSkikoWasmRuntimeTask
+import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
 import org.jetbrains.kotlin.gradle.targets.js.ir.KotlinJsIrTarget
+import org.jetbrains.kotlin.gradle.targets.js.testing.KotlinJsTest
+import org.jetbrains.kotlin.gradle.targets.js.testing.karma.KotlinKarma
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 internal fun Project.configureWeb(
     composeExt: ComposeExtension,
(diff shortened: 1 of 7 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -110,11 +115,141 @@ internal fun configureWebApplication(
...
+                  fs.writeFileSync(loaderFile, `
+                  (function() {
+                    if (!window.__karma__) return;
+                    const originalLoaded = window.__karma__.loaded.bind(window.__karma__);
+                    let skikoReady = null;
+                    window.__karma__.loaded = function() {
+                      if (!skikoReady) {
+                        skikoReady = import("/base/kotlin/js-reexport-symbols.mjs")
+                          .then((mod) => mod?.api?.awaitSkiko || Promise.resolve());
+                      }
```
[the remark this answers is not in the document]
Schahen: You can either:
1) filter by exact key 
2) at least just use ?. operator 

I agree with Mr. Copilot - this obe is twoo fragile

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -110,11 +115,141 @@ internal fun configureWebApplication(
...
+                      skikoReady.then(() => originalLoaded()).catch((error) => {
+                        const message = error && error.stack ? error.stack : String(error);
```
Schahen: the same - we can safely use null coalescing

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -110,11 +115,141 @@ internal fun configureWebApplication(
...
+                        window.__karma__.error(message);
+                      });
+                    };
+                  })();
+                  `.trim());
+
+                  const hasLoader = files.some((entry) =>
+                    entry === loaderFile ||
+                    (entry && typeof entry === "object" && entry.pattern === loaderFile)
```
Schahen, line 242: Not a big deal but `===` wont be a real benefit for anything

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -110,11 +115,141 @@ internal fun configureWebApplication(
...
+                  const loaderFile = path.resolve(path.dirname(testEntry), "compose-skiko-loader.js");
+                  if (!fs.existsSync(reexportModule)) return;
+                
+                  const ensureServed = (filePath) => {
+                    if (!fs.existsSync(filePath)) return;
+                    const exists = files.some((entry) =>
+                      entry === filePath ||
+                      (entry && typeof entry === "object" && entry.pattern === filePath)
+                    );
+                    if (!exists) {
```
ilgonmic: nit: `exists` -> `served`?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -110,11 +115,141 @@ internal fun configureWebApplication(
...
+                      files.push({
```
ilgonmic: probably better to use `unshift` instead of `push`
Because the order matters, it will be the same order in a browser, but we  already have "main" entry file in `files` array. So 
I think we need to serve "dependencies" first

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -110,11 +115,141 @@ internal fun configureWebApplication(
...
+            testTask.compilation.compilationName != KotlinCompilation.TEST_COMPILATION_NAME
+        ) {
+            return@configureEach
+        }
+
+        testTask.dependsOn(generateConfigTask)
+        testTask.dependsOn(compilationProcessResourcesTaskName)
+
+        val configDirectoryPath = configDir.get().asFile
+        (testTask.testFramework as? KotlinKarma)?.useConfigDirectory(configDirectoryPath)
```
ilgonmic, line 264: Just in case, `useConfigDirectory` overrides the default directory. If a user has their own configuration in `karma.config.d`, as far as I understand it stops to work

terrakok: Added a copying configs from the default directory (if they exist). For a custom config dir we don't have an API to access it.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
terrakok: done

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
terrakok: 👌

Schahen (commented):
My comments doesn't worth trying to implement quite honestly  - as overall make sense as a temporary decision

Schahen (approved):
Ignore my very minor comments, as a temporary solution to achieve what we want looks absolutely justified

PR #5624, opened 2026-06-16
Increase load on the CanvasDrawing test

ASalavei:
CanvasDrawing became too easy for 60 FPS devices and became less useful.
The MR increase number of draw calls during this tests to provide more versatile data.

Also, enable Always On screen on iOS to eliminate screen deeming issues during the tests.

## Release Notes
N/A

### DIFF
--- a/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/benchmarks/canvasdrawing/CanvasDrawing.kt
+++ b/benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/benchmarks/canvasdrawing/CanvasDrawing.kt
@@ -95,7 +95,7 @@ private fun CanvasDrawingItem(index: Int) {
             )
         )
 
-        repeat(200) { i ->
+        repeat(400) { i ->
             val angle = animatedValue + i * 7.2f
             val radius = size.minDimension / 4f
             val x = size.width / 2 + cos(angle * PI / 180.0).toFloat() * radius * random.nextFloat()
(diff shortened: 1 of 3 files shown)


benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/benchmarks/canvasdrawing/CanvasDrawing.kt
```diff
@@ -95,15 +95,15 @@ private fun CanvasDrawingItem(index: Int) {
             )
         )
 
-        repeat(200) { i ->
+        repeat(300) { i ->
```
pjBooms, line 197: We already have a history for CanvasDrawing on the benchmark server. So I would suggest to introduce parameters for the benchmark and register two benchmarks: with old values and new ones (say CanvasDrawingHeavy). See also similar thing for LazyGrid or MultipleComponents.

pjBooms (changes requested):

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/benchmarks/canvasdrawing/CanvasDrawing.kt
ASalavei: Done. I checked a solution how to tweak params to increase complexity, but at the end - we have to tweak too much, so I added a new list cell here.

benchmarks/multiplatform/benchmarks/src/commonMain/kotlin/benchmarks/canvasdrawing/CanvasDrawing.kt
```diff
@@ -61,13 +65,16 @@ fun CanvasDrawing() {
...
+                CanvasDrawingComplexity.easy -> EasyCanvasDrawingItem(index)
+                CanvasDrawingComplexity.medium -> MediumCanvasDrawingItem(index)
+            }
         }
     }
 }
 
 @Composable
-private fun CanvasDrawingItem(index: Int) {
+private fun EasyCanvasDrawingItem(index: Int) {
```
pjBooms, line 77: Easy and Medium Items have many things in common. I would de-duplicate them.

pjBooms (approved):

PR #5627, opened 2026-06-16
Handle exceptions when loading resources and fetching assets

terrakok:
Handle exceptions when loading resources and fetching assets

Fixes https://youtrack.jetbrains.com/issue/CMP-10325

## Testing
Add related test cases

### Fixes - Resources
- Web: fixed a app crash when a resource fetch is failed


Schahen (approved):
Have nothing to object to

PR #5629, opened 2026-06-19
[benchmarks] CMP-10298: Extract Compose scene API into versioned modules for cross-version compatibility

pjBooms:
Benchmarks used internal Compose UI API (`CanvasLayersComposeScene`,
`ComposeScene.render`) directly in `MeasureComposable.kt`, which broke
when [compose-multiplatform-core#3012](https://github.com/JetBrains/compose-multiplatform-core/pull/3012) replaced the single `render()`
method with `FrameRecomposer` + `measureAndLayout()` + `draw()` phases.

Introduce a versioned abstraction layer:

- `:compose-scene-api` — BenchmarkComposeScene interface (`setContent`,
  `render`, `close`) depending only on public Compose API.
- `:compose-scene-impl-1` — implementation for Compose versions before
  PR [compose-multiplatform-core#3012](https://github.com/JetBrains/compose-multiplatform-core/pull/3012) (≤1.11.x, `1.12.0-alpha01`, `alpha02` dev builds <4221),
  using ComposeScene.render(canvas, nanoTime).
- `:compose-scene-impl-2` — implementation for Compose versions after
  PR [compose-multiplatform-core#3012](https://github.com/JetBrains/compose-multiplatform-core/pull/3012) (`1.12.0-alpha02+dev4221+`), using `FrameRecomposer` with
  `performFrame()`, `measureAndLayout()`, and `draw()`.

The correct impl module is auto-selected in `settings.gradle.kts`
based on the Compose version (from `libs.versions.toml` or
`-Pcompose.version`), and mapped as `:compose-scene-impl` so
`MeasureComposable.kt` uses only the stable `BenchmarkComposeScene`
interface without any `@InternalComposeUiApi` usage.

Fixes [CMP-10298](https://youtrack.jetbrains.com/issue/CMP-10298)

## Testing
Tested manually.

## Release Notes
N/A

### DIFF
--- a/benchmarks/multiplatform/settings.gradle.kts
+++ b/benchmarks/multiplatform/settings.gradle.kts
@@ -28,4 +28,97 @@ dependencyResolutionManagement {
     }
 }
 
+include(":compose-scene-api")
+
+/**
+ * Selects the appropriate compose-scene-impl module based on the Compose Multiplatform version.
+ *
+ * - compose-scene-impl-1: for versions before compose-multiplatform-core#3012
+ *   Uses CanvasLayersComposeScene and ComposeScene.render(canvas, nanoTime)
+ *
+ * - compose-scene-impl-2: for versions after compose-multiplatform-core#3012
+ *   Uses FrameRecomposer + ComposeScene.measureAndLayout() + ComposeScene.draw(canvas)
+ *
+ * The breaking change (PR #3012) was introduced between dev builds 4213 and 4221
+ * of 1.12.0-alpha02, so dev builds before dev4221 still use the old API.
+ * 1.12.0-alpha01 also uses the old API (before the PR).
+ *
+ * The selected module is included as ":compose-scene-impl" so that the benchmarks module
+ * always depends on ":compose-scene-impl" regardless of which actual module is used.
+ */
+fun resolveComposeVersion(): String {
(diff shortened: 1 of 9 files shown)


benchmarks/multiplatform/settings.gradle.kts
```diff
@@ -28,4 +28,97 @@ dependencyResolutionManagement {
...
+        for (line in lines) {
+            val match = versionRegex.find(line)
+            if (match != null) return match.groupValues[1]
+        }
+    }
+    error("Could not find compose-multiplatform version in gradle/libs.versions.toml")
+}
+
+fun isAfterPR3012(version: String): Boolean {
+    // Version format examples:
```
MatkovIvan: I suggest copying https://github.com/JetBrains/compose-multiplatform-core/blob/jb-main/buildSrc/public/src/main/kotlin/androidx/build/Version.kt to buildSrc here and get generic parsing with comparable versions at the end. It will make this function trivial (now it's kinda complicated and hard to read)

pjBooms: Embedded Version to `settings.gradle.kts` because it does not have access to `buildSrc` (only `build.gradle.kts` has).

MatkovIvan (approved):

PR #5639, opened 2026-07-06
Fix CMP-6819 by properly setting previousDrawSize

rock3r:
Fixes https://youtrack.jetbrains.com/issue/CMP-6819/Too-much-rendering-in-SvgPainter

The SvgPainter was supposed to be caching the rasterised SVGs, but it failed to properly update the size it was rasterised at. As a result, the cached bitmap was never used even if the size had not changed.

This fixes it by saving the cached draw size, so that it can actually be reused as it was originally intended. It also adds a test that verifies the behaviour with the size and scale modifiers is correct.

Fixes CMP-6819

## Testing
Added test as per above.

### Fixes - Desktop
- Fixed SVG rasterization caching to properly reuse cached bitmap unless size changes.


igordmn:
@terrakok, could you review it

terrakok (approved):

PR #5640, opened 2026-07-08
Support desktop-jvm-all and deprecate desktop.currentOs

hub-bla:
Fixes [CMP-9176](https://youtrack.jetbrains.com/issue/CMP-9176)

Follow up to: https://github.com/JetBrains/compose-multiplatform-core/pull/3198

## Testing
Tested locally with a fresh compose project

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
@@ -132,6 +132,11 @@ abstract class ComposePlugin : Plugin<Project> {
         @Deprecated("Specify dependency directly", replaceWith = ReplaceWith("\"org.jetbrains.compose.ui:ui-test-junit4:${ComposeBuildConfig.composeVersion}\""))
         val uiTestJUnit4 get() = composeDependency("org.jetbrains.compose.ui:ui-test-junit4")
 
+        @Deprecated(
+            "Use org.jetbrains.compose.desktop:desktop-jvm-all instead. " +
+                "It works for every desktop OS; packaging drops unrelated OS binaries.",
+            replaceWith = ReplaceWith("\"org.jetbrains.compose.desktop:desktop-jvm-all:${ComposeBuildConfig.composeVersion}\"")
+        )
         val currentOs by lazy {
             composeDependency("org.jetbrains.compose.desktop:desktop-jvm-${currentTarget.id}")
         }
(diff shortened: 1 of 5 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -132,8 +131,9 @@ abstract class ComposePlugin : Plugin<Project> {
         @Deprecated("Specify dependency directly", replaceWith = ReplaceWith("\"org.jetbrains.compose.ui:ui-test-junit4:${ComposeBuildConfig.composeVersion}\""))
         val uiTestJUnit4 get() = composeDependency("org.jetbrains.compose.ui:ui-test-junit4")
 
+        @Deprecated("Specify dependency directly", replaceWith = ReplaceWith("\"org.jetbrains.compose.desktop:desktop:${ComposeBuildConfig.composeVersion}\""))
         val currentOs by lazy {
-            composeDependency("org.jetbrains.compose.desktop:desktop-jvm-${currentTarget.id}")
+            composeDependency("org.jetbrains.compose.desktop:desktop")
```
MatkovIvan: Shouldn't existing/deprecated property maintain previous behavior?

hub-bla: The most reasonable thing for me would be to remove it as we're not going to have os specific artifacts anymore but I was not sure if we can

igordmn:
Approval from [Alexander Maryanovsky (@m-sasha)](https://github.com/m-sasha) is also needed.

m-sasha (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/tasks/AbstractJarsFlattenTask.kt
```diff
@@ -49,6 +54,9 @@ abstract class AbstractJarsFlattenTask : AbstractComposeDesktopTask() {
     @get:OutputFile
     val flattenedJar: RegularFileProperty = objects.fileProperty()
 
+    @get:Input
+    val stripSkikoNativesFromAllJars: Property<Boolean> = objects.property(Boolean::class.java).value(false)
```
MatkovIvan: Is it a public DSL API for our Gradle plugin?

hub-bla: No, its setup only internally. I did it for the case when proguard is set with `joinOutputJars` to true but actually it works without it as well so I'll just drop that

MatkovIvan (approved):

igordmn (approved):

PR #5642, opened 2026-07-09
Simplify Skiko runtime setup

terrakok:
New version of skiko has its runtime as a part of the library. The PR adopts this change

Related to https://youtrack.jetbrains.com/issue/SKIKO-1104

## Release Notes
N/A


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt
```diff
@@ -11,118 +11,81 @@ import org.gradle.api.Project
...
-                    }
-                }
-            } else {
-                // Kotlin/JS depends on Skiko through global space.
-                // Bundler cannot know anything about global externals, so that’s why we need to copy it to final dist
-                project.tasks.named(compilation.processResourcesTaskName, ProcessResources::class.java) {
-                    it.from(unpackedRuntimeDir)
+    target.compilations.all { compilation ->
+        // `wasmTargetType` is available starting with kotlin 1.9.2x
+        if (target.wasmTargetType != null) {
```
MatkovIvan, line 70: It seems we can drop pre 1.9.2x fallback

terrakok: actually, this is a just check weather the target is wasm or js

MatkovIvan: Could you please make this comment less misleading?

MatkovIvan (approved):

PR #5644, opened 2026-07-10
Add support for AppCDS and AOT

m-sasha:
[AppCDS](https://dev.java/learn/jvm/cds-appcds/) and [AOT](https://openjdk.org/jeps/514) are new mechanisms in the JDK/JVM which allow creating a "class archive" that helps the JVM to load classes much faster than normally, significantly reducing startup time.

# Quick-start
To create/use this archive, the application needs to be run with special JVM flags.

Note that the distributable will be larger than usual because at minimum, the archive of the JRE classes needs to be packaged with the app.

The API is a new section in `compose.desktop { application { ... } }`, e.g.
```
compose.desktop {
    application {
        aot {
            mode = AotMode.AotPrebuild
            logging = true
        }
    }
}
```

There are currently four modes:
## None
The "normal" mode where nothing special happens and everything runs as before. This is the default.

## AppCdsAuto
In this mode, an [AppCDS](https://dev.java/learn/jvm/cds-appcds/) class archive is created when the *end-user* first runs the app.
**Requires JDK 19 or later.**

### Advantages:
- Simplest - no additional step is needed to build the archive.
- Creates a smaller distributable.
- Subsequent executions of the app will be faster.

### Drawbacks:
- The archive is not available at the first execution of the app, so it is slower (possibly even slower than regular execution).
- The archive is created at shutdown time of the first execution, which also takes a little longer.

## AppCdsPrebuild
In this mode Gradle executes a "training run" of the app before packaging it to create the [AppCDS](https://dev.java/learn/jvm/cds-appcds/) class archive.
**Requires JDK 21 or later.**

### Advantages:
- Fast startup, including the first run.

### Drawbacks:
 - Requires an additional step of running the app when building the distributable. See the "Training Run" section for details.
 - The distributable is larger because it includes the archive of the app's classes.

## AotPrebuild
This mode is similar to `AppCdsPrebuild`, but it uses the [JDK 25 AOT feature](https://openjdk.org/jeps/514) which is an enhancement of AppCDS that adds even more information to the class archive. It improves linking and JIT warmup too.

This mode also requires a "training run" of the app before packaging it to create the AOT class archive.
**Requires JDK 25 or later.**

### Advantages:
- Fast startup, including the first run.
- Potentially faster startup than `AppCdsPrebuild`.
- Faster JIT warmup.

### Drawbacks:
 - Requires an additional step of running the app when building the distributable. See the "Training Run" section for details.
 - The distributable is larger because it includes the archive of the app's classes.

## Training Run
In the "Prebuild" modes, the Compose Multiplatform Gradle plugin runs the app (the "Training Run") when packaging it for distribution to create the class archive. When it does this, it sets a system property `compose.aot.training-run` to `true`,  letting the app know that it's being run to create the archive. The app, when given this flag, should cause all the classes it wants to be in the archive to be loaded.

This can be very simple if the developer only wants to optimize startup. In this case, and if building the distributable manually, the developer can just manually quit the app after it is run. If building automatically (e.g. on a CI/CD), the app should check for the flag, wait until it has finished starting up, and exit. Something like this will work for most apps:
```
application {
    ...
    if (System.getProperty("compose.aot.training-run") == "true") {
        LaunchedEffect(Unit) {
            delay(10.seconds)  // Or wait for a custom event indicating the app finished startup
            exitApplication()
        }
    }
}
```

We could also add some support in Compose when running with this flag.
- An `isComposeIdle` state, so the app can wait on that.
- A completely automatic mode when we exit the app when `isComposeIdle` becomes `true` after startup.

Of course the developer may want to include classes other than the ones used on startup. For example, there may be a feature/screen in the app that loads many classes, and therefore takes a long time when used the first time. The app should then "use" this feature (or show the screen) and then quit.

Note that when building the distributable on a CI/CD machine, it will need to be able to run the app, and therefore will need to have a graphical environment. In the future, maybe we could execute the app using `ImageComposeScene` and not require a graphical environment.

# Startup Time comparison on a small project
All values in milliseconds

### Cold start

| Build mode | Shell → Main | Main → Window | Total | Speedup (vs. no AOT) |
|---|---:|---:|---:|---------------------:|
| No AOT | 1133 | 884 | 2017 |                    — |
| AppCDS Prebuild | 48 | 478 | 526 |                 3.8× |
| AOT Prebuild | 67 | 357 | 424 |                 4.8× |

### Hot start (2nd run)

| Build mode | Shell → Main | Main → Window | Total | Speedup (vs. no AOT) |
|---|---:|---:|---:|---------------------:|
| No AOT | 76 | 860 | 936 |                    — |
| AppCDS Prebuild | 46 | 488 | 534 |                 1.8× |
| AOT Prebuild | 69 | 368 | 437 |                 2.1× |

### Cold start

| Build mode | Shell → Main | Main → Window | Total | Speedup (vs. no AOT) |
|---|---:|---:|---:|---------------------:|
| No AOT | 145 | 1211 | 1356 |                    — |
| AppCDS Prebuild | 110 | 718 | 828 |                 1.6× |
| AppCDS Prebuild (relocated*) | 106 | 1111 | 1217 |                 1.1× |
| AOT Prebuild | 130 | 520 | 650 |                 2.1× |
| AOT Prebuild (relocated*) | 402 | 517 | 919 |                 1.5× |

### Hot start (2nd run)

| Build mode | Shell → Main | Main → Window | Total | Speedup (vs. no AOT) |
|---|---:|---:|---:|---------------------:|
| No AOT | 137 | 1193 | 1330 |                    — |
| AppCDS Prebuild | 113 | 688 | 801 |                 1.7× |
| AppCDS Prebuild (relocated*) | 103 | 1092 | 1195 |                 1.1× |
| AOT Prebuild | 126 | 530 | 656 |                 2.0× |
| AOT Prebuild (relocated*) | 391 | 526 | 917 |                 1.5× |

*Executed in a such a way that the jars at the classpath with which the app was packages don't exist.
This is important due to https://youtrack.jetbrains.com/issue/JBR-9098/AppCDS-writes-local-paths-into-.jsa-archive

Fixes https://youtrack.jetbrains.com/issue/CMP-8338/Support-AppCDS-in-Compose-for-Desktop
Fixes https://youtrack.jetbrains.com/issue/CMP-10425/AOT-Compilation-for-desktop-distributables

## Testing
Tested manually on a demo app.
The app is at https://github.com/m-sasha/DesktopPlaygroundAot

This should be tested by QA

### Highlights - Desktop
- Implemented support for [AppCDS](https://dev.java/learn/jvm/cds-appcds/) and [AOT](https://openjdk.org/jeps/514), which can significantly speed up application startup.


m-sasha:
I haven't (re-)tested this on a demo app yet, but the tests seem to pass locally.

m-sasha:
There is an [issue](https://youtrack.jetbrains.com/issue/JBR-9098/AppCDS-writes-local-paths-into-.jsa-archive) remaining from the previous attempt to support AppCDS where the JVM "looks at" the class files on the classpath even though they are present in the archive. On Windows this results in failing file lookups which take a long time and this negates a large part of the benefit of AppCDS (but it's still faster than without AppCDS).

I'm still in conversation with the IJ/JBR team to hopefully fix this one day, but for now I guess we'll have to live with it.

terrakok:
to make the CI happy, please, update the compose version in `gradle.properties` to [1.12.0-beta02](https://github.com/JetBrains/compose-multiplatform/releases/tag/untagged-dc53e7807fa7f00c6225)

terrakok (approved):

terrakok:
CI will be red. https://github.com/JetBrains/compose-multiplatform/pull/5651

terrakok:
macos fails unrelated. ignore them, I'm working on it

m-sasha:
> macos fails unrelated. ignore them, I'm working on it

Maybe https://github.com/JetBrains/compose-multiplatform/pull/5646 can fix them ?

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/AotSettings.kt
```diff
@@ -0,0 +1,295 @@
+package org.jetbrains.compose.desktop.application.dsl
```
igordmn, line 1: > On Windows this results in failing file lookups which take a long time and this negates a large part of the benefit of AppCDS

We should mention this in KDocs, with the link to https://youtrack.jetbrains.com/issue/JBR-9098/AppCDS-writes-local-paths-into-.jsa-archive, so users also can vote for it

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/AotSettings.kt
```diff
@@ -0,0 +1,295 @@
...
+        override val needsTrainingRun: Boolean get() = false
+        override val generateJreClassesArchive: Boolean get() = false
+        override fun runtimeJvmArgs(configuration: AotConfiguration) = emptyList<String>()
+        @Suppress("unused") private fun readResolve(): Any = None
+    }
+
+    /**
+     * The base class for the AppCDS modes introduced in JDK 21.
+     */
+    abstract class AppCdsMode(name: String) : AotMode(name) {
```
igordmn, line 108: Will CDS still be useful if we have AOT?

Besides that it works with a lower JDK, but usually it is no issue to update JDK for packaging.

m-sasha: Not that I'm aware of.

But JDK 25 is new enough that I think we should support JDK 21.

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/AotSettings.kt
m-sasha: Done

igordmn (approved):

AChep:
There seem to be a regression where `createReleaseDistributable` no longer has any declared outputs (the internal `createReleaseDistributableImpl` does). @m-sasha @igordmn

m-sasha:
> There seem to be a regression where `createReleaseDistributable` no longer has any declared outputs (the internal `createReleaseDistributableImpl` does). Alexander Maryanovsky (@m-sasha) Igor Demin (@igordmn)

Is there a visible effect of this?
`createReleaseDistributable` should depend on `createReleaseDistributableImpl`.

AChep:
Yes, it breaks custom packaging tasks that consume its outputs. I have a tar task using:

```
tasks.register<Tar>("packageReleaseDistributable") {
    from(tasks.named("createReleaseDistributable"))
    // Additional Linux metadata files…
}
```

After this change, the tarball does not contain the output of `createReleaseDistributable` as it doesn’t automatically include the outputs of its dependencies.

m-sasha:
Thanks for the report. I created a ticket here: https://youtrack.jetbrains.com/issue/CMP-10846/Restore-outputs-of-createDistributable-tasks

PR #5651, opened 2026-07-15
Update compose.version to 1.12.10-alpha01+dev4472

terrakok:
## Release Notes
N/A

### DIFF
--- a/gradle-plugins/gradle.properties
+++ b/gradle-plugins/gradle.properties
@@ -8,7 +8,7 @@ kotlin.code.style=official
 dev.junit.parallel=false
 
 # Default version of Compose Libraries used by Gradle plugin
-compose.version=1.12.10-alpha01+dev4445
+compose.version=1.12.10-alpha01+dev4472
 compose.material3.version=1.9.0
 # The latest version of Kotlin compatible with compose.tests.compiler.version. Used only in tests/CI.
 compose.tests.kotlin.version=2.3.20
--- a/components/gradle.properties
+++ b/components/gradle.properties
@@ -9,7 +9,7 @@ android.useAndroidX=true
 #Versions
 kotlin.version=2.3.20
 agp.version=8.9.0
-compose.version=1.12.10-alpha01+dev4445
+compose.version=1.12.10-alpha01+dev4472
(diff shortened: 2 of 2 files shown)


igordmn (approved):

gradle-plugins/gradle.properties
```diff
@@ -8,7 +8,7 @@ kotlin.code.style=official
 dev.junit.parallel=false
 
 # Default version of Compose Libraries used by Gradle plugin
-compose.version=1.12.10-alpha01+dev4445
+compose.version=1.12.0-beta02
```
igordmn: What is the reason for downgrading it?

Generally it is better to keep the latest version here to make sure that there are no regressions

terrakok: this is not downgrade. actually 1.12.0-beta02 is latter

terrakok: and 1.12.0-beta02 has the new format of the web runtime puplication

igordmn: `1.12.0-beta02` is built from the release branch, so it can have fewer fixes. Can we use the latest `1.12.10-alpha01+dev4472` from the main branch?

igordmn: After thinking about it, maybe keeping `1.12.0-beta02` is better because we can't use a new API from `1.12.10`, `1.13.0` accidentally.

There is an issue that we don't run tests with the latest Compose, but maybe the solution is just running them separately, not hardcoding the latest version here.

So, in the end, `1.12.0-beta02` looks good for me.

igordmn: The failed tests say that we already use something new in the Gradle plugin, or we introduced a breaking change between `1.12.10-alpha01+dev4445` and  `1.12.0-beta02`

terrakok: oh. you are right. it should be 1.12.10-alpha01+dev4472

terrakok: there is the new format of the web runtime puplication in the compose core main branch

terrakok: it is expected that web tests fail when a new gradle plugin + an old compose are used

PR #5652, opened 2026-07-15
Fix "Desktop release build fails with VerifyError: Bad return type"

igordmn:
Fixes https://youtrack.jetbrains.com/issue/CMP-10488/Desktop-release-build-fails-with-VerifyError-Bad-return-type

## Testing
The proguard test passes with the new fix

### Fixes - Desktop
- _(prelease fix)_ Fix "Desktop release build fails with VerifyError: Bad return type"


m-sasha (approved):
Let's create a ticket to revert this when ProGuard fix the bug?

igordmn:
https://youtrack.jetbrains.com/issue/CMP-10503/Revert-ProGuard-rule-for-JvmMultifileClass

PR #5658, opened 2026-07-20
Run full Gradle checks

igordmn:
GitHub checks miss validation issues because they don't call `validatePlugins`:
```
Configuration cache entry stored.
Execution failed for task ':compose:validatePlugins' (registered by plugin 'org.gradle.java-gradle-plugin').
> Plugin validation failed with 1 problem:
    - Error: Type 'org.jetbrains.compose.desktop.application.tasks.AbstractCreateAotArchiveTask' property 'dependencyFiles$compose' is annotated with @InputFiles but missing a normalization strategy.
```
(The fix in https://github.com/JetBrains/compose-multiplatform/pull/5657)
## Release Notes
N/A

### DIFF
--- a/.github/workflows/gradle-plugin.yml
+++ b/.github/workflows/gradle-plugin.yml
@@ -60,9 +60,14 @@ jobs:
 
       - name: Print summary
         shell: bash
-        working-directory: gradle-plugins/compose/build/test-summary
+        working-directory: gradle-plugins
         if: always()
         run: |
+          if [[ ! -d compose/build/test-summary ]]; then
+            exit 0
+          fi
+
+          cd compose/build/test-summary
           for SUMMARY_FILE in `find . -name "*.md"`; do
             FILE_NAME=`basename $SUMMARY_FILE`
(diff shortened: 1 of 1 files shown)


igordmn:
The executing time is comparable to the previous runs

terrakok (approved):

.github/workflows/gradle-plugin.yml
```diff
@@ -49,8 +49,7 @@ jobs:
         shell: bash
         working-directory: gradle-plugins
         run: |
-          ./gradlew assemble
-          ./gradlew --continue :preview-rpc:test :compose:test ':compose:test-Gradle[${{ matrix.gradle }}]-Agp[${{ matrix.agp }}]'
```
MatkovIvan, line 53: Removing `--continue` will report only the first issue

igordmn: https://github.com/JetBrains/compose-multiplatform/pull/5661

PR #5668, opened 2026-07-23
Update currentOs deprecation warning

hub-bla:
This PR updates the deprecation warning for `currentOs` as the `desktop-jvm-all` artifact will no longer exist. Instead, `ui-skiko` will provide the Skiko runtime fat jar. As a result, the deprecation message should only communicate that `currentOs` is going to be removed

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
@@ -132,11 +132,7 @@ abstract class ComposePlugin : Plugin<Project> {
         @Deprecated("Specify dependency directly", replaceWith = ReplaceWith("\"org.jetbrains.compose.ui:ui-test-junit4:${ComposeBuildConfig.composeVersion}\""))
         val uiTestJUnit4 get() = composeDependency("org.jetbrains.compose.ui:ui-test-junit4")
 
-        @Deprecated(
-            "Use org.jetbrains.compose.desktop:desktop-jvm-all instead. " +
-                "It works for every desktop OS; packaging drops unrelated OS binaries.",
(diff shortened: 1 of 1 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -132,11 +132,7 @@ abstract class ComposePlugin : Plugin<Project> {
         @Deprecated("Specify dependency directly", replaceWith = ReplaceWith("\"org.jetbrains.compose.ui:ui-test-junit4:${ComposeBuildConfig.composeVersion}\""))
         val uiTestJUnit4 get() = composeDependency("org.jetbrains.compose.ui:ui-test-junit4")
 
-        @Deprecated(
-            "Use org.jetbrains.compose.desktop:desktop-jvm-all instead. " +
-                "It works for every desktop OS; packaging drops unrelated OS binaries.",
-            replaceWith = ReplaceWith("\"org.jetbrains.compose.desktop:desktop-jvm-all:${ComposeBuildConfig.composeVersion}\"")
-        )
+        @Deprecated("currentOs is deprecated and will be removed in next releases.")
```
MatkovIvan: It's not clear that users are supposed to just delete it and it will work fine

hub-bla: Does adding something like "There is no replacement and it can be safely removed" enough?

MatkovIvan: I guess it's better to explain more details like previously - "packaging drops unrelated OS binaries"

hub-bla: Updated

MatkovIvan (approved):
Please make sure that CI is green before merging

PR #5673, opened 2026-07-24
Benchmarks: register the Skiko implementation before creating a scene

MatkovIvan:
[CMP-10555](https://youtrack.jetbrains.com/issue/CMP-10555) Benchmarks run failure

## Release Notes
N/A

### DIFF
--- a/benchmarks/multiplatform/compose-scene-impl-3/src/androidMain/kotlin/BenchmarkComposeSceneImpl.kt
+++ b/benchmarks/multiplatform/compose-scene-impl-3/src/androidMain/kotlin/BenchmarkComposeSceneImpl.kt
@@ -1,0 +1,14 @@
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.Canvas
+
+/**
+ * Fake (no-op) implementation of [BenchmarkComposeScene] for Android.
+ */
+class BenchmarkComposeSceneImpl(width: Int, height: Int) : BenchmarkComposeScene {
+    override fun setContent(content: @Composable () -> Unit) {}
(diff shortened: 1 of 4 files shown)


benchmarks/multiplatform/compose-scene-impl-3/src/androidMain/kotlin/BenchmarkComposeSceneImpl.kt
pjBooms, line 1: Won't we have problems with sym-links on Windows?

MatkovIvan: As far as I remember, we've already used it previously
cc @igordmn

igordmn: We use them in the core repo and they work on [Windows](https://github.com/JetBrains/compose-multiplatform-core/blob/ea4a895ac39061e0b6ab73823c563c8876fcfe2d/MULTIPLATFORM.md?plain=1#L5).

But personally I don't like them, it makes navigation and editing confusing.

Is there a way to implement this without them? For example, as usual, just moving it into a common module/package

pjBooms (approved):

benchmarks/multiplatform/compose-scene-impl-3/src/androidMain/kotlin/BenchmarkComposeSceneImpl.kt
MatkovIvan: Removed

PR #5676, opened 2026-07-27
Remove prerelease notes for 1.5.0...1.11.0

MatkovIvan:
## Release Notes
N/A


mazunin-v-jb (approved):

igordmn:
If the purpose is to not pollute the main Changelog, maybe is it better to move them into CHANGELOG_PRERELEASE? This changelog is still useful.

And modify the script/docs.

MatkovIvan:
> This changelog is still useful.

Since we're duplicating everything in the stable release notes AND have this written down in GitHub releases page AND have them written down in git history, could you please clarify for what purpose they are useful?

igordmn:
We are duplicating, but we don't preserve the versions in which a change appeared.

The stable changelog is for people who wants to know the difference between stable versions.

The prerelease changelog is for people who wants to know the difference between prerelease versions. There are use cases, for example:
- switching from 1.7.0-alpha02 to 1.8.0-beta01 - want to know exactly what changed
- downgrading to 1.7.0-beta02 from 1.9.0, because of some regression

kropp:
I understand the use case, but I question how real and widespread is it.

igordmn:
I consider this is at least comparable to using prerelease versions itself. If we consider them useful, full prerelease changelog is also useful.

Also, with removing old prerelease changelog, there are more questions:
- when to remove? We can't remove with a stable version, because people who use non-stable still want to know what changed in the stable
- how it can be automated or watched for the correct process?

With keeping prerelease in a separate file, the question is simpler - in which file the version belongs.

PR #5677, opened 2026-07-28
Resources: read resources on an interrupted thread without a crash

terrakok:
`stringResource` and other composable resource getters bridge a suspend loading function to a synchronous call. `runBlocking` did that bridging, but `BlockingCoroutine.joinBlocking` checks the interruption flag of the calling thread before it checks whether the coroutine is already
completed. So with an interrupted composition thread any access to a Compose resource threw an `InterruptedException` - even a cache hit, whose value was already computed.

Replace `runBlocking` in the blocking `rememberResourceState` with a minimal `runResourceBlocking` bridge:
- a block which doesn't suspend (a cache hit) is executed on the calling thread and returns its result directly, without any parking and without touching the interruption flag at all;
- if the block really suspends, the thread waits in `LockSupport.park`, which never throws an `InterruptedException`. The flag is taken away for the time of waiting, otherwise `park` returns immediately and the wait turns into a busy spin, and it is given back to its real addressee afterwards.

Native targets keep `runBlocking` - there is no thread interruption there. `ResourceCaches.clearBlocking` on desktop uses the new bridge too.

Fixes https://youtrack.jetbrains.com/issue/CMP-6615

## Testing
New unit tests

### Fixes - Android
- Fixed rare `InterruptedException` crashes when reading resources during composition on an interrupted thread.

### DIFF
--- a/components/resources/library/src/blockingMain/kotlin/org/jetbrains/compose/resources/ResourceState.blocking.kt
+++ b/components/resources/library/src/blockingMain/kotlin/org/jetbrains/compose/resources/ResourceState.blocking.kt
@@ -1,7 +1,6 @@
 package org.jetbrains.compose.resources
 
 import androidx.compose.runtime.*
-import kotlinx.coroutines.runBlocking
 
 @Composable
 internal actual fun <T> rememberResourceState(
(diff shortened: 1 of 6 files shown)


igordmn:
@pjBooms, could you review this?

pjBooms (approved):
> Fixed rare android ANR errors during the resources read.

The PR does not actually fixes ANRs (they can't be actually fixed by their nature) but fixes app crashes when a thread is interrupted during resource reading. So I would rephrase the release note to something like:

Fixed rare InterruptedException crashes when reading resources during composition on an interrupted thread.

PR #5678, opened 2026-07-29
Imageviewer example improvements

SebastianAigner:
This PR refreshes the Imageviewer example project.

That includes:
- CMP resources replace home-baked solution
- Nav3
- A proper camera abstraction that has a more compose-like API rather than just three platform specific screens
- Dark theme
- Concurrency fixes for both app and map widget
- Scroll fixes
- Layout corrections

It also _downgrades_ AGP to a version that's actually compatible with IntelliJ, so people aren't constrained to Android Studio (9.1.x).

(delivered as one big pull request as per @kropp's request 😁) 

## Release Notes
N/A


examples/imageviewer/shared/src/webMain/kotlin/example/imageviewer/utils/UUID.web.kt
```diff
@@ -1,5 +1,6 @@
 package example.imageviewer.utils
 
+@OptIn(ExperimentalWasmJsInterop::class)
 @JsModule("uuid")
```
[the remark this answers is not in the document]
SebastianAigner: IDE reports no imports required here.

kropp:
image
Image description styling is not adjusted to the dark theme

examples/imageviewer/kotlin-js-store/wasm/yarn.lock
kropp, line 1:
```
* What went wrong:
Execution failed for task ':kotlinWasmStoreYarnLock' (registered by plugin class 'org.jetbrains.kotlin.gradle.targets.wasm.yarn.WasmYarnPlugin').
> Lock file was changed. Run the `kotlinWasmUpgradeYarnLock` task to actualize lock file
```
I tried to run it on Wasm target in IJ and got this. Removing the file regenerated it without `format-utils`

SebastianAigner: updated, looks like that slipped my final checks!

examples/imageviewer/shared/src/commonMain/kotlin/example/imageviewer/NameAndDescription.common.kt
```diff
@@ -16,23 +21,18 @@ class NameAndDescription(
...
-    return remember {
-
-        Clock.System.now().toLocalDateTime(TimeZone.UTC)
+    val kotlinConfName = stringResource(Res.string.kotlin_conf_name)
+    val kotlinConfDescription = stringResource(Res.string.kotlin_conf_description, getCurrentPlatform())
+    val newPhotoName = stringResource(Res.string.new_photo_name)
+    val newPhotoDescription = stringResource(Res.string.new_photo_description)
+    return remember(kotlinConfName, kotlinConfDescription, newPhotoName, newPhotoDescription) {
         val kotlinConfEndTime =
             LocalDateTime(2023, Month.APRIL, 14, hour = 23, minute = 59).toInstant(TimeZone.UTC)
```
kropp: Is it still actual?

kropp:
Do we need a separate JS app?

SebastianAigner:
Rendering issue in dark mode is now fixed.

Screenshot 2026-07-29 at 17 28 51

SebastianAigner:
@kropp re. separate JS app: I didn't touch this part here; if you prefer only having the wasm part, I'm happy to remove the JS app bits as part of this PR. I personally don't have a need for it 😁

examples/imageviewer/shared/src/commonMain/kotlin/example/imageviewer/NameAndDescription.common.kt
SebastianAigner: I kept it in there for old times sake, but unless you turn your clock back, you certainly won't hit that branch. I'll remove it :)

SebastianAigner: removed. Also let me remove some functions that were only used by this path!

kropp:
@SebastianAigner it should be possible to have a single `index.html` and `Main.kt` for both JS and Wasm. So, the suggestion is to retain the JS target, but remove all unneeded files

SebastianAigner:
@kropp Moved to a common entry point (had to remove `onWasmReady` since that's not available in the shared source set, but looks like it works nonetheless). I've also unified and cleaned up the two `index.html` files.

I got misled by https://kotlinlang.org/docs/multiplatform/whats-new-compose-170.html#web into thinking skiko.js is still required; turns out its not. Unfortunately not mentioned in the What's New Pages, I'll prep a PR that retroactively adds it to the 1.9.3 What's New Page.

SebastianAigner:
(PR: https://github.com/JetBrains/kotlin-multiplatform-dev-docs/pull/753)

kropp (approved):

PR #5679, opened 2026-07-29
Update Compose version and remove desktop-jvm-all dependencies

hub-bla:
`desktop-jvm-all `no longer exists as the Skiko runtime is now pulled transitively through `ui-skiko`

## Release Notes
N/A

### DIFF
--- a/components/AnimatedImage/demo/build.gradle.kts
+++ b/components/AnimatedImage/demo/build.gradle.kts
@@ -13,7 +13,7 @@ kotlin {
             }
         }
         jvmMain.dependencies {
-            implementation(libs.compose.desktop.jvm)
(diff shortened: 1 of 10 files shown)


igordmn:
> desktop-jvm-all

Does `compose.desktop.currentOs` still work? Won't existing users of stable versions be affected?

igordmn (approved):

hub-bla:
Yes, `currentOs` still works and is now deprecated. It still points to the desktop OS-specific artifacts that we continue to publish. The only thing that changes there is that these desktop OS-specific artifacts no longer contain the Skiko runtime themselves and instead they pull it in transitively from `ui-skiko`

terrakok (approved):

PR #5682, opened 2026-07-31
Resolve Swift compatibility libraries at iOS link time

svastven:
Resolve the active Xcode toolchain’s Swift compatibility-library directory lazily when an iOS consumer links.

relates to https://github.com/JetBrains/compose-multiplatform-core/pull/3274

Fixes [CMP-10607](https://youtrack.jetbrains.com/issue/CMP-10607) Resolve Swift compatibility libraries from the consumer’s active Xcode during iOS linking

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/SwiftCompatibilityLinking.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/SwiftCompatibilityLinking.kt
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2026 JetBrains s.r.o. and respective authors and developers.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package org.jetbrains.compose
+
+import org.gradle.api.Project
+import org.jetbrains.compose.internal.KOTLIN_MPP_PLUGIN_ID
+import org.jetbrains.compose.internal.mppExt
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTarget
+import org.jetbrains.kotlin.gradle.plugin.mpp.NativeBinary
+import org.jetbrains.kotlin.konan.target.KonanTarget
+import java.io.File
+
+/** Configures iOS linker tasks to resolve the active Xcode's Swift compatibility-library path at link time. */
+internal fun Project.configureSwiftCompatibilityLinking() {
+    plugins.withId(KOTLIN_MPP_PLUGIN_ID) {
+        mppExt.targets.withType(KotlinNativeTarget::class.java).all { target ->
+            target.configureSwiftCompatibilityLinking()
+        }
+    }
+}
+
+private fun KotlinNativeTarget.configureSwiftCompatibilityLinking() {
+    if (System.getProperty("os.name") != "Mac OS X") return
+
+    val sdkName =
+        when (konanTarget) {
+            KonanTarget.IOS_ARM64 -> "iphoneos"
+            KonanTarget.IOS_X64,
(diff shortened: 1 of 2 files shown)


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/SwiftCompatibilityLinking.kt
```diff
@@ -0,0 +1,58 @@
...
+internal fun Project.configureSwiftCompatibilityLinking() {
+    plugins.withId(KOTLIN_MPP_PLUGIN_ID) {
+        mppExt.targets.withType(KotlinNativeTarget::class.java).all { target ->
+            target.configureSwiftCompatibilityLinking()
+        }
+    }
+}
+
+private fun KotlinNativeTarget.configureSwiftCompatibilityLinking() {
+    if (System.getProperty("os.name") != "Mac OS X") return
```
terrakok, line 26: When Gradle caches the configuration phase, it must be aware of all environment variables and system properties that affect the build.

There is a proper API:
`if (!OperatingSystem.current().isMacOsX) return`

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/SwiftCompatibilityLinking.kt
```diff
@@ -0,0 +1,58 @@
...
+            .exec { spec -> spec.commandLine("xcrun", "--find", "swiftc") }
+            .standardOutput
+            .asText
+            .map { swiftcPath ->
+                File(swiftcPath.trim())
+                    .parentFile
+                    .parentFile
+                    .parentFile
+                    .resolve("usr/lib/swift/$sdkName")
+                    .absolutePath
```
terrakok, line 46: extract and add a comment with an explanation, please.

terrakok (changes requested):

terrakok (commented):
From a gradle code perspective the PR seems good. Regarding linker flags I don't have a proper knowledge

terrakok (approved):
From a gradle code perspective the PR seems good. Regarding linker flags I don't have a proper knowledge

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/SwiftCompatibilityLinking.kt
```diff
@@ -0,0 +1,55 @@
...
+            .exec { spec -> spec.commandLine("xcrun", "--sdk", sdkName, "--show-toolchain-path") }
+            .standardOutput
+            .asText
+            .map { toolchainPath ->
+                File(toolchainPath.trim())
+                    .resolve("usr/lib/swift/$sdkName")
+                    .absolutePath
+            }
+
+    binaries.withType(NativeBinary::class.java).all { binary ->
```
MatkovIvan: As far as I know, KGP provides kotlin friendly generics for such things (instead of `::class.java`)

MatkovIvan (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/SwiftCompatibilityLinking.kt
svastven: Updated, thank you

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/SwiftCompatibilityLinking.kt
svastven: Updated

PR #5684, opened 2026-07-31
Fix `skikoWebRuntime` resolve in a configuration phase when `whenTaskAdded` is used.

terrakok:
Fix `skikoWebRuntime` resolve in a configuration phase.

Fixes https://youtrack.jetbrains.com/issue/CMP-10608

## Testing
Added a new test for the case.

## Release Notes
N/A


MatkovIvan:
> Fixes - Gradle Plugin
> Fix skikoWebRuntime resolve in a configuration phase when whenTaskAdded is used.

Do we already have a published version with this bug? If no - please remove it. If it was only published as prerelease (alpha/beta), add a "prerelease fix" note

eymar (approved):

PR #5685, opened 2026-07-31
Refactor Web configuration: don't use afterEvaluate.

terrakok:
Refactor Web configuration: don't use afterEvaluate.

And delete deprecated Compose Gradle Plugin API: `compose.web.targets(...)`

### Breaking Changes - Gradle Plugin
- Deleted deprecated Compose Gradle Plugin API: `compose.web.targets(...)`


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/WebExtension.kt
```diff
@@ -5,75 +5,6 @@
...
-                )
-                emptySet()
-            }
-        }
-
-        return emptySet()
-    }
-}
\ No newline at end of file
+abstract class WebExtension : ExtensionAware
```
MatkovIvan, line 10: Should we add KDoc for public class?

MatkovIvan (approved):

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/WebExtension.kt
terrakok: That was there quite some time. I would delete it completely but someone may use it as a context for extensions

eymar (approved):

PR #5686, opened 2026-08-03
Fix packaging on Linux

m-sasha:
Pass `--app-image` and `--icon` correctly.

Before https://github.com/JetBrains/compose-multiplatform/pull/5644 the packaging for Linux was done in one step. Now the packaging is always done in [two steps](https://docs.oracle.com/en/java/javase/17/jpackage/image-and-runtime-modifications.html): first create an "app image" and then from that an installer.

The packaging task was not set up correctly for this (on Linux).

Fixes https://youtrack.jetbrains.com/issue/CMP-10546

## Testing
Tested manually and added a check in a unit test

## Release Notes
N/A

### DIFF
--- a/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/DesktopApplicationTest.kt
+++ b/gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/DesktopApplicationTest.kt
@@ -672,14 +672,7 @@ class DesktopApplicationTest : GradlePluginTestBase() {
     @Test
     fun testAppCdsNoneOnJdk17() {
         with(aotProject(AotMode.None, javaVendor = JvmVendor.KnownJvmVendor.AMAZON, javaVersion = 17)) {
-//        fun testRunTask(runTask: String) {
-//            gradleFailure(runTask).checks {
-//                check.logContains("AotMode 'AppCdsAuto' is not supported on JDK earlier than 19; current is 17")
-//            }
-//        }
             gradle(":packageReleaseDmg")
-
-//        testRunTask(":runReleaseDistributable")
         }
     }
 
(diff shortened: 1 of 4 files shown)


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/DesktopApplicationTest.kt
```diff
@@ -658,6 +669,20 @@ class DesktopApplicationTest : GradlePluginTestBase() {
...
+//        fun testRunTask(runTask: String) {
+//            gradleFailure(runTask).checks {
+//                check.logContains("AotMode 'AppCdsAuto' is not supported on JDK earlier than 19; current is 17")
+//            }
+//        }
+            gradle(":packageReleaseDmg")
+
+//        testRunTask(":runReleaseDistributable")
+        }
+    }
```
terrakok, line 676: could you explain this?

terrakok: Generally I'd recommend to avoid such heavy tests. Instead we need to check a task configuration only

m-sasha: Remnants from ancient times; removed.

m-sasha: > Generally I'd recommend to avoid such heavy tests. Instead we need to check a task configuration only

It tests failure, so should pass (fail) quickly.

terrakok (approved):

MatkovIvan:
> Fixes - Desktop
> - Fixed native app packaging on Linux

@m-sasha was it broken in 1.12?

m-sasha:
> @m-sasha was it broken in 1.12?

If I understand correctly, 1.12.10-alpha01+dev4534 is not actually 1.12

MatkovIvan:
@m-sasha it seems not a part of any release, so we do not need to write release notes in this case. Could you please remove it to avoid user's confusion?

m-sasha:
> Alexander Maryanovsky (@m-sasha) it seems not a part of any release, so we do not need to write release notes in this case. Could you please remove it to avoid user's confusion?

I marked it as _(prerelease fix)_; is that enough?

MatkovIvan:
_(prerelease fix)_ is only for cases where we're fixing something that was released as alpha/beta.
For this case it wasn't released at all (we do not count dev builds here), so it should be just N/A here

m-sasha:
Ok, removed

PR #5690, opened 2026-08-10
Refactor changelog script to streamline library version parsing

mazunin-v-jb:
Changes:

- The `### Libraries` table is built from a list of rows instead of a hardcoded markdown template. 
- Libraries from `excludedLibraries` are filtered out of `androidxLibToVersion`. Rows for libraries not published in this release are taken from the previous table in `CHANGELOG.md` (`parsePublishedLibraries`): both our version and the Jetpack one.
- `Navigation` and `WindowManager Core` are no longer hardcoded

Fixes:
https://youtrack.jetbrains.com/issue/CMP-8567/Changelog-script.-Wrong-redirection-version-for-disabled-libraries

## Release Notes
N/A

### DIFF
--- a/tools/changelog/changelog.main.kts
+++ b/tools/changelog/changelog.main.kts
@@ -70,6 +70,9 @@ val argsKeyToValue = args
 
 val token = argsKeyToValue["token"]
 
+val teamcityConfigRepo = "ssh://git@git.jetbrains.team/ui/compose-teamcity-config.git"
+val libraryKtPath = ".teamcity/compose/Library.kt"
+
 // Parse sections from [PR_FORMAT.md]
 fun parseSections(title: String) = prFormatFile
     .readText()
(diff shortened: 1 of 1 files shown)


igordmn:
~~Is this only a refactor without changing the behavior?~~

I see `parsePublishedLibraries` - now it is not hardcoded and read from the CHANGELOG.md. Worth to add into the PR description

tools/changelog/changelog.main.kts
```diff
@@ -70,6 +70,9 @@ val argsKeyToValue = args
```
igordmn, line 70: Looks like this PR fixes https://youtrack.jetbrains.com/issue/CMP-8567/Changelog-script.-Wrong-redirection-version-for-disabled-libraries - add it into the description

igordmn (approved):

tools/changelog/changelog.main.kts
mazunin-v-jb: Thanks! That's exactly what I wanted to fix, added

PR #5693, opened 2026-08-24
Add a common kotlinx-browser subset for Compose HTML

clemenskoprolin:
This PR adds common browser type definitions needed to prepare Compose HTML for server-side rendering.

A generator derives a common DOM API subset from [kotlinx-browser](https://github.com/Kotlin/kotlinx-browser) 0.5.0. Common code receives expect declarations, JS and Wasm/JS preserve browser type identity, and JVM receives type-correct stubs for compilation. The generated sources and API manifest are checked in and verified against fresh output.

The PR also updates the shared Karma reporter path and Yarn lock required by the current browser-test configuration.

Further details are available in `html/kotlinx-browser-common-subset/README.md`.

Fixes [CMP-10656](https://youtrack.jetbrains.com/issue/CMP-10619/Research-the-feasibility-of-adding-k-jvm-target-kotlinx-browser-library)

## Testing
```
./gradlew checkKotlinxBrowserCommonSubset
./gradlew -p kotlinx-browser-common-subset/generator test
./gradlew :kotlinx-browser-common-subset:check
./gradlew :html-core:jsBrowserTest
./gradlew :html-svg:jsBrowserTest
```

## Release Notes
N/A


html/test-utils/conf/karma.config.common.d/patch.js
```diff
@@ -1,6 +1,9 @@
 config.plugins = config.plugins || [];
 config.plugins = config.plugins.filter(it => it !== "kotlin-test-js-runner/karma-kotlin-reporter.js");
-config.plugins.push("../../../../../test-utils/conf/karma-kotlin-runner-decorator/karma-kotlin-reporter-decorated.js");
+config.plugins.push(require("path").resolve(
+    __dirname,
+    "../../../../test-utils/conf/karma-kotlin-runner-decorator/karma-kotlin-reporter-decorated.js"
```
eymar: is this change relevant to this PR? 
Something not working without it?

html/build.gradle.kts
```diff
@@ -20,6 +20,26 @@ kotlinKarmaConfig.rootDir = rootProject.rootDir.toString()
 
 apply<SeleniumDriverPlugin>()
 
+val kotlinxBrowserSubsetGenerator = gradle.includedBuild("kotlinx-browser-common-subset-generator")
+
+tasks.register("generateKotlinxBrowserCommonSubset") {
```
eymar: Were these tasks added here for convenience only? Or is there another reason?

If not, I'd keep them in the subproject.

html/settings.gradle.kts
```diff
@@ -85,6 +89,7 @@ module(":compose-compiler-integration", "compose-compiler-integration")
 module(":compose-compiler-integration-lib", "compose-compiler-integration/lib")
 module(":internal-html-core-runtime", "internal-html-core-runtime")
 module(":html-test-utils", "test-utils")
+module(":kotlinx-browser-common-subset", "kotlinx-browser-common-subset")
```
eymar: Was this (and includeBuild above) added here because html-core will depend on this module?

Since this PR doesn't change html-core yet, can we skip this change for now?

I think that kotlinx-browser-common-subset might have its own settings.gradle (be independent of html project since we plan an independent publication for it)

html/kotlinx-browser-common-subset/generator/verification/src/jvmTest/kotlin/GeneratedApiManifestTest.kt
```diff
@@ -0,0 +1,185 @@
...
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+// Guards the generated API manifest and its checked-in baseline.
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertTrue
+
+/** Verifies that every input declaration has an explicit, reviewable manifest outcome. */
+class GeneratedApiManifestTest {
```
eymar, line 12: This and some other tests here test the generator output, right?

Can they be moved to the verification module? And here `jvmTest` will keep the test checking the runtime behaviour of the generated declarations. WDYT?

html/kotlinx-browser-common-subset/generator/ksp-runner/build.gradle.kts
```diff
@@ -0,0 +1,166 @@
+import org.gradle.api.tasks.Sync
```
eymar, line 1: When seeing this name "runner" I started to wonder what does it run.

WDYT about naming it `ksp-runner`? To provide a bit of context right in the module name?

html/kotlinx-browser-common-subset/README.md
```diff
@@ -0,0 +1,147 @@
...
+- `api-manifest.txt`: every declaration in the selected browser input files, marked `EMITTED` or
+  `EXCLUDED` with a structured reason.
+
+The API manifest is checked in at
+[`api/dom-api-manifest.txt`](api/dom-api-manifest.txt). Generation fails for unaccounted
+declarations or stale exclusions, and `GeneratedApiManifestTest` fails when generated output differs
+from the checked-in baseline.
+
+[`dom-api-exclusions.txt`](generator/src/main/resources/dom-api-exclusions.txt) is reserved for
+specific declaration-level decisions that cannot be expressed by classifier selection.ååå
```
eymar:
```
ååå
```

LLM glitch? 😁

html/kotlinx-browser-common-subset/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Run
```
eymar: Let's move this section higher for convenience? (before "How generation works")

html/kotlinx-browser-common-subset/src/commonMain/kotlin/kotlinx/browser/Interop.kt
```diff
@@ -0,0 +1,93 @@
+/*
```
eymar, line 1: I noticed that the logic generating this file takes more lines than this file itself. 
The logic is hardcoded and seem to not depend on the kotlinx-browser sources.

Does it make sense to hardcode the source code for it without generating it?

clemenskoprolin: I agree, good catch. It would also make that part more readable.

html/test-utils/conf/karma.config.common.d/patch.js
clemenskoprolin: I think it's necessary. This adds HTML's first Wasm/JS browser tests. The old path happened to work for repository-local JS Karma but fails when Wasm/JS uses Karma from the Kotlin tooling directory.

html/kotlinx-browser-common-subset/generator/verification/src/jvmTest/kotlin/GeneratedApiManifestTest.kt
clemenskoprolin: Agreed, I’ll move those and their report helpers into the verification module.

html/test-utils/conf/karma.config.common.d/patch.js
clemenskoprolin: Though after I've split it it will no longer be part of the html build, so it's not needed anymore.

html/kotlinx-browser-common-subset/generator/gradle.properties
```diff
@@ -0,0 +1,3 @@
+kotlin.mpp.applyDefaultHierarchyTemplate=false
```
eymar, line 2:
```suggestion
org.gradle.jvmargs=-Xmx4g
kotlin.mpp.applyDefaultHierarchyTemplate=false
```

I tried to run verification tests and the task failed with OutOfMemory error.
I added -Xmx4g locally. Does the task fail for you too without it?

eymar: Btw, there is no explanation of what `verification` module actually verifies.
Could you please add a brief README.md in `verification/` directory? With relevant gradle tasks

html/kotlinx-browser-common-subset/README.md
```diff
@@ -0,0 +1,148 @@
...
+The generator resolves
+`org.jetbrains.kotlinx:kotlinx-browser:<version>:sources`, unpacks its `webMain` sources under the
+KSP runner's `build/`, and feeds the source set to KSP. Fresh output is staged
+under `generator/ksp-runner/build/generated/kotlinxBrowserCommonSubset`. Reviewed source output is
+checked in under `src/`, so compiling the library does not run KSP or require the sources JAR.
+Generator checks and tests do run KSP and resolve the pinned sources artifact.
+
+## Run
+
+Run all commands in this section from the repository's `html/` directory, which contains the Gradle
```
eymar:
```
from the repository's `html/` directory
```

Not true anymore? Since, from gradle perspective this project is an independent from `html`?

eymar: Could you please update the tasks below accordingly?

html/kotlinx-browser-common-subset/README.md
```diff
@@ -0,0 +1,148 @@
...
+Run the generator tests and the library's multiplatform checks:
+
+```shell
+./gradlew -p kotlinx-browser-common-subset/generator :check
+./gradlew -p kotlinx-browser-common-subset check
+```
+
+## How generation works
+
+1. Gradle resolves and unpacks the pinned Maven sources artifact into the KSP runner's `build/`.
```
eymar:
```suggestion
1. Gradle resolves and unpacks the pinned kotlinx-browser sources artifact into the KSP runner's `build/`.
```

html/kotlinx-browser-common-subset/src/commonMain/kotlin/kotlinx/browser/dom/OptionDictionaries.kt
```diff
@@ -0,0 +1,520 @@
...
+  capture: Boolean? = null,
+): AddEventListenerOptions
+
+public expect fun AssignedNodesOptions(flatten: Boolean? = null): AssignedNodesOptions
+
+public expect fun BoxQuadOptions(box: CSSBoxType? = null, relativeTo: JsAny? = null): BoxQuadOptions
+
+public expect fun CanvasRenderingContext2DSettings(alpha: Boolean? = null): CanvasRenderingContext2DSettings
+
+public expect fun EventInit(
```
eymar, line 310: I wonder if we need such `fun *Init(...)` in the common.

I see only one usage of this fun EventInit:

```
public actual open class Event actual constructor(
  type: String,
  eventInitDict: EventInit = EventInit(),
)
```

And in this case it could use `JvmEventInit(null, null, null)`.

WDYT about it and similar `Init` cases?

html/kotlinx-browser-common-subset/src/commonMain/kotlin/kotlinx/browser/dom/OptionDictionaries.kt
```diff
@@ -0,0 +1,520 @@
...
+import kotlinx.browser.fetch.RequestCredentials
+
+@Suppress("EXPECT_ACTUAL_IR_INCOMPATIBILITY")
+public expect interface EventListenerOptions : JsAny {
+  public var capture: Boolean?
+}
+
+@Suppress("EXPECT_ACTUAL_IR_INCOMPATIBILITY")
+public expect interface AddEventListenerOptions : EventListenerOptions, JsAny {
+  public var passive: Boolean?
```
eymar, line 30: What if we keep such dictionaries (Options, Init, etc) empty - without members in the commonMain and jvmMain?
I haven't checked all, but seems like they are not used in jvmMain. 

I think it won't affect compose-html ssr use case, right?

html/kotlinx-browser-common-subset/generator/gradle.properties
clemenskoprolin: Good idea. For me, it works without -Xmx4g, but I think it's good to add for stability anyway.

html/kotlinx-browser-common-subset/src/commonMain/kotlin/kotlinx/browser/dom/OptionDictionaries.kt
clemenskoprolin: Yes, only for SSR we don't need it. But compose-html isn't the only consumer of this API: When we move DOMScope.scopeElement to commonMain, users write raw DOM calls in common code against exactly this.  E.g. `attachShadow(ShadowRootInit(...))` and the options overloads of `addEventListener ` require them.

html/kotlinx-browser-common-subset/src/commonMain/kotlin/kotlinx/browser/dom/OptionDictionaries.kt
clemenskoprolin: Same here: They wouldn't be necessary only for SSR, but some behaviour would change. This for example wouldn't work anymore in common:
```kotlin
val options = MutationObserverInit()
options.childList = true

if (observeAttributes) {
    options.attributes = true
    options.attributeOldValue = true
}

observer.observe(element, options)
```
But if we don't want to support that in common code, we could remove it.

html/kotlinx-browser-common-subset/README.md
```diff
@@ -0,0 +1,181 @@
...
+checked in under `src/`, so compiling the library does not run KSP or require the sources JAR.
+Generator checks and tests do run KSP and resolve the pinned sources artifact.
+
+## Run
+
+Run all commands in this section from this directory. Generate staged output without changing 
+checked-in sources:
+
+```shell
+../gradlew generateKotlinxBrowserCommonSubset
```
eymar: Since this module is independent, let's add gradle/wrapper and gradlew files to git?
then it will be easier to run the tasks.

html/kotlinx-browser-common-subset/src/commonMain/kotlin/kotlinx/browser/dom/OptionDictionaries.kt
eymar: I see. 
Let's keep it then.

eymar (approved):

PR #5694, opened 2026-08-25
Allow excluding libraries from the dependency compatibility check

MatkovIvan:
[CMP-10512](https://youtrack.jetbrains.com/issue/CMP-10512) Provide a way to disable skiko compatibility warning for specific libraries

Currently, the check could only be turned off as a whole, via `org.jetbrains.compose.library.compatibility.check.disable`. A project that knowingly tolerates one mismatched dependency had to give up the check for every other dependency too.

This PR adds a `compose` extension:
```kt
compose {
    dependencyCompatibility {
        enabled = false // default: !org.jetbrains.compose.library.compatibility.check.disable
        exclude("com.example") // whole group
        exclude("org.jetbrains.skiko:skiko") // single module
    }
}
```

### Features - Gradle Plugin
- Allow excluding libraries from the dependency compatibility check via `compose { dependencyCompatibility { } }` extension

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/DependencyCompatibilityExtension.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/DependencyCompatibilityExtension.kt
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2020-2026 JetBrains s.r.o. and respective authors and developers.
+ * Use of this source code is governed by the Apache 2.0 license that can be found in the LICENSE.txt file.
+ */
+
+package org.jetbrains.compose
+
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import org.gradle.api.provider.ProviderFactory
+import org.gradle.api.provider.SetProperty
+import org.jetbrains.compose.desktop.application.internal.ComposeProperties
+import org.jetbrains.compose.internal.utils.property
+import javax.inject.Inject
+
+/**
+ * Configuration of the compatibility check for Compose Multiplatform runtime dependencies.
+ *
+ * The check warns when resolved versions of Compose Multiplatform libraries don't match the Gradle plugin version,
+ * or when Skiko is resolved to a version that is incompatible with the requested one.
+ */
+abstract class DependencyCompatibilityExtension @Inject constructor(
+    objects: ObjectFactory,
+    providers: ProviderFactory,
+) {
+    /**
+     * Whether the compatibility check runs.
+     *
+     * Default is `true`, or `false` when `org.jetbrains.compose.library.compatibility.check.disable=true`
+     * is set in `gradle.properties`.
+     */
+    val enabled: Property<Boolean> = objects.property<Boolean>().convention(
+        ComposeProperties.disableLibraryCompatibilityCheck(providers).map { disabled -> !disabled }
+    )
+
+    internal val excludedModules: SetProperty<String> = objects.setProperty(String::class.java)
+
+    /**
+     * Excludes a library from the compatibility check.
+     *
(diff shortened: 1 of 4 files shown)


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/RuntimeLibrariesCompatibilityCheckTest.kt
```diff
@@ -130,6 +225,8 @@ class RuntimeLibrariesCompatibilityCheckTest : GradlePluginTestBase() {
     }
 }
 
+private const val CHECK_CONFIGURATION_MARKER = "// compatibility check configuration"
```
terrakok: a top level const for a single function. I'd move it to the local val/

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/RuntimeLibrariesCompatibilityCheckTest.kt
```diff
@@ -91,6 +91,101 @@ class RuntimeLibrariesCompatibilityCheckTest : GradlePluginTestBase() {
...
+                configurations.named("jvmRuntimeClasspath") {
+                    resolutionStrategy.force("$OLD_SKIKO_DEPENDENCY")
+                }
+            """.trimIndent()
+        }
+        checkJvmMainCompatibility(logMsg, warningExpected = true)
+
+        // Compose foundation requests Skiko as well, so excluding Compose UI alone doesn't silence the warning.
+        configureCompatibilityCheck(excludes = arrayOf("org.jetbrains.compose.ui"))
+        checkJvmMainCompatibility(logMsg, warningExpected = true)
```
terrakok: why do we need to check this resolution case?

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/RuntimeLibrariesCompatibilityCheckTest.kt
```diff
@@ -91,6 +91,101 @@ class RuntimeLibrariesCompatibilityCheckTest : GradlePluginTestBase() {
...
+
+        configureCompatibilityCheck(
+            excludes = arrayOf("org.jetbrains.compose.ui", "org.jetbrains.compose.foundation")
+        )
+        checkJvmMainCompatibility(logMsg, warningExpected = false)
```
terrakok: the test will fail if we add some different transitive dependencies. I think the scenario is not needed here because it tests unrelated logic in general

gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/DependencyCompatibilityExtension.kt
```diff
@@ -0,0 +1,65 @@
...
+import org.jetbrains.compose.internal.utils.property
+import javax.inject.Inject
+
+/**
+ * Configuration of the compatibility check for Compose Multiplatform runtime dependencies.
+ *
+ * The check warns when resolved versions of Compose Multiplatform libraries don't match the Gradle plugin version,
+ * or when Skiko is resolved to a version that is incompatible with the requested one.
+ */
+abstract class DependencyCompatibilityExtension @Inject constructor(
```
terrakok, line 22: don't we mark any new API as experimental?

MatkovIvan: It's not a default rule. If we're sure of its shape and do not expect changes, we shouldn't add it.

terrakok (approved):

igordmn (approved):

PR #5701, opened 2026-09-01
Fix a skiko runtime when a skiko dependency is added in a test source set only

terrakok:
Fixes https://youtrack.jetbrains.com/issue/CMP-8235

### Fixes - Web
- Fix a skiko runtime when a skiko dependency is added in a test source set only


eymar (approved):

eymar (approved):

MatkovIvan:
> Fix a skiko runtime when a skiko dependency is added in a test source set only

Was it broken 1.12? If it's only about 1.13.0-alpha01, please add `prerelease fix` note

terrakok:
> > Fix a skiko runtime when a skiko dependency is added in a test source set only
> 
> Was it broken 1.12? If it's only about 1.13.0-alpha01, please add `prerelease fix` note

yes. and there is a workaround. so, it is not critical

PR #5703, opened 2026-09-07
Commonize Compose HTML core for JVM

clemenskoprolin:
This PR prepares Compose HTML for SSR on JVM.

Move Compose HTML's element declarations, attributes, events, and CSS APIs into common source sets so they can be shared with JVM code.

This PR provides the common/JVM foundation only. JVM TagElement, Text, and DOM element creation remain unsupported placeholders.

The implementation introduces shared DOM declarations through kotlinx-browser-common-subset, JVM CSS value implementations, and common runtime scope types. Browser rendering remains in the JS implementation. Adds `org.jetbrains.compose.html:kotlinx-browser-common-subset` as a runtime dependency.

Involves [CMP-10656](https://youtrack.jetbrains.com/issue/CMP-10656/Integrate-kotlinx-browser-common-subset-in-compose-html-core)

## Public API/ABI changes for JS consumers:

- `GridAutoFlow`: changes from an object to an interface with companion constants. Expressions such as GridAutoFlow.Row remain supported. Imports must change to `GridAutoFlow.Companion.Row`. The singleton itself can no longer be used as a GridAutoFlow value. 
- CSS unit constants: Values now use typed singleton objects instead of raw JS strings, changing runtime representation, type checks, and equality behavior. Previously compiled inline callers retain the old representation.
- Anchor attributes (fixes typo): adds `rel(ARel)` and retains `ref(ARel) `as a warning-level deprecated alias. Both write the same rel attribute.

Observable behavior changes:
- CSS number formatting follows a shared binary32 policy. Long values and integral values with absolute value below `2^53` retain exact digits, other numbers use the shortest decimal for the corresponding float. This is needed for having matching behavior on JVM and JS.
- Numeric StylePropertyValue factories now return formatted strings on JS.
- Inline custom properties receive the required `--` prefix.

## Testing
```
# Common and JVM CSS tests
./gradlew :html-core:jvmTest \
  -Pcompose.web.tests.skip.benchmarks=true

# Common tests on JS, plus core and runtime browser tests (has 1 known failure, not related to this PR)
./gradlew :html-core:jsBrowserTest \
  :internal-html-core-runtime:jsBrowserTest --continue \
  -Pcompose.web.tests.skip.benchmarks=true
```

## Release Notes
N/A


.github/workflows/compose-html-eap.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
+
+      - name: Run Compose HTML EAP tests
+        working-directory: html
+        run: |
+          ./gradlew --no-daemon --stacktrace --rerun-tasks --max-workers=1 \
+            -Pcompose.html.eap.enabled=true \
+            :html-core-eap:check \
```
eymar, line 50: let's also run publishToMavenLocal for both of them too?

html/build.gradle.kts
```diff
@@ -78,8 +104,23 @@ subprojects {
...
+                            else -> "JetBrains Compose Multiplatform HTML library"
+                        }
+                    )
+                    description.set(
+                        when (projectName) {
+                            "html-core-eap" ->
+                                "Experimental commonized JetBrains Compose Multiplatform HTML core library"
+                            "internal-html-core-runtime-eap" ->
+                                "Experimental commonized JetBrains Compose Multiplatform HTML internal runtime"
+                            else -> "JetBrains Compose Multiplatform HTML library"
```
eymar: Noticed just now.
The old text was lying.
It was not Multiplatform at all.

Let's remove "Multiplatform" word from all branches?

html/eap/core/build.gradle.kts
```diff
@@ -0,0 +1,72 @@
...
+                implementation(kotlin("test"))
+            }
+        }
+
+        val jsMain by getting {
+            languageSettings {
+                optIn("org.jetbrains.compose.web.internal.runtime.ComposeWebInternalApi")
+            }
+            dependencies {
+                implementation(project(":internal-html-core-runtime-eap"))
```
eymar: Since it's already added in commonMain, it's redundant here.

html/eap/core/build.gradle.kts
```diff
@@ -0,0 +1,72 @@
...
+    sourceSets {
+        val commonMain by getting {
+            dependencies {
+                implementation(compose.runtime)
+                implementation(libs.kotlinx.coroutines.core)
+                implementation(
+                    "org.jetbrains.compose.html:kotlinx-browser-common-subset:" +
+                        composeHtmlEapKotlinxBrowserCommonSubsetVersion
+                )
+                implementation(project(":internal-html-core-runtime-eap"))
```
eymar: Since it was already `api(...)` in the existing html-core, it's okay to have it as `api` here too. 
```suggestion
                api(project(":internal-html-core-runtime-eap"))
```

html/build.gradle.kts
```diff
@@ -32,8 +47,19 @@ subprojects {
...
-    group = "org.jetbrains.compose.html"
-    version = COMPOSE_WEB_VERSION
+    val isComposeHtmlEapProject = project.name in setOf(
+        "html-core-eap",
+        "internal-html-core-runtime-eap",
+    )
+    if (isComposeHtmlEapProject) {
+        group = "org.jetbrains.compose.html.eap"
+        version = COMPOSE_HTML_EAP_VERSION
+        rootProject.extra["composeHtmlEapKotlinxBrowserCommonSubsetVersion"] =
```
eymar: is it for subprojects, right?


The subprojects can access `compose.html.eap.kotlinx-browser-common-subset.version` directly:

```
val kotlinxBrowserCommonSubsetVersion: String = providers.gradleProperty("compose.html.eap.kotlinx-browser-common-subset.version").get()
```

clemenskoprolin: Good point, thanks.

html/eap/core/build.gradle.kts
```diff
@@ -0,0 +1,70 @@
...
+            }
+        }
+    }
+}
+
+configurations.configureEach {
+    resolutionStrategy.dependencySubstitution {
+        substitute(project(":internal-html-core-runtime"))
+            .using(project(":internal-html-core-runtime-eap"))
+            .because("Compose HTML EAP tests must use the EAP runtime exclusively")
```
eymar, line 65: Just to double check. Is this dependency substituion only for tests?

eymar: If it's only for tests, then let's make it explicit? 

```
configurations.matching { it.name.contains("Test") }.configureEach {
    resolutionStrategy.dependencySubstitution {
        substitute(project(":internal-html-core-runtime"))
            .using(project(":internal-html-core-runtime-eap"))
            .because("Compose HTML EAP tests must use the EAP runtime exclusively")
    }
}
```

html/eap/internal-html-core-runtime/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+
+    sourceSets {
+        val commonMain by getting {
+            dependencies {
+                implementation(compose.runtime)
+                implementation(libs.kotlinx.coroutines.core)
+                implementation(
+                    "org.jetbrains.compose.html:kotlinx-browser-common-subset:" +
+                        kotlinxBrowserCommonSubsetVersion
+                )
```
eymar:
```suggestion
                implementation( "org.jetbrains.compose.html:kotlinx-browser-common-subset:$kotlinxBrowserCommonSubsetVersion")
```

html/eap/core/build.gradle.kts
```diff
@@ -0,0 +1,70 @@
...
+
+    sourceSets {
+        val commonMain by getting {
+            dependencies {
+                implementation(compose.runtime)
+                implementation(libs.kotlinx.coroutines.core)
+                implementation(
+                    "org.jetbrains.compose.html:kotlinx-browser-common-subset:" +
+                        kotlinxBrowserCommonSubsetVersion
+                )
```
eymar:
```suggestion
                implementation( "org.jetbrains.compose.html:kotlinx-browser-common-subset:$kotlinxBrowserCommonSubsetVersion")
```

html/eap/core/build.gradle.kts
clemenskoprolin: Yes. The runtime only enters through html-test-utils in the test dependency graph. I've scoped the substitution explicitly to test configurations.

eymar (approved):

PR #5705, opened 2026-09-09
Add HTML string rendering to commonized Compose HTML core

clemenskoprolin:
This PR adds HTML string rendering to the commonized Compose HTML EAP core, enabling Compose HTML content to be rendered on JVM without a browser DOM.

`composeHtmlToString` performs the initial composition into an in-memory HTML tree, serializes it, and disposes the composition afterward. The existing browser renderer and the new string renderer share the element APIs through an internal rendering context.

The serializer handles HTML escaping, void elements, boolean attributes, classes, inline styles, stylesheets, and raw-text elements. Raw text for `script`, `style`, `iframe`, and similar elements is emitted unescaped only after validation that it cannot terminate the surrounding element. Element children in raw-text elements and inline script content combined with a `src` attribute are rejected.

The PR also adds the document-level `Html`, `Head`, `Body`, `Title`, `Meta`, `Link`, `Script`, and `Style` APIs, including an explicit `InlineScript` carrier for trusted inline scripts. Client-only behavior such as event listeners, refs, and DOM property updates is not serialized. Coroutine effects such as `LaunchedEffect` do not run, while synchronous `SideEffect` and `DisposableEffect` behavior is retained.

An EAP example renders a complete page to `html/eap/example/build/index.html`.

Fixes [CMP-10656](https://youtrack.jetbrains.com/issue/CMP-10656/Integrate-kotlinx-browser-common-subset-in-compose-html-core)

## Testing

```shell
./gradlew :html-core:jvmTest :html-core:jsBrowserTest \
  -Pcompose.web.tests.skip.benchmarks=true

./gradlew :html-eap-example:run \
  -Pcompose.html.eap.enabled=true
```

## Release Notes
N/A


html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/Elements.kt
```diff
@@ -512,18 +639,38 @@ fun Button(
...
 ) {
-    // Changes to this key trigger controlled input state restoration.
+    val context = LocalComposeHtmlContext.current
+
+    // Changes to this key trigger controlled input state restoration in a DOM renderer.
     val keyForRestoringControlledState: MutableState<Int> = remember { mutableStateOf(0) }
 
-    TagElement(
-        elementBuilder = InputElementBuilder,
+    val domEffects: ContentBuilder<HTMLInputElement>? = if (context.supportsDomElementAccess) {
```
eymar: What do you think about simplifying a bit like this:

pass the lambda directly to content:

```kotlin
content = {
    if (!context.supportsDomElementAccess) return@TagElement

   if (type == InputType.Radio) {
        DisposeRadioGroupEffect()
    }
    DisposableEffect(keyForRestoringControlledState.value) {
        restoreControlledInputState(inputElement = scopeElement)
        onDispose { }
    }
}
```

And similar in TextArea?

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/Elements.kt
```diff
@@ -567,20 +706,35 @@ fun <K> Input(type: InputType<K>) {
  *      defaultValue("Some Default Text")
  * }
  * ```
+ *
+ * String rendering currently omits [value] and `defaultValue`.
```
eymar: Let's leave a TODO to support it. Perhaps to set either of value to the innerText of `<textarea>defaultValue</textArea>`?

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/InputElements.kt
```diff
@@ -236,6 +237,7 @@ fun TelInput(value: String = "", attrs: InputAttrsScope<String>.() -> Unit = {})
  * It's a controlled Input of [InputType.Text].
  * Controlled input means that its state is always equal [value].
  * If you need an uncontrolled behaviour, see [Input].
+ * String rendering currently omits [value].
```
eymar: Let's create a YT ticket for this and similar cases and link it here in a TODO.

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/Elements.kt
clemenskoprolin: Yes, good idea!

eymar (approved):

PR #5706, opened 2026-09-09
Strip JRE fonts by default

m-sasha:
Added a property to control JRE runtime font stripping and made it `true` by default.

To include the fonts, use 
```
compose.desktop {
    application {
        nativeDistributions {
            stripJreFonts = false
        }
    }
}
```

Note that not all JREs have bundled fonts, so in those JREs this change won't affect the distributable size.
In JBR 25, this should save about 5MB.

Fixes https://youtrack.jetbrains.com/issue/CMP-10772/Strip-JRE-fonts-from-desktop-distributable

## Testing
Added tests and tested manually

### Migration Notes - Desktop
- JRE fonts are now stripped from the distributable by default. If you need them, use `stripJreFonts = false` in the `nativeDistributions` section of your Gradle build file.

### DIFF
--- a/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
+++ b/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/desktop/application/dsl/JvmApplicationDistributions.kt
@@ -18,6 +18,7 @@ abstract class JvmApplicationDistributions : AbstractDistributions() {
         this.modules.addAll(modules.toList())
     }
     var includeAllModules: Boolean = false
+    var stripJreFonts: Boolean = true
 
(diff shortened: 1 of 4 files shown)


m-sasha:
I checked Swing interop with stripped fonts and it works fine.

igordmn:
Needs 2 approvals

igordmn (approved):

igordmn:
This needs to be documented in https://kotlinlang.org/docs/multiplatform/compose-native-distribution.html#platform-specific-options.

I would create a separate issue, and add the To be documented tag

m-sasha:
> This needs to be documented in https://kotlinlang.org/docs/multiplatform/compose-native-distribution.html#platform-specific-options.
> 
> I would create a separate issue, and add the To be documented tag

Done: https://youtrack.jetbrains.com/issue/CMP-10772/Strip-JRE-fonts-from-desktop-distributable

terrakok (approved):

PR #5708, opened 2026-09-09
Resources: perfomance improvements in web

terrakok:
Use a default dispatcher for async resource caches and yield before decoding image resources on web

### Fixes - Resources
- Optimise resource library on web to let a browser to draw frames between icons decoding.

### DIFF
--- a/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
+++ b/components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
@@ -150,9 +150,15 @@ private sealed interface ImageCache {
 
 private val imageCache = AsyncCache<String, ImageCache>()
 
+internal expect suspend fun webYield()
+
 private suspend fun loadImage(
     path: String,
     cacheKey: String,
     resourceReader: ResourceReader,
     decode: (ByteArray) -> ImageCache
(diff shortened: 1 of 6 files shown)


components/resources/library/src/commonMain/kotlin/org/jetbrains/compose/resources/ImageResources.kt
```diff
@@ -150,9 +150,15 @@ private sealed interface ImageCache {
...
+
 private suspend fun loadImage(
     path: String,
     cacheKey: String,
     resourceReader: ResourceReader,
     decode: (ByteArray) -> ImageCache
-): ImageCache = imageCache.getOrLoad(cacheKey) { decode(resourceReader.read(path)) }
+): ImageCache = imageCache.getOrLoad(cacheKey) {
+    val bytes = resourceReader.read(path)
+    webYield()
```
eymar, line 162: > let a browser to draw frames between icons decoding.

yielding doesn't always reschedule the continuation to the next frame.

Dispatchers.Default executes in batches of 16 items. If the first item takes a lot (e.g. 16ms), all other items in the batch will continue in this same frame. 

To "respect" the frame budget we likely need a custom dispatcher. 

__

Anyway, yielding can help to split one larger task into smaller ones - increasing the chances of scheduling to the next frame.

eymar (approved):

PR #5710, opened 2026-09-11
Add browser hydration to Compose HTML EAP

clemenskoprolin:
This PR adds hydration to the commonized Compose HTML EAP core.

`renderHydratedDocument` and `HydrationRoot` render a complete HTML document on JVM with a hydratable root and safely embedded public initial state. In the browser, `hydrateRoot` deserializes that state and adopts the existing server-rendered DOM. `hydrateComposable` supports adopting an existing root directly.

Hydration validates the server DOM before applying properties, effects, and event listeners. Structural mismatches are reported and fall back to a normal client render. `allowHydrationMismatch` supports intentionally dynamic server content.

Preserves logical text boundaries, handles raw-text, rejects HTML that the parser cannot reproduce reliably and cleans up compositions and recomposers. JVM-generated fixtures verify that server output can be hydrated and updated in the browser.

An EAP hydration example with an interactively hydrated counter is in `html/eap/examples/hydration`.

## Testing

```
./gradlew --stacktrace \
  :html-core-eap:check \
  :internal-html-core-runtime-eap:check
```

## Release Notes

N/A


html/test-utils/conf/karma.config.common.d/serve-html-test-resources.js
```diff
@@ -0,0 +1,7 @@
+// Each module's generated SSR hydration fixtures are copied into its Kotlin test resources.
```
eymar: Let's specify that this is necessary for eap/core k/js tests

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
```diff
@@ -126,6 +126,25 @@ interface AttrsScope<out TElement : Element> : EventsListenerScope {
...
+     *      Text(OffsetDateTime.now().toString())
+     * }
+     * ```
+     *
+     * Applies one level deep only: nested elements need their own call. Structural differences,
+     * such as a different tag name or a different number of children, still fail hydration.
+     *
+     * The call is never rendered as an attribute and has no effect outside hydration.
+     */
+    fun allowHydrationMismatch() = Unit
```
eymar, line 149: what's the expected behaviour if `allowHydrationMismatch()` is not called, but there is still a mismatch?

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/StringHtmlNodes.kt
```diff
@@ -168,23 +221,27 @@ internal class StringHtmlElementNode private constructor(
...
     var content: RawTextContent,
 ) : StringHtmlNode {
-    override fun appendHtmlTo(builder: StringBuilder) {
+    override fun appendHtmlTo(builder: StringBuilder, hydratable: Boolean) {
         builder.append(content.text)
     }
 }
 
+private fun StringBuilder.appendHydrationTextBoundaryMarker() {
+    append("<!--").append(HydrationTextBoundaryMarker).append("-->")
```
eymar: any reason not to have just one append?

```
 append("<!--c-->")
```

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/ComposeHtmlToString.kt
```diff
@@ -29,11 +29,20 @@ import org.jetbrains.compose.web.dom.StringHtmlNodeWrapper
...
  *
  * @throws IllegalArgumentException if a raw-text element contains unsafe text or element children,
- * or serialized `noscript` contents contain a `</noscript>` end tag.
+ * serialized `noscript` contents contain a `</noscript>` end tag, or ordinary text, RCDATA,
+ * or attribute values contain NUL (U+0000), which HTML parsing cannot preserve.
  */
 fun composeHtmlToString(
-    content: @Composable () -> Unit
-): String {
+    hydratable: Boolean = true,
```
eymar, line 38: Could you please add the description of `hydratable` parameter to the KDoc?

html/eap/core/src/commonTest/kotlin/org/jetbrains/compose/web/dom/ComposeHtmlToStringTest.kt
```diff
@@ -88,6 +102,20 @@ class ComposeHtmlToStringTest {
         assertEquals("<button>Click</button>", html)
     }
 
+    @Test
+    fun hydrationMismatchAllowanceIsNotSerialized() {
+        val html = composeHtmlToString {
+            Span(attrs = {
+                allowHydrationMismatch()
```
eymar, line 110: This test passes regardless of `allowHydrationMismatch()`. I tried to comment this line and the test passed. Is it expected?

html/eap/core/src/commonTest/kotlin/org/jetbrains/compose/web/SsrHydrationStateFixture.kt
```diff
@@ -0,0 +1,67 @@
...
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import org.jetbrains.compose.web.dom.Button
+import org.jetbrains.compose.web.dom.Span
+import org.jetbrains.compose.web.dom.Text
+
+// Shared declarations for the JVM-rendered fixture and the JS hydration test.
+internal const val SSR_HYDRATION_STATE_FIXTURE_URL = "/base/kotlin/ssr-hydration-state.html"
```
eymar: This path might change later when we update Kotlin version. Since we update the Kotlin version in all the projects using a special script,  later it might add additional fricition to investigate why the tests started to fail here.
 
It would be more reliable to put the fixtures directly to kjs-test-resources. 
The the path will be simply "./ssr-hydration-state.html"

Similar to:
https://github.com/JetBrains/compose-multiplatform/blob/master/gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/web/internal/configureWebApplication.kt#L74

html/eap/examples/hydration/src/commonMain/kotlin/App.kt
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    @Composable
+    fun Counter(initialCount: Int) {
+        var count by remember(initialCount) { mutableStateOf(initialCount) }
+
+        Article {
+            Div(attrs = { classes(PageStyles.counter) }) {
+                Div {
+                    Text(count.toString())
```
eymar:
```suggestion
                    Text("${'$'}count")
```

It will look like this in browser:

`Text("$count")` 

it's a bit more in Kotlin style :)

html/eap/examples/hydration/src/commonMain/kotlin/App.kt
```diff
@@ -0,0 +1,207 @@
...
+
+    Article(attrs = {
+        classes(PageStyles.demo)
+        if (count != initialCount) {
+            classes(if (count % 2 == 0) PageStyles.bloomEven else PageStyles.bloomOdd)
+        }
+    }) {
+        Div(attrs = { classes(PageStyles.counter) }) {
+            Div {
+                B(attrs = { classes(PageStyles.number) }) { Text(count.toString()) }
```
eymar:
```suggestion
                B(attrs = { classes(PageStyles.number) }) { Text("$count")) }
```

html/test-utils/conf/karma.config.common.d/serve-html-test-resources.js
eymar: Perhaps this won't be needed after this - https://github.com/JetBrains/compose-multiplatform/pull/5710#discussion_r3989873430

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/attributes/AttrsScope.kt
clemenskoprolin: Compose falls back to a normal client render, after calling onHydrationMismatch (logged by default). I'll add a comment explaining that.

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/StringHtmlNodes.kt
clemenskoprolin: Good point, thanks!

html/eap/core/src/commonTest/kotlin/org/jetbrains/compose/web/dom/ComposeHtmlToStringTest.kt
clemenskoprolin: Yes, that is expected. `allowHydrationMismatch()` must not change the generated HTML. I'll add a comment explaining it better.

html/eap/core/src/commonTest/kotlin/org/jetbrains/compose/web/SsrHydrationStateFixture.kt
clemenskoprolin: Thanks, updated the fixtures to be copied into the Kotlin/JS test working directory.

eymar (approved):

PR #5715, opened 2026-09-15
Add common SVG support to Compose HTML EAP

clemenskoprolin:
This PR adds a commonized Compose HTML SVG EAP artifact for JVM and Kotlin/JS

It preserves the existing Compose HTML SVG API while moving its declarations to `commonMain` using `kotlinx-browser-common-subset` types. Namespace-aware element creation, string rendering, and hydration preserve SVG names and attributes across platforms.

JVM-generated fixtures verify that rendered SVG can be hydrated and updated in the browser. The new `org.jetbrains.compose.html.eap:html-svg-eap` artifact is included in EAP publication and CI.

Solves [CMP-10794](https://youtrack.jetbrains.com/issue/CMP-10794/Add-SVG-support-to-Compose-HTML-EAP)

## Testing

```shell
./gradlew --stacktrace \
  :html-core-eap:check \
  :html-svg-eap:check
```

## Release Notes

N/A


html/eap/svg/src/commonMain/kotlin/org/jetbrains/compose/web/svg/svg.kt
```diff
@@ -0,0 +1,776 @@
...
+@Composable
+@ExperimentalComposeWebSvgApi
+fun Svg(
+  viewBox: String? = null,
+  attrs: AttrBuilderContext<SVGElement>? = null,
+  content: ContentBuilder<SVGElement>? = null
+) {
+    TagElementNS<SVGElement>(
+        tagName = "svg",
+        namespace = SVG_NS,
```
eymar: To avoid the repetition of SVG_NS, let's introduce a special TagElementSvgNS:

```
fun <TElement : Element> TagElementSvgNS(
    tagName: String,
    applyAttrs: (AttrsScope<TElement>.() -> Unit)?,
    content: (@Composable ElementScope<TElement>.() -> Unit)?,
) {
    val context = LocalComposeHtmlContext.current

    key(tagName) {
        context.TagElement(
            elementBuilder = context.elementBuilderNS(tagName, SVG_NS),
            applyAttrs = applyAttrs,
            content = content,
        )
    }
}
```

WDYT?

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/Base.kt
```diff
@@ -14,17 +14,23 @@ import kotlinx.browser.dom.HTMLStyleElement
...
 import org.jetbrains.compose.web.css.utils.serializeRules
+import org.jetbrains.compose.web.internal.runtime.ComposeWebInternalApi
 
 
-// Match the HTML parser's ASCII-only attribute-name normalization in both render paths.
-internal fun Map<String, String>.containsAttribute(name: String): Boolean =
-    keys.any { it.asciiLowercase() == name }
+// HTML attribute names are ASCII-insensitive; foreign attribute names retain their case.
+internal fun Map<String, String>.containsAttribute(name: String, namespace: String?): Boolean =
+    if (namespace == HtmlNamespace) keys.any { it.asciiLowercase() == name } else name in this
```
eymar: I haven't asked last time, but now I wonder:

```
it.asciiLowercase()
```

does it make sense to normalize the attributes names before putting the in a map?
Or even simpler: can we assume the attributes are already normalized?

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/ElementBuilder.kt
```diff
@@ -6,6 +6,12 @@
 package org.jetbrains.compose.web.dom
 
 import kotlinx.browser.dom.Element
+import org.jetbrains.compose.web.internal.runtime.ComposeWebInternalApi
+
+internal const val HtmlNamespace = "http://www.w3.org/1999/xhtml"
+
+internal fun normalizeElementTagName(tagName: String, namespace: String): String =
```
eymar: same here.

Can we assume the name is already normalized? 

We control most of the names, unless someone uses TagElement directly - but it's their responsibilty.

Am I mistaken?

eymar: Noticing `asciiLowercase` in many places. 

Running the loop repeatedly might affect the performance.

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/HtmlParserValidation.kt
```diff
@@ -52,3 +66,182 @@ private fun requireHtmlParserStableTableText(parentTagName: String?, text: Strin
...
+    }
+    if (childNamespace == parserNamespace) return
+
+    val location = parentTagName?.let { "inside <$it>" } ?: "at the string-rendering root"
+    val suggestion = if (childNamespace == SvgNamespace) {
+        "; wrap SVG fragments in an SVG <svg> root"
+    } else {
+        ""
+    }
+    throw IllegalArgumentException(
```
eymar: Does it verify that html tags not include svg tags and vice versa?

I think this can be achieved at compile time rather than runtime - we have a task to do that in compose  by using  ComposableTargetMarker html https://youtrack.jetbrains.com/issue/CMP-7107 (can be done separately, not in this PR)


A related idea:

For simplicity, can we assume the API is used correctly so runtime validation/parsing is not necessary?

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/ElementBuilder.kt
clemenskoprolin: I agree with you there, but `ElementBuilder.createBuilder("DIV")` already normalized before any of the SSR work. So that would be a compatibility change.

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/Base.kt
clemenskoprolin: The problem is we allow arbitrary values through `attr`:
```
attr("CLASS", "first")
attr("class", "second")
```
That wasn't a problem for k/JS: it implicitly normalized those values. But with String rendering, we need to that explicitly. Otherwise, we we wouldn't e.g. update the class field for this element correctly.

We cannot simply normalize everything before putting it in the map since foreign attributes like `viewBox` would be saved as `viewbox`. SVG attribute names are case-sensitive in the DOM and would produce the wrong results. 

We could make `AttrsScopeBuilder.attr()` namespace-aware and normalize at insertion, but that would be a larger change.

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/ElementBuilder.kt
clemenskoprolin: For the repeated `asciiLowercase` calls: I'll remove unnecessary ones, thanks :)

html/eap/core/src/commonMain/kotlin/org/jetbrains/compose/web/dom/HtmlParserValidation.kt
clemenskoprolin: Yes, but it does more than that. ComposeableTargetMarker couldn't fully replace it, I think. Markers could catch incompatible composable calls, but wouldn’t validate dynamic `SvgElement(name)` names or attribute casing. 

e.g.
```
Svg {
      SvgElement<SVGElement>("myCustomTag")
  }
```
causes an hydration mismatch (only lowercase custom tag names are allowed for non-standard SVG tags).

React for example does this kind of validation only in dev and not in production, might this be a option? Or ofc, we could just ignore these cases as a known limitation.

eymar: > we could just ignore these cases as a known limitation.

yes, tbh, i'd prefer going simpler for now.  The hydration mismatch will be reported anyway - something helpful.

We may indeed come up with an additional parameter/flag to enable strict validation on SSR side, since running the same validation repeatedly on already valid Compose tree is redundant (at least for production).

eymar (approved):

PR #5718, opened 2026-09-16
Upgrade to Kotlin 2.4.20

eymar:
paired with https://github.com/JetBrains/compose-multiplatform-core/pull/3426

Green Snapshot build - https://teamcity.jetbrains.com/buildConfiguration/JetBrainsPublicProjects_Compose_AllPersonalBuild/6591811

## Release Notes
N/A


gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
```diff
@@ -116,11 +116,14 @@ class GradlePluginTest : GradlePluginTestBase() {
         }
     }
 
+    // Note: we can't test non-jvm targets with Kotlin older than 2.4.0, because of klib abi version bump in 2.4.0
+    private val oldestSupportedKotlinMultiplatformVersion = "2.4.0"
```
MatkovIvan, line 120: It's only valid for native. Web requires 2.4.20 as far as I understand

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
```diff
@@ -99,7 +99,7 @@ class GradlePluginTest : GradlePluginTestBase() {
             }
         }
 
-    // Note: we can't test non-jvm targets with Kotlin older than 2.3.0, because of klib abi version bump in 2.3.0
+    // Note: older version is fine for k/jvm, k/android projects
     private val oldestSupportedKotlinVersion = "2.3.0"
```
MatkovIvan, line 103: Let's rename it to something about JVM

gradle-plugins/gradle.properties
```diff
@@ -8,10 +8,10 @@ kotlin.code.style=official
 dev.junit.parallel=false
 
 # Default version of Compose Libraries used by Gradle plugin
-compose.version=1.12.10-alpha01+dev4472
+compose.version=1.13.0-alpha02+snapshot.ok-upd-kotlin-2-4-20
```
MatkovIvan, line 11: probably requires to be updated before merge

html/kotlin-js-store/yarn.lock
MatkovIvan, line 1: Maybe we can add it to ignores like in core? Does it matter here?

html/gradle.properties
```diff
@@ -2,7 +2,7 @@
 org.gradle.jvmargs=-Xmx8g
 
 #Versions
-compose.version=1.10.1
+compose.version=1.12.0
```
MatkovIvan, line 5: It breaks compatibility with older runtime versions. Libraries should depend on the lowest compatible version, not the latest.
What's the reason for this change? It does not seem related to the scope here

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
eymar: The test passed though.
ABI bump doesn't always happen in .*.20 releases. Usually in major releases (2.3, 2.4. etc)

html/gradle.properties
eymar: automatically updated by the script.

But, it's just outdated source code, since CI propagates the newest -Pcompose.version for every build.

____

> It breaks compatibility with older runtime versions.

Do you mean compose-runtime? (I assume it's backward compatible)

MatkovIvan: > compose-runtime

yep. realized that it's probably not what it uses in dependencies

gradle-plugins/compose/src/test/kotlin/org/jetbrains/compose/test/tests/integration/GradlePluginTest.kt
eymar: fixed

gradle-plugins/gradle.properties
eymar: fixed

MatkovIvan (approved):

html/kotlin-js-store/yarn.lock
eymar: Added it to gitignore. Now this PR also removes the file (-3k lines) to untrack it. It's a bit irrelevant. Maybe let's move it to a separate change?

MatkovIvan: Up to you

PR #5722, opened 2026-09-17
Add Kotlin/Wasm JS support to Compose HTML EAP

clemenskoprolin:
This PR adds Kotlin/Wasm JS as a target for the Compose HTML EAP core, SVG, runtime, and shared test utilities.

Browser implementations and tests move into shared `webMain` and `webTest` source sets. Platform-specific DOM operations use implementations for JS and Wasm, including weak maps, microtask scheduling, input events, nonce access, event listeners and test utilities. Typed implementations shared by JVM and Wasm live in `nonJsMain`.

Typed Wasm interop exposed two existing problems. `TextAreaAttrsScope` used an `HTMLInputElement` value setter that only worked on JS because the cast erased. Controlled input restoration could happen before all event listeners ran (on WASM/JS). Text areas now use the correct setter and restoration waits until all listeners finish.

Existing browser tests run on both JS and Wasm. Cases requiring JS-only interop remain in `jsTest`.

## Testing

```shell
./gradlew -Pcompose.html.eap.enabled=true \
  :html-core-eap:check \
  :html-svg-eap:check \
  :internal-html-core-runtime-eap:check \
  :html-test-utils:check
```

Fixes [CMP-10816](https://youtrack.jetbrains.com/issue/CMP-10816/Compose-HTML-EAP-Support-Kotlin-Wasm-as-a-browser-target)

## Release Notes

N/A


html/test-utils/build.gradle.kts
```diff
@@ -13,27 +19,61 @@ repositories {
...
                 implementation(compose.runtime)
                 implementation(kotlin("stdlib-common"))
-                implementation(project(":internal-html-core-runtime"))
                 implementation(libs.kotlinx.coroutines.core)
             }
         }
-        val jsTest by getting {
+        val webMain by getting {
+            dependencies {
+                implementation(
```
eymar: `test-utils` is a public non-eap module and this will force the users of non-eap compose-html to add a dev maven repo.

Will it be enough to replace it by the dependency on the original kotlinx-browser?

eymar (commented):
Let's add the k/wasm target to the examples/hydration?

html/test-utils/build.gradle.kts
clemenskoprolin: Good point, thanks. Yes, it seems to be enough.

eymar (approved):

PR #5723, opened 2026-09-21
Update currentOs deprecation message

hub-bla:
In some cases, such as the `desktopApp` module in the current KMP project structure, `currentOs` was used to transitively get other Compose modules (like Compose UI for `androidx.compose.ui.window.application` and `androidx.compose.ui.window.Window`). This PR updates the deprecation message to clarify that in certain cases a replacement for `currentOs` is required.

## Release Notes
N/A


gradle-plugins/compose/src/main/kotlin/org/jetbrains/compose/ComposePlugin.kt
```diff
@@ -135,7 +135,7 @@ abstract class ComposePlugin : Plugin<Project> {
 
         @Deprecated(
             "currentOs is deprecated and will be removed in a future release. " +
-                    "It has no replacement and can be safely removed. " +
+                    "If you relied on currentOs providing Compose modules transitively (for example, org.jetbrains.compose.ui:ui), add those dependencies explicitly. " +
```
igordmn: Worth to add the exact lines users need to add. I didn't test it (please, test), but probably this is it:
```
implementation("org.jetbrains.compose.ui:ui:<composeVersion>")
implementation("org.jetbrains.compose.foundation:foundation:<composeVersion>")
implementation("org.jetbrains.compose.material:material:<composeVersion>")
```

and if the user uses version catalog, they would know how to transform it to its DLS themselves.

hub-bla: What needs to be added depends on where `currentOs` was used. For a project created with the KMP wizard, adding just `org.jetbrains.compose.ui:ui:<composeVersion>` as a replacement is enough to get access to `androidx.compose.ui.window.*`

In fact, the `webApp` module already does exactly that. It only depends on `org.jetbrains.compose.ui:ui:<composeVersion>`

igordmn: > What needs to be added depends on where currentOs was used

We can list the major ones. Usually, users don't know what exactly is missing, thus we need to provide an easy to-copy-paste snippet, even if it contains unnecessary dependencies.

igordmn: > In fact, the webApp module already does exactly that. 

We are deprecating desktop-only `currentOs`, so it shouldn't be taken into an account. We need to provide an easy-to-use migration for the `currentOs` users only.

hub-bla: Got it, updated the message

igordmn (approved):

